PR #1, opened 2016-06-21
Fix event list performance issues.

jawalonoski:


lib/entity/event_list.rb
```diff
@@ -1,32 +1,62 @@
...
 
-    def before(date)
-      self.select{|x|x.time <= date}
+    def before(date,type=:all)
+      list = @events[type]
+      return EMPTY if list.nil? || list.empty?
+      end_index = list.index{|e|e.time >= date}
+      end_index = list.length if end_index.nil?
+      if !end_index.nil?
+        list[0..(end_index-1)]
```
mzwong: If a date is used that comes before any event in the list, then end_index is 0. Line 37: list[0..-1] may cause problems.

Also the else statement is never called because end_index will never be nil since it's set in the previous line.

lib/modules/encounters.rb
```diff
@@ -38,17 +38,17 @@ class Encounters < Synthea::Rules
                   6.months
               end
             end
-            next_date = time + Distribution::Normal.rng(delta, delta*schedule_variance).call
+            next_date = time + delta #Distribution::Normal.rng(delta, delta*schedule_variance).call
```
mzwong: Changed to 'delta' for running jruby. Need to change back.

test/unit/immunizations_test.rb
```diff
@@ -6,7 +6,7 @@ def setup
   	@patient = Synthea::Person.new
     @patient[:age] = 0
     @time = Synthea::Config.start_date
-	@patient.events.create(@time, :birth, :birth, true)
+	  @patient.events.create(@time, :birth, :birth, true)
```
mzwong: space?

PR #10, opened 2016-07-07
print rules in order

mzwong:


jawalonoski:
I don't think you want to merge this...

PR #11, opened 2016-07-08
Metabolic treatment

mzwong:


lib/modules/metabolic_syndrome.rb
```diff
@@ -221,13 +284,88 @@ def update_diabetes(severity,time,entity)
...
+          entity[:medications][:metformin] = [ time, :diabetes ] if entity[:medications][:metformin].nil?
+          entity[:medications][:glp1ra] = [ time, :diabetes ] if entity[:medications][:glp1ra].nil?     
+          entity[:medications][:sglt2i] = [ time, :diabetes ] if entity[:medications][:sglt2i].nil?  
+          # prescribe insulin
+          if entity[:medications][:basal_insulin]
+            entity[:medications].delete(:basal_insulin)
+            entity[:medications][:prandial_insulin] = [ time, :diabetes ]            
+          else
+            entity[:medications][:basal_insulin] = [ time, :diabetes ]
+          end
```
cmoesel: This previous `if`/`else`/`end` block gets triggered (for diabetes severity of 4) with every rule cycle (currently weekly).  Based on the current logic, this would cause the following to happen:

* week 1: add `:basil_insulin`
* week 2: remove `:basil_insulin` and add `:prandial_insulin`
* week 3: add `:basil_insulin` (again)
* week 4: remove `basil_insulin` and add `:prandial_insulin` (again)

_rinse, lather, repeat_

lib/records/fhir.rb
```diff
@@ -240,6 +241,84 @@ def self.immunization(imm, fhir_record, patient, encounter)
         fhir_record.entry << entry
       end
 
+      def self.careplans(plan, fhir_record, patient, encounter)
+        careplanData = CAREPLAN_LOOKUP[plan['type']]
+        reasonCode = COND_LOOKUP[plan['reason']][:codes]['SNOMED-CT'][0]
+        reason = fhir_record.entry.find{|e| e.resource.is_a?(FHIR::Condition) && e.resource.code.coding.find{|c|c.code==reasonCode} }
+        careplan = FHIR::CarePlan.new({
+          'subject' => {'reference'=> "Patient/#{patient.fullUrl}"},
```
cmoesel: Is this right?  Unless `#{patient.fullUrl}` resolves only to an ID, then I don't think this will be valid.

lib/records/ccda.rb
```diff
@@ -114,6 +116,36 @@ def self.procedure(procedure, ccda_record)
...
+            'codes' => CAREPLAN_LOOKUP[entry][:codes],
+            'description' => CAREPLAN_LOOKUP[entry][:description],
+            'start_time' => time.to_i,
+            'status' => plan['status'],
+            'reason' => COND_LOOKUP[plan['reason']]
+          })
+          care_goal['end_time'] = plan['stop'].to_i if plan['stop']
+          ccda_record.care_goals << care_goal
+        end
+      end
```
cmoesel: I've inspected the resultin C-CDA and noted a few things:

1. Only the narrative is exported.  There is no structured data for the care plan or goals.
2. In the narrative, the plan itself (e.g., _Diabetes self management plan_) is listed as a goal -- so we end up with three goals: the plan itself, diet, and exercise.

If this is a known limitation, fine.  I just wanted to mention it in case it was somehow missed.

lib/records/ccda.rb
```diff
@@ -114,6 +116,36 @@ def self.procedure(procedure, ccda_record)
...
+
+      def self.medications(prescription, ccda_record)
+        type = prescription['type']
+        time = prescription['time']
+        medication = Medication.new({
+          "codes" =>  MEDICATION_LOOKUP[type][:codes],
+          "description" => MEDICATION_LOOKUP[type][:description],
+          "start_time" => time.to_i,
+          'reason' => COND_LOOKUP[prescription['reason']]
+        })
```
cmoesel: The `status` is not recorded.  Is this intentional?

lib/records/fhir.rb
jawalonoski: `fullUrl` is the name of the FHIR field. The field is a `uri` and as such it can contain a `UUID`

lib/records/ccda.rb
jawalonoski: This is a known limitation. Fixing this would require updates to `healthdatastandards`.

lib/records/ccda.rb
mzwong: @jawalonoski was there a reason that there isn't a `status` field in the medications record hash?

jawalonoski: Yes, this is intentional... the status is a FHIR or CCDA thing... I'm making a change related to this... I will also be removing the status from CarePlan... it shouldn't be there either. The `stop` key determines whether or not this item is active, and the `status` code should be determined by the appropriate exporter.

lib/records/fhir.rb
cmoesel: Typically (in FHIR doc anyway), a UID only field has something like `urn:uuid:` as a prefix.  I think that the "full" part in "fullUrl" is meant to hint that at least a url scheme should be included.

PR #12, opened 2016-07-13
Cardiovascular treatment

mzwong:
Refactored some metabolic treatment code to use new medications functions that allow for multiple reasons. Added treatment and medications for cardiovascular diseases.

### DIFF
--- a/lib/modules/cardiovascular_disease.rb
+++ b/lib/modules/cardiovascular_disease.rb
@@ -356,7 +356,42 @@ module Synthea
           end
         end
       end
+      #-----------------------------------------------------------------------#
+      #Treatments and Medications
+      #-----------------------------------------------------------------------#
+      rule :heart_healthy_lifestyle, [:coronary_heart_disease, :stroke, :cardiac_arrest, :myocardial_infarction], [] do |time, entity|
+        if entity[:coronary_heart_disease] || entity[:stroke_history] || entity[:cardiac_arrest] || entity[:myocardial_infarction]
+          entity[:careplan] ||= Hash.new
+          entity[:careplan][:cardiovascular_disease] ||= [:exercise, :stress_management, :stop_smoking, :healthy_diet]
+        end
+      end
 
+      rule :chd_treatment, [:coronary_heart_disease], [:coronary_heart_disease] do |time, entity|
+        meds = [:clopidogrel, :simvastatin, :amlodipine, :nitroglycerin]
+        if entity[:coronary_heart_disease]
+          entity[:medications] ||= Hash.new
+          meds.each {|m| prescribeMedication(m, :coronary_heart_disease, time, entity)}
+        elsif entity[:medications]
+          meds.each do |m|
+            stopMedication(m, :coronary_heart_disease, time, entity) if entity[:medications][m]
+          end
+        end
+      end
+
+      rule :atrial_fibrillation_treatment, [:atrial_fibrillation], [:atrial_fibrillation] do |time, entity|
+        meds = [:warfarin, :verapamil, :digoxin]
+        if entity[:atrial_fibrillation]
+          entity[:medications] ||= Hash.new
+          meds.each {|m| prescribeMedication(m, :atrial_fibrillation, time, entity)}
+
+          entity[:cardiovascular_procedures] ||= Hash.new
+          entity[:cardiovascular_procedures][:atrial_fibrillation] ||= [:electrical_cardioversion, :catheter_ablation]
+        elsif entity[:medications]
+          meds.each do |m|
+            stopMedication(m, :atrial_fibrillation, time, entity) if entity[:medications][m]
+          end
+        end
+      end
       #-----------------------------------------------------------------------#
 
       def self.perform_encounter(entity, time)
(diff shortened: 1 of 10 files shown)


lib/records/fhir.rb
```diff
@@ -299,9 +302,12 @@ def self.medications(prescription, fhir_record, patient, encounter)
           'patient' => {'reference'=> "Patient/#{patient.fullUrl}"},
           'encounter' => {'reference'=> "Encounter/#{encounter.fullUrl}"},
           'dateWritten' => convertFhirDateTime(prescription['time']),
-          'reasonReference' => {'reference'=> "Condition/#{reason.fullUrl}"},
+          'reasonReference' => [] ,
           'activity' => []
         })
+        reasons.each do |r|
+          medOrder.reasonReference << {'reference'=> "Condition/#{r.fullUrl}"}
```
jawalonoski: You can only use the hashes inside the `FHIR::MedicationOrder.new`... if you add a field subsequently, you need to create the object. In this case, something like `FHIR::Reference.new`

lib/modules/module.rb
```diff
@@ -50,5 +50,11 @@ def prescribeMedication(med, reason, time, entity)
         entity[:medications][med][1] << reason
       end
     end
+
+    def stopMedication(med, reason, time, entity)
+      return if entity[:medications][med].nil?
+      entity[:medications][med][1].delete(reason) if entity[:medications][med][1].include?(reason)
+      entity[:medications].delete(med) if entity[:medications][med].empty?
```
jawalonoski: `entity[:medications][med]` will never be empty because it will always contain a time, right?

lib/modules/cardiovascular_disease.rb
```diff
@@ -356,7 +356,48 @@ class CardiovascularDisease < Synthea::Rules
...
       end
+      #-----------------------------------------------------------------------#
+      #Treatments and Medications
+      #-----------------------------------------------------------------------#
+      rule :heart_healthy_lifestyle, [:coronary_heart_disease, :stroke, :cardiac_arrest, :myocardial_infarction], [] do |time, entity|
+        reasons = []
+        [:coronary_heart_disease, :stroke, :cardiac_arrest, :myocardial_infarction].each do |disease|
+          reasons << disease if entity[disease] || entity.had_event?(disease)
+        end
+        puts entity[:name_last] if reasons.length > 1
```
cmoesel: Leftover from debugging?

lib/modules/cardiovascular_disease.rb
```diff
@@ -356,7 +356,48 @@ class CardiovascularDisease < Synthea::Rules
...
+      end
+
+      rule :atrial_fibrillation_treatment, [:atrial_fibrillation], [:atrial_fibrillation] do |time, entity|
+        meds = [:warfarin, :verapamil, :digoxin]
+        if entity[:atrial_fibrillation]
+          entity[:medications] ||= Hash.new
+          meds.each {|m| prescribeMedication(m, :atrial_fibrillation, time, entity)}
+
+          entity[:cardiovascular_procedures] ||= Hash.new
+          entity[:cardiovascular_procedures][:atrial_fibrillation] ||= [:electrical_cardioversion, :catheter_ablation]
```
cmoesel: I'm not sure that *every* afib patient is going to get all of those meds *plus* electrical cardioversion, *plus* catheter ablation.  This is probably not realistic.  According to WebMD, electrical cardioversion is only done right away if afib is discovered in its first 48 hours -- otherwise it might be weeks after the meds.  Catheter ablation seems even more rare, perhaps only being done when medication is proving ineffective or when the patient refuses the medication.  We may want to space some of these out by time and/or only apply them to selective people (not all afib patients).

lib/modules/cardiovascular_disease.rb
```diff
@@ -366,16 +407,72 @@ def self.perform_encounter(entity, time)
...
+            entity.record_synthea.update_careplan_reasons(:cardiovascular_disease, entity[:careplan][:cardiovascular_disease]['reasons'],time)
+          end
+        elsif entity.record_synthea.careplan_active?(:cardiovascular_disease)
+          entity.record_synthea.careplan_stop(:cardiovascular_disease,time)
+        end
+
+        if entity[:medications]
+          [:clopidogrel, :simvastatin, :amlodipine, :nitroglycerin, :atorvasatatin, :captopril, :warfarin, 
+            :verapamil, :digoxin, :epinephrine, :amiodarone, :atropine].each do |med|
+            if entity[:medications][med]
```
cmoesel: Is the above just a repetition (and aggregation) of all the meds that could have been prescribed in the rules of the module?  If so, consider if there is a better way to do this that is less repetitious and less error prone.  As it is now, we need to make sure we keep these lists in sync.  Potential approaches might involve using constants, having the module keep track of its own prescribed medications, dynamically looking at medications and reasons to get the list, etc.

All that said, if the solution is overly complex, then... just keep this all as is. ;-)

lib/modules/cardiovascular_disease.rb
```diff
@@ -366,16 +407,72 @@ def self.perform_encounter(entity, time)
...
+              # Add a prescription to the record if it hasn't been recorded yet
+              if !entity.record_synthea.medication_active?(med)
+                entity.record_synthea.medication_start(med, time, entity[:medications][med][1])
```
cmoesel: I think it would be easier to read (and author) if you used a hash to store time and reason, rather than array (that you have to remember which index is which).  This (obviously) goes for anywhere you're using `entity[:medications][med][1]` so I won't repeat myself in the other cases.

lib/modules/cardiovascular_disease.rb
```diff
@@ -366,16 +407,72 @@ def self.perform_encounter(entity, time)
...
+              entity.record_synthea.medication_stop(med, time, :cardiovascular_improved)
+            end
+          end
+        end
+
+        if entity[:cardiovascular_procedures]
+          entity[:cardiovascular_procedures].each do |reason, procedures|
+            reason_code = COND_LOOKUP[reason][:codes]['SNOMED-CT'][0]
+            procedures.each do |proc|
+              if !entity.record_synthea.present[proc]
```
cmoesel: Maybe an exercise for later (i.e., just mark a #TODO), but this assumes that a procedure will only ever be performed once.  Are we sure that's the case?  Again, for now it's probably fine, but we may want to revisit at some point.

lib/modules/cardiovascular_disease.rb
```diff
@@ -366,16 +407,72 @@ def self.perform_encounter(entity, time)
...
+                entity.record_synthea.procedure(proc, time, reason_code, :procedure, :procedure) 
+              end
+            end
+          end
+        end
       end
```
cmoesel: The above 5 lines are one of the (few) things I dislike about ruby.  End end end end end end end end end end end...

(Nothing to do here.  I just like complaining.)

lib/modules/metabolic_syndrome.rb
```diff
@@ -303,7 +303,7 @@ def update_diabetes(severity,time,entity)
           # delete medications above this stage...
           [:glp1ra,:sglt2i,:basal_insulin,:prandial_insulin].each{|m| entity[:medications].delete(m)}
           # prescribe metformin if it isn't already there...
-          entity[:medications][:metformin] = [ time, :diabetes ] if entity[:medications][:metformin].nil?
+          prescribeMedication(:metformin, :diabetes, time, entity)
```
cmoesel: I definitely like that you cleaned all this stuff up.  Kudos!

lib/modules/module.rb
```diff
@@ -42,5 +42,23 @@ def self.rule(name,inputs,outputs,&block)
...
       return 1-((1-risk) ** (Synthea::Config.time_step.to_f/original_period))
     end
+
+    #These functions are used to start/update medications and to stop medications, respectively.
+    #Similar functions were not written for care plans because a care plan is typically only started/updated in one
+    #rule, whereas it is common for a medication to be used in multiple rules. This may change in future modules(?) 
+    def prescribeMedication(med, reason, time, entity)
+      entity[:medications] ||= Hash.new
+      if entity[:medications][med].nil?
+        entity[:medications][med] = [ time, [reason ] ]
```
cmoesel: As noted previously (and discussed on slack), let's track time and reason with a hash rather than an array.

lib/records/ccda.rb
```diff
@@ -126,7 +125,7 @@ def self.careplans(plan, ccda_record)
             'codes' => CAREPLAN_LOOKUP[entry][:codes],
             'description' => CAREPLAN_LOOKUP[entry][:description],
             'start_time' => time.to_i,
-            'reason' => COND_LOOKUP[plan['reason']]
+            'reason' => COND_LOOKUP[plan['reasons'][0]]
```
cmoesel: We're potentially throwing away data here (and in line 148).  Does C-CDA just not support multiple reasons?

lib/records/record.rb
```diff
@@ -109,7 +118,8 @@ def careplan_start(type, activities, time, reason)
           'type' => type,
           'activities' => activities,
           'time' => time,
-          'reason' => reason
+          'start_time' => time,
+          'reasons' => reason
```
cmoesel: Might want to rename `reason` argument to `reasons` (for consistency and to avoid unintended misuse of the API).

lib/records/ccda.rb
mzwong: C-CDA to my knowledge doesn't support multiple reasons. Or at least Health data standards doesn't. (it's not always a single element array)

PR #15, opened 2016-08-01
Ccda upload

mzwong:
Added capability to sftp ccda records to server. Ran a bundle update on gemfile.


jawalonoski:
Remove the SFTP information from the config file. Pass the the URL into the rake task as a command-line parameter and prompt for username/password.

mzwong:
@jawalonoski 
done.

PR #17, opened 2016-08-18
add appendicitis using generic module framework

dehall:
Adds a generic module for Appendicitis. Basic flow is that patients are filtered by gender, and then are assigned to get or not get appendicitis. It picks an age that they will get it with a Delay, and then gives them the Appendicitis condition. There is a chance the appendix will rupture but it leads back into the same sequence of encounters and procedures for an appendectomy. 

Some things not modeled here that could be if we thought it was worthwhile: 
- Appendectomies where they didn't actually have appendicitis
- Complications of surgery
- Lab tests at the appendicitis encounter

This also adds a new State type: Simple. A simple state has no special logic like delay or guard, and doesn't add anything to the patient data, like encounter or procedures. I'm using it here so that we can have a sequence of branches.

appendicitis

### DIFF
--- a/lib/generic/modules/appendicitis.json
+++ b/lib/generic/modules/appendicitis.json
@@ -1,0 +1,195 @@
+{
+  "name": "Appendicitis",
+  "states": {
+    "Initial" : {
+      "type" : "Initial",
+      "conditional_transition": [
+        {
+          "condition": {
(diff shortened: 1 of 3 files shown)


jawalonoski:
Please add documentation of the Simple state to the generic module portion of the project wiki.

PR #18, opened 2016-08-23
sequential generator can generate 1 city per thread

dehall:
Adds the ability to generate 1 city at a time, using real city demographics, to the sequential generator. The generator can now run multithreaded or single-threaded. (Configurable by a synthea.yml property - for fastest performance use JRuby and enable multithreading. Don't enable multithreading on MRI)
Patients are generated in 1 thread per city, and then put into a thread pool to export them.

Adds a new optional parameter to the generate rake task, which accepts a json file of city demographics to use to generate patients. If no file is given it generates pure random patients, as it does today.

Example: 
`bundle exec rake synthea:sequential['./config/Barnstable_County.json']`

Some other changes were needed to accommodate targeting patients to demographics instead of just choosing them completely randomly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - 2.2.5
+  - "2.0.0"
 script:
   - bundle exec rake test
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - "2.0.0"
+  - 2.2.5
 script:
   - bundle exec rake test
 notifications:
--- a/lib/modules/encounters.rb
+++ b/lib/modules/encounters.rb
@@ -38,7 +38,7 @@ module Synthea
                   6.months
               end
             end
-            next_date = time + Distribution::Normal.rng(delta, delta*schedule_variance).call
+            next_date = time + Distribution::Normal.rng(delta*1.0, delta*schedule_variance).call
             entity.events.create(next_date, :encounter, :schedule_encounter)
           end
         end
--- a/lib/world/sequential.rb
+++ b/lib/world/sequential.rb
@@ -4,7 +4,7 @@ module Synthea
 
       attr_reader :stats
         
-      def initialize()
+      def initialize(datafile)
         @start_date = Synthea::Config.start_date
         @end_date = Synthea::Config.end_date
         @time_step = Synthea::Config.time_step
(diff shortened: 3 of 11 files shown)


lib/world/sequential.rb
```diff
@@ -28,28 +34,139 @@ def initialize()
...
-            Synthea::Rules.apply(@date,person)
-          end
-          record_stats(person)
-          puts "##{i+1}  #{person[:name_last]}, #{person[:name_first]}. #{person[:race].to_s.capitalize} #{person[:ethnicity].to_s.gsub('_',' ').capitalize}. #{person[:age]} y/o #{person[:gender]}."
-          export(person)
+        @threads = []
+
+        # 16 picked by experimentation to be a good number of threads to use for exporting records
+        # using a cachedthreadpool has no upper bound on the # and if it gets too high then everything grinds to a halt
+        @pool = Concurrent::FixedThreadPool.new(16) if Synthea::Config.sequential.multithreading
```
jawalonoski: Number of threads (16) should probably be a config parameter

lib/world/sequential.rb
```diff
@@ -18,6 +18,12 @@ def initialize()
 
         @population_count = Synthea::Config.sequential.population
 
+        @scaling_factor = @population_count / 6794422.0
```
jawalonoski: Not sure I understand this `@scaling_factor` -- let's chat

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - "2.0.0"
+  - 2.2.5
```
jawalonoski: I'd rather not bump the ruby version. We should be able to tie the version of `rack` and other gems to keep buy at 2.0.0

lib/modules/encounters.rb
```diff
@@ -38,7 +38,7 @@ class Encounters < Synthea::Rules
                   6.months
               end
             end
-            next_date = time + Distribution::Normal.rng(delta, delta*schedule_variance).call
+            next_date = time + Distribution::Normal.rng(delta*1.0, delta*schedule_variance).call
```
dehall: Without the 1.0 multiplier this throws an error in JRuby. Normal.rng expects a float and when it gets something else it can't find the right function to call so it errors out. I didn't try any other ways to convert to a float so there may be something more intuitive.

lib/world/sequential.rb
dehall: The intent was so that if you want to generate a population smaller than 7M but still with accurate ratios, you can scale the populations of individual cities down by this amount.

PR #20, opened 2016-08-26
Filter export to past _ years

dehall:
Exported files now contain only the past _ years (configurable) plus old data that is still relevant (active conditions, medications, careplans). Procedures are not kept if old but if they have permanent effects they should add a "History of xyz" condition, a few of which are introduced here. 

Also adds configurable options for exported file formats (fhir, ccda, html)

### DIFF
--- a/config/synthea.yml
+++ b/config/synthea.yml
@@ -1,5 +1,9 @@
 synthea:
-
+  export:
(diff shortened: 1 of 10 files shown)


jawalonoski:
Looks great.

PR #39, opened 2016-09-20
Adjust weight gain and loss and glucose estimates

jawalonoski:
Adjusted weight gain and loss to be more realistic. Folks gain weight until 49 and start losing weight at age 60. The gain/loss ranges are more dramatic than the citations because we don't account for socioeconomic status or diet yet.

The glucose estimate now gives higher numbers, so diabetes is up around 14% +- 5% per run, which is where it should roughly be for Massachusetts.

### DIFF
--- a/config/synthea.yml
+++ b/config/synthea.yml
@@ -39,11 +39,13 @@ synthea:
     weight_gain_male_stddev:    2     # kilograms per year
     weight_gain_female_average: 2.725 # kilograms per year
     weight_gain_female_stddev:  1     # kilograms per year
-    # http://www.livestrong.com/article/142567-average-increase-weight-adults/
-    adult_male_weight_gain:     0.0034 # percent increase each year (where 1.0==100%)
-    adult_female_weight_gain:   0.0052 # percent increase each year (where 1.0==100%)
-    adult_max_weight_age:       44    # age (years) when adults stop packing on weight
-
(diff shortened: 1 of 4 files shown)


config/synthea.yml
```diff
@@ -39,11 +39,13 @@ synthea:
...
     weight_gain_female_average: 2.725 # kilograms per year
     weight_gain_female_stddev:  1     # kilograms per year
-    # http://www.livestrong.com/article/142567-average-increase-weight-adults/
-    adult_male_weight_gain:     0.0034 # percent increase each year (where 1.0==100%)
-    adult_female_weight_gain:   0.0052 # percent increase each year (where 1.0==100%)
-    adult_max_weight_age:       44    # age (years) when adults stop packing on weight
-
+    # http://www.nhlbi.nih.gov/news/press-releases/2010/trials-use-technology-to-help-young-adults-achieve-healthy-weights
+    # "Previous NHLBI research has shown that people aged 18-49 gain an average of 1-2 pounds each year"
+    adult_weight_gain:          [1,2]   # kilograms per year (this is more than 1-2 pounds, but we want obese people since we don't account for other factors effecting weight-gain)
```
dehall: Just a note that because these are integers the rand() call will only choose integers 1 or 2 for the adult weight gain. If we needed intermediate values we could change these to [1.0,2.0] . The average weight in either case should be about the same though and the differences will be fractions of a pound so it's not worth changing here.

jawalonoski: Good catch. I'm going to change that.

PR #42, opened 2016-09-23
SHR Patient Identification

jawalonoski:
Enhanced Synthea output to include [Standard Health Record](http://standardhealthrecord.org) (SHR) Patient Identification Entries to the medical records -- excepting patient photograph. All of the data is exported in FHIR (some as extensions), but much of the data is not included in C-CDA because the [health-data-standards](https://github.com/projectcypress/health-data-standards) library does not include support for these fields or entries.


lib/entity/fingerprint.rb
```diff
@@ -0,0 +1,80 @@
...
+        y *= @@height
+        icon = @@template.crop(x,y,@@width,@@height)
+        # add the feature
+        fingerprint.compose!(icon,where.first,where.last)
+      end
+
+      fingerprint.resample_nearest_neighbor!(128,128)
+    end
+
+    def self.blur(image)
```
dehall: this function is never used - will we need this going forward?

lib/modules/lifecycle.rb
```diff
@@ -46,11 +48,31 @@ def initialize
...
+          mothers_surname = "#{mothers_surname}#{(mothers_surname.hash % 999)}"
+          entity[:name_mother] = "#{mothers_name} #{mothers_surname}"
 
-          # TODO update awareness
+          fathers_name = Faker::Name.first_name
+          fathers_name = "#{fathers_name}#{(fathers_name.hash % 999)}"
+          entity[:name_father] = "#{fathers_name} #{entity[:name_last]}"
+
+          #identifiers
+          entity[:identifier_ssn] = "999-#{rand(100..999)}-#{rand(1000..9999)}"
```
dehall: SSN = 9 digits so the first rand should be rand(10..99)

Gemfile.lock
```diff
@@ -23,6 +23,7 @@ PATH
   specs:
     synthea (1.0.0)
       area
+      chunky_png
```
dehall: is there a corresponding change in Gemfile?

lib/entity/fingerprint.rb
jawalonoski: Possibly. I wrote it and tested it, but it makes the fingerprints look even weirder. I think we'll need it for patient photos (to be added later) and I didn't want to have to rewrite it then.

lib/modules/lifecycle.rb
jawalonoski: I purposely used 999 for the first digits, because those are reserved, and so we'd never get a collision with a real SSN.

jawalonoski: Oh, the middles... I see.

Gemfile.lock
jawalonoski: No, the change is in the gemspec which is probably the proper place for this, since it is a runtime dependency for the gem.

lib/modules/lifecycle.rb
jawalonoski: Pushed a fix for this.

PR #43, opened 2016-09-23
Make hash on person names configurable

jamesagnew:
Just a pull request to make the hash that is added to person names configurable (default is to leave it enabled, but I don't need that for my testing)


jawalonoski:
Thanks for the contribution!

PR #44, opened 2016-09-23
adds Symptom state and symptom + priorState logic to GMF

dehall:
Adds a new state type to add a Symptom to a patient, and logic based on Symptoms and prior states.
Symptoms use the existing Symptom API.  PriorState logic checks if a specific state is in the module's state history.

For the wiki:
(state)
## Symptom

The `Symptom` state type adds or updates a patient's symptom. Synthea tracks symptoms in order to drive a patient's encounters, on a scale of 1-100. A symptom may be tracked for multiple conditions, in these cases only the highest value is considered. See also the Symptom logic. (link here)

**Supported Properties**

* **type**: must be "Symptom" _(required)_
* **symptom**: the name of the symptom being tracked _(required)_
* **cause**: the underlying cause of the symptom. Defaults to the name of the module if not set. _(optional)_
* **exact**: an exact value to score this symptom _(required if `range` is not set)_
  * **quantity**: the score to set (e.g., 4) _(required)_
* **range**: a range indicating the allowable symptom values.  The actual value will be chosen randomly from the range. _(required if `exact` is not set)_
  * **low**: the lowest number (inclusive) allowed (e.g., 5) _(required)_
  * **high**: the highest number (inclusive) allowed (e.g., 7) _(required)_

**Examples**

The following is an example of a Symptom state that sets the symptom value for Chest Pain to be exactly 27.

```json
{
  "type": "Symptom",
  "symptom" : "Chest Pain",
  "cause" : "Asthma",
  "exact": {
    "quantity": 27
  }
}
```

The following is an example of a Symptom state that sets the symptom value for Chest Pain to be between 90 and 100.
```json
{
  "type": "Symptom",
  "symptom" : "Chest Pain",
  "cause" : "Heart Attack",
  "range": {
    "low": 90,
    "high": 100
  }
}
```

(logic)

## Symptom

The `Symptom` condition type tests a patient's current symptoms. Synthea tracks symptoms in order to drive a patient's encounters, on a scale of 1-100. A symptom may be tracked for multiple conditions, in these cases only the highest value is considered. See also the Symptom state. (link here)

**Supported Properties**

* **condition_type**: must be "Symptom" _(required)_
* **symptom**: the name of the symptom to test against. _(required)_
* **operator**: indicates how to compare the current symptom score against the _value_.  Valid _operator_ values are: `<`, `<=`, `==`, `>=`, `>`, and `!=`. _(required)_
* **value**: the value to test the current symptom score against _(required)_

**Example**

The following Symptom condition will return `true` if the patient's symptom score for "Chest Pain" is greater than or equal to 50

```json
{
  "condition_type": "Symptom",
  "symptom" : "Chest Pain",
  "operator": ">=",
  "value": 50
}
```

## PriorState

The `PriorState` condition type tests the progression of the patient through the module, and checks if a specific state has already been processed (in other words, the state is in the module's state history). 

**Supported Properties**

* **condition_type**: must be "PriorState" _(required)_
* **name**: the name of the state to check for in the module's history. _(required)_

**Example**

The following PriorState condition will return `true` if the patient has already passed through a state called 'EmergencyEncounter', false otherwise.

```json
{
  "condition_type" : "PriorState",
  "name" : "EmergencyEncounter"
}
```

### DIFF
--- a/lib/generic/logic.rb
+++ b/lib/generic/logic.rb
@@ -19,6 +19,10 @@ module Synthea
           self.testDate(condition, context, time, entity)
         when 'Attribute'
           self.testAttribute(condition, context, time, entity)
+        when 'Symptom'
+          self.testSymptom(condition, context, time, entity)
+        when 'PriorState'
+          self.testPriorState(condition, context, time, entity)
         when 'True'
           self.testTrue(condition, context, time, entity)
         when 'False'
@@ -72,6 +76,14 @@ module Synthea
 
       def self.testAttribute(condition, context, time, entity)
         self.compare(entity[ condition['attribute'] ], condition['value'], condition['operator'])
+      end
+
+      def self.testSymptom(condition, context, time, entity)
+        self.compare(entity.get_symptom_value(condition['symptom']), condition['value'], condition['operator'])
+      end
+
+      def self.testPriorState(condition, context, time, entity)
+        !(context.most_recent_by_name(condition['name']).nil?)
       end
 
       def self.testTrue(condition, context, time, entity)
--- a/lib/generic/states.rb
+++ b/lib/generic/states.rb
@@ -291,6 +291,28 @@ module Synthea
         end
       end
 
+      class Symptom < State
+        def initialize (context, name)
+          super
+          cfg = context.state_config(name)
+          @symptom = cfg['symptom']
+          @cause = cfg['cause'] || context.config['name']
+          range = cfg['range']
+          exact = cfg['exact']
+          if range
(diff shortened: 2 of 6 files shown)


dehall:
Thanks good catch, I updated the doco. The wiki has 2 separate pages for States and Logic so that's why it's listed twice with the same title, the sections will be split up.   https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-States  and https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-Logic

PR #49, opened 2016-09-29
adds graphviz graph generation as a rake task

dehall:
Migrates synthea_graphviz into synthea proper as a rake task. Graphviz is required to run the new `graphviz` rake task, but not for any other tasks, i.e. `generate` and `sequential` still work without graphviz installed


jamesagnew:
Sorry for the noise from the peanut gallery, but holy crap this is cool.

PR #53, opened 2016-10-03
Adding a module for medication reconciliation

eedrummer:
This module will generate a documentation of current medications procedure at 45% of wellness encounters. Through the Meaningful Use program, ONC is striving for medication reconciliation at 50% of encounters. This assumes that we are not
at the goal level yet.

Source: https://www.healthit.gov/providers-professionals/achieve-meaningful-use/core-measures-2/medication-reconciliation

### DIFF
--- a/lib/generic/modules/med_rec.json
+++ b/lib/generic/modules/med_rec.json
@@ -1,0 +1,40 @@
+{
+  "name": "Medication Reconciliation",
+  "remarks": [
+    "This module will generate a documentation of current medications procedure at 45% of",
+    "wellness encounters. Through the Meaningful Use program, ONC is striving for",
+    "medication reconciliation at 50% of encounters. This assumes that we are not",
+    "at the goal level yet.",
+    "Source: https://www.healthit.gov/providers-professionals/achieve-meaningful-use/core-measures-2/medication-reconciliation"
+  ],
+  "states": {
+    "Initial" : {
+      "type" : "Initial",
+      "direct_transition" : "Wellness_Encounter"
+    },
+    "Wellness_Encounter" : {
+      "type" : "Encounter",
+      "wellness" : true,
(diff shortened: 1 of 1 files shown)


jawalonoski:
This doesn't appear to be working as desired. I'm seeing a medication reconciliation at every single wellness encounter... but not at emergency encounters, etc.

jawalonoski:
You might be able to fix this by having the loopback point to a delay state before looping back to the wellness encounter.... I think it just goes back to the wellness encounter continuously until it chooses the medication reconciliation. Hypothesis. Not confirmed.

dehall:
Looks like it's a side effect of how I implemented the change that prevents the same state from being added to history multiple times when there's a loop.  In this case, when the wellness encounter finally happens, it sets the `processed` field to true for the Encounter state, which isn't reset when it loops back into the state. So in the 55% of the time that it loops back, it just keeps looping until it hits the Procedure, and then a new Encounter state is created with processed=false which does wait.

There's a quick fix: just reset `processed` to false in a loop situation, but there's likely a better solution to handling these loops that's more generic. I'll keep thinking about how to do that.

eedrummer:
My initial thought when I was implementing this was to have the 55% transition to the initial state as opposed to the wellness encounter. Would that address the issue?

dehall:
Yes that would resolve the issue. Long term though we should still review how loops are handled

eedrummer:
This change **may** fix the issue, but I'm a little suspect. It is definitely generating wellness encounters where the is no documentation of medications procedure. However, on the patient set I generated, the documentation was happening 73% of the time. I'm guessing that it somehow gets "stuck" as there were a couple of patients I manually inspected where all of their encounters had meds documented. Not sure where to go from here.

jawalonoski:
Med rec needs to go back to initial too.

PR #55, opened 2016-10-04
Added Ear Infection Module

cjduffett:
Added a module to model ear infections. These primarily occur in children. For now I model treatment with either a standard antibiotic (Amoxicillin) or an over-the-counter (OTC) painkiller.

Some things that could be considered in the future:

* Other antibiotics used for treatment (e.g. Azithromycin, Augmentin)
* Recurring ear infections that eventually lead to ear damage and surgery (or hearing loss)
* Seasonal incidence (they're far more likely in the winter)

ear infections

### DIFF
--- a/lib/generic/modules/ear_infections.json
+++ b/lib/generic/modules/ear_infections.json
@@ -1,0 +1,208 @@
+{
+  "name": "Ear Infections",
+  "states": {
+
+    "Initial": {
+      "type": "Initial",
+      "direct_transition": "No_Infection"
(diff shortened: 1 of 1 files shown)


jawalonoski:
This looks good. We probably want the condition to end at the next wellness encounter. Talk to Dylan about how to make that happen.

cjduffett:
Updated the module to have a ConditionEnd after the next scheduled wellness encounter:

ear infections

PR #58, opened 2016-10-06
Added Asthma Module

cjduffett:
This module models a basic path for chronic asthma. Some things that are not included in this initial version:

* Increased likelihood of asthma from smoking or obesity
* Death from asthma (this is VERY unlikely, but possible)
* Allergy or exercise induced asthma (as-is, the cause of the occasional asthma attack is not documented, they just kind of happen)
* Additional asthma medications (beyond the standard maintenance and emergency inhalers there are also long-term steroids and nebulizers)

The granularity of the time step for asthma attacks is 6 months. Given this distribution patients could have between 0 and 2 asthma attacks per year. Most of the statistics for this module came from the CDC or the American Academy of Allergy Asthma and Immunology (AAAAI).

asthma

### DIFF
--- a/lib/generic/modules/asthma.json
+++ b/lib/generic/modules/asthma.json
@@ -194,8 +194,8 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": "185347001",
-          "display": "Encounter for problem"
+          "code": "394701000",
+          "display": "Asthma follow-up"
         }
       ],
       "conditional_transition": [
--- a/lib/generic/modules/asthma.json
+++ b/lib/generic/modules/asthma.json
@@ -129,6 +129,7 @@
 
     "Prescribe_Maintenance_Inhaler": {
         "type": "MedicationOrder",
+        "target_encounter": "Asthma_Diagnosis",
         "codes": [
           {
             "system": "RxNorm",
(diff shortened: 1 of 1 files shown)


dehall:
Looks really good. Just one thought: is there a condition or something that could indicate the asthma attack? Otherwise we just see "emergency room admission" followed by "encounter for problem" in the patient record, and it's not clear that it was an asthma attack.

cjduffett:
I went through all of the [SNOMED codes for asthma](http://www.snomedbrowser.com/Codes/ConceptList?term=asthma&type=) and nothing stood out to me. I did find a more specific code for [asthma-related ED admission](http://www.snomedbrowser.com/Codes/Details/183478001) though. Would that be a more appropriate code in place of the current ED Encounter to indicate it was asthma related?

cjduffett:
With updated SNOMED codes for the ED visit and follow-up encounter:

asthma

jawalonoski:
SNOMED codes for Asthma Condition:

195967001 "Asthma (disorder)"
370218001 "Mild asthma (disorder)"
233678006 "Childhood asthma (disorder)"

Asthma attack:

708038006 "Acute exacerbation of asthma (disorder)"

cjduffett:
Jay, would you like me to update the condition codes and make the distinction between childhood and adult asthma (as opposed to the generic "asthma" condition)?

jawalonoski:
Yes

cjduffett:
The updated module visualization:

asthma

Unless you can think of any other changes I think this is a good initial version for the asthma module.

dehall:
The MedicationOrders and the Lifelong_Asthma_Begins need to have a target_encounter set otherwise they don't do anything.

PR #60, opened 2016-10-10
add cause of death to patient record

dehall:
Adds a "Death Certificate" with "Cause of Death" to patient records in FHIR when a cause of death is known. Cause of death may be specified in `Death` state in GMF.  Updates all modules to specify a reason for death.

One note: In the case of a Death state with a specified amount of time to live, the death certificate encounter + observation are added to the patient with future dates. It's theoretically possible that the patient could "die again" of something else before reaching those future dates. I'm thinking about how to fix that & open to suggestions.

### DIFF
--- a/lib/generic/modules/dementia.json
+++ b/lib/generic/modules/dementia.json
@@ -74,6 +74,7 @@
             "code" : "26929004",
             "display" : "Alzheimer's disease (disorder)"
         }],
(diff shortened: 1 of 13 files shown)


jawalonoski:
Doesn't seem like there is anything to fix. If the patient is going to die of lung cancer in 9 months, and then they die immediately of an auto accident, then the death reason is going to be overwritten.

PR #61, opened 2016-10-11
Run simulation in parallel using JRuby

eedrummer:
**Warning** - This PR makes the assumption that the synthea code, in particular `Synthea::Rules.apply` is thread safe.

This PR allows the simulation to parallelize some of it's execution when running under JRuby. The modifications first check to see what platform the code is running on. If synthea is running in plain old ruby, this should in no way modify the behavior.

When JRuby is detected, the code will use java [Executors](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/Executor.html) to parallelize the logic. When synthea runs, it starts at a configured date, and then applies rules to all patients in the simulation (ignoring logic for births at the moment). It then advances the time and applies rules to all patients again. When running in plain old ruby, application of the rules happens in an each loop. So the rules are applied serially to all patients. Now, under JRuby, each call to apply happens in an Runnable that is passed to a fixed thread pool executor. This pool consists of 8 threads that can run the application of synthea rules. This allows the application of rules in parallel.

Testing on my mid-2015 Retina MacBook Pro shows the JRuby code taking 30 seconds to complete the simulation vs 1 minute 8 seconds for plain old ruby.

Potential improvements:

- The thread pool is created and destroyed at each time step. This is the simplest implementation, but it is likely to have better performance if the pool can be reused between time steps. I'm not sure if you would need additional logic to prevent a patient from being evaluated at two different times in parallel on separate threads
- Parallelize export

### DIFF
--- a/lib/world/population.rb
+++ b/lib/world/population.rb
@@ -1,6 +1,4 @@
-if RUBY_PLATFORM == 'java'
-  require 'java'
-end
+require 'java' if RUBY_PLATFORM == 'java'
 
 module Synthea
   module World
--- a/lib/world/population.rb
+++ b/lib/world/population.rb
@@ -1,6 +1,27 @@
+if RUBY_PLATFORM == 'java'
+  require 'java'
+end
+
 module Synthea
   module World
     class Population
+      if RUBY_PLATFORM == 'java'
+        include_package 'java.util.concurrent'
+
+        class PatientRunner
+          include java.lang.Runnable
+
+          def initialize(date, person)
+            @date = date
+            @person = person
+          end
+
+          def run
+            Synthea::Rules.apply(@date, @person)
+          end
+        end
+      end
+
       attr_reader :date, :people, :dead
 
       def initialize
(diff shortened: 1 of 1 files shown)


dehall:
Note that we do have a similar ability to run multithreaded on the sequential generation task (see https://github.com/synthetichealth/synthea/blob/master/lib/world/sequential.rb#L30 ) , but it depends on a configuration setting, when in reality it probably is clear cut that JRuby should always run multithreaded and MRI should not.  Can you tell whether using raw java Executors is at all faster than the concurrent-ruby thread pools? It seems like more direct access to the same thing. If it is faster we should probably use the same thing in sequential.

eedrummer:
@dehall I didn't know about the code in sequential and I haven't tried performance testing one against the other. The concurrent-ruby gem seems to do a nice job of hiding the plumbing for you, but there is the tradeoff that you may lose what some of the underlying tools provide, which I think is what you are getting at. I would lean towards the raw approach, since it is fairly simple for what we are doing and it allows you to minimize dependencies.

PR #66, opened 2016-10-21
Added GMF module for seasonal allergies

cjduffett:
Also known as “allergic rhinitis”. This is the first GMF module to
implement the new CarePlan states as well.

I added .palate to the gitignore - I keep a small file in my Atom IDE
called .palate with code snippets, for example the JSON for a code
(system, code, display) used in many of the GMF states.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/.DS_Store
 .vscode
+.palate
 output
 coverage
 resources
--- a/lib/generic/modules/allergic_rhinitis.json
+++ b/lib/generic/modules/allergic_rhinitis.json
@@ -126,39 +126,6 @@
           "display": "Encounter for symptom"
         }
       ],
-      "direct_transition": "Prescribe_OTC_Medication"
-    },
-
-    "Prescribe_OTC_Medication": {
-      "type": "MedicationOrder",
-      "target_encounter": "Allergic_Rhinitis_Diagnosis",
-      "remarks": [
-        "There are several antihistamines and delivery methods to choose from. Different ones work for each person.",
-        "For variety I listed oral (Claritin, and Benadryl), nasal (Flonase), and decongestants (Sudafed) medicines, ",
-        "but used the generic medicine names and dosages. Most doctors would likely recommend any combination of these."
-      ],
-      "codes": [
-        {
-          "system": "RxNorm",
-          "code": "311372",
-          "display": "Loratadine 10 MG Oral Tablet [Claritin]"
-        },
-        {
-          "system": "RxNorm",
-          "code": "1049630",
-          "display": "diphenhydrAMINE Hydrochloride 25 MG Oral Tablet [Benadryl]"
-        },
-        {
-          "system": "RxNorm",
-          "code": "896322",
-          "display": "Fluticasone propionate 0.05 MG/ACTUAT [Flonase]"
-        },
-        {
-          "system": "RxNorm",
-          "code": "1049160",
-          "display": "Pseudoephedrine Hydrochloride 30 MG Oral Tablet [Sudafed]"
-        }
-      ],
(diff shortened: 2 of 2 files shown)


cjduffett:
allergic rhinitis

lib/generic/modules/allergic_rhinitis.json
```diff
@@ -0,0 +1,351 @@
...
+
+    "Prescribe_OTC_Medication": {
+      "type": "MedicationOrder",
+      "target_encounter": "Allergic_Rhinitis_Diagnosis",
+      "remarks": [
+        "There are several antihistamines and delivery methods to choose from. Different ones work for each person.",
+        "For variety I listed oral (Claritin, and Benadryl), nasal (Flonase), and decongestants (Sudafed) medicines, ",
+        "but used the generic medicine names and dosages. Most doctors would likely recommend any combination of these."
+      ],
+      "codes": [
```
dehall: Would OTC drugs really be recorded in a patient's history?

lib/generic/modules/allergic_rhinitis.json
```diff
@@ -0,0 +1,351 @@
...
+        {
+          "system": "RxNorm",
```
dehall: Adding multiple codes is for representing the same concept in different systems, ex SNOMED and ICD-10 for a condition. In this case adding 4 codes isn't going to add multiple different medications, it's going to add one medication with 4 different codes. If we want 4 different medications prescribed it would have to be 4 separate MedicationOrder states

jawalonoski:
It looks like `Immunotherapy_Not_Given` should go to `Terminal`...

Also, I don't think that `Prescribe_OTC_Medication` is doing what you think... that is prescribing a single medication with four codes -- each for a different medication. You need to separate these medications out... if the really should be given all of them, then make them serial... if not, then give them a random distribution and converge back on `Living_With_Allergic_Rhinitis`

.gitignore
```diff
@@ -1,5 +1,6 @@
 **/.DS_Store
 .vscode
+.palate
```
dehall: what is this?

cjduffett: I keep a small text file in my Atom editor called `.palate` where I keep code snippets like the basic template for codes in the GMF:

```
{
  "system": "",
  "code": "",
  "display": ""
}
```

lib/generic/modules/allergic_rhinitis.json
cjduffett: Now that I think about it probably not. Should I substitute these for a single prescription drug in a select number of cases?

lib/generic/modules/allergic_rhinitis.json
cjduffett: See above

jawalonoski:
I'd probably put the medication order back in... sometimes doctors prescribe OTC medications, because then the patient's insurance covers the cost... depending on the cost and the insurance, the medication could then be free. So, that does happen.

cjduffett:
All of the updates are finished and tested. The final graphviz:

allergic rhinitis

lib/generic/modules/allergic_rhinitis.json
```diff
@@ -0,0 +1,376 @@
...
+      ]
+    },
+
+    "Immunotherapy_Treatment_Complete": {
+      "type": "Simple",
+      "direct_transition": "Allergic_Rhinitis_Subsides"
+    },
+
+    "Allergic_Rhinitis_Subsides": {
+      "type": "ConditionEnd",
```
dehall: When the condition ends should the medication end too?

jawalonoski: Yes, that would make sense.

cjduffett: Alright, added that.

PR #67, opened 2016-10-21
Fix geospatial extension to use existing extension on Address.

jawalonoski:
This changes our WKT geospatial extension to use the existing extension defined in FHIR:

http://hl7.org/fhir/2016Sep/extension-geolocation.html

This change will effect downstream systems that rely on our geospatial coordinates, so should be merged in coordination with changes there.


dehall:
@gregquinn2001 @cjduffett is syntheticmass currently using the patient coordinates?

PR #69, opened 2016-10-25
Use system and code for Quantity units

jamesagnew:
Observations with a value type of `Quantity` currently have their `unit` set, but not the `system` or `quantity` (which are the actual coded fields typically used by applications).

This PR corrects this.


dehall:
The only unit we're using that's a little hazy in UCUM is `mL/min/{1.73_m2}` for GFR: https://github.com/synthetichealth/synthea/blob/master/lib/records/lookup.rb#L34  . But that looks to be ok per : http://unitsofmeasure.org/trac/ticket/98

PR #72, opened 2016-10-27
Respiratory diseases

dehall:
Adds new generic modules for some of the most common respiratory diseases, for which people are likely to visit their PCP.
Specifically, Bronchitis (Acute only), Sore Throat (Viral or Strep), and Sinusitis (Acute Viral, Acute Bacterial, and Chronic)

Adds 2 new logical conditions: "At Least" and "At Most". These are grouping conditions that return true based on the number of sub-conditions that return true.

Also includes another minor framework change - complex transitions now allow for a direct transition instead of always requiring a distribution.

**For the wiki** : 
## At Least

The `At Least` condition type tests that a minimum number of conditions from set of sub-conditions are true.  If the _minimum_ number or more sub-conditions are true, it will return `true`, but if less than the minimum are true, it will return `false`. (If the _minimum_ is the same as the number of sub-conditions provided, this is equivalent to the `And` condition. If _minimum_ is 1, this is equivalent to the `Or` condition.)

**Supported Properties**
- **condition_type**: must be "And" _(required)_
- **minimum**: the minimum number of sub-conditions that must return true _(required)_
- **conditions[]**: an array of sub-conditions to test _(required)_

**Example**

The following "At Least" condition will return `true` if 2 or more of the following 3 conditions are met: 1) the patient is male, 2) the patient is at least 40 years old, 3) the patient has an active Diabetes self management plan; `false` otherwise. For example, if the patient is Male, age 46, and does not have a Diabetes self management plan, this condition will return true. If the patient is Female, age 67, and does not have a Diabetes self management plan, this condition will return false.

``` json
{
  "condition_type": "At Least",
  "minimum" : 2,
  "conditions": [
    {
      "condition_type": "Gender",
      "gender": "M" 
    },
    {
      "condition_type": "Age",
      "operator": ">=",
      "quantity": 40,
      "unit": "years"
    },
    {
      "condition_type" : "Active CarePlan",
      "codes": [{
        "system": "SNOMED-CT",
        "code": "698360004",
        "display": "Diabetes self management plan"
      }]
    }
  ]
}
```
## At Most

The `At Most` condition type tests that a maximum number of conditions from set of sub-conditions are true.  If the _maximum_ number or fewer sub-conditions are true, it will return `true`, but if more than the maximum are true, it will return `false`. 

**Example**

The following "At Most" condition will return `true` if 2 or fewer of the following 3 conditions are met: 1) the patient is male, 2) the patient is at least 40 years old, 3) the patient has an active Diabetes self management plan; `false` otherwise. For example, if the patient is Male, age 46, and does not have a Diabetes self management plan, this condition will return true. If the patient is Female, age 67, and does not have a Diabetes self management plan, this condition will return true. If the patient is Male, age 41, and has a Diabetes self management plan, this condition will return false.

``` json
{
  "condition_type": "At Most",
  "maximum" : 2,
  "conditions": [
    {
      "condition_type": "Gender",
      "gender": "M" 
    },
    {
      "condition_type": "Age",
      "operator": ">=",
      "quantity": 40,
      "unit": "years"
    },
    {
      "condition_type" : "Active CarePlan",
      "codes": [{
        "system": "SNOMED-CT",
        "code": "698360004",
        "display": "Diabetes self management plan"
      }]
    }
  ]
}
```
## Complex

`Complex` transitions are a combination of direct, distributed, and conditional transitions.  A `complex_transition` consists of an array of condition/transition pairs which are tested in the order they are defined.  The first condition that evaluates to `true` will result in a transition based on its corresponding `transition` or `distributions`.  If the module defines a `transition`, it will transition directly to that named state. If the module defines `distributions`, it will then transition to one of these according to the same rules as the `distributed_transition`. See [Distributed](#distributed) for more detail. The last element in the `complex_transition` array may omit the `condition` to indicate a fallback transition when all other conditions are `false`.

If none of the `conditions` evaluated to `true`, and no fallback transition was specified, the module will transition to a default `Terminal` state.

Please see the [[Logic|Generic Module Framework: Logic]] section for more information about creating logical conditions.

**Example**

The following example demonstrates a state that for male patients should transition to the "Foo" state with 15% probability and the "Bar" state with 85% probability, and for female patients should transition directly to the "Baz" state.

``` json
{
    "type": "Initial",
    "complex_transition": [
        {
            "condition": {
                "condition_type": "Gender",
                "gender": "M" 
            },
            "distributions" : [
                {
                    "distribution": 0.15,
                    "transition": "Foo"
                },
                {
                    "distribution": 0.85,
                    "transition": "Bar"
                }
            ]
        },
        {
            "condition": {
                "condition_type": "Gender",
                "gender": "F" 
            },
            "transition" : "Baz"
        }
    ]
}
```
# New Modules:

sore throat

sinusitis

bronchitis

### DIFF
--- a/lib/entity/entity.rb
+++ b/lib/entity/entity.rb
@@ -41,9 +41,9 @@ module Synthea
     # Symptom API
     #-----------------------------------------------------------------------
 
-    # Set value for a symptom, providing cause (ie :diabetes), type (ie :fatigue), and value ranging from 1-100
+    # Set value for a symptom, providing cause (ie :diabetes), type (ie :fatigue), and value ranging from 0-100
     def set_symptom_value(cause, type, value)
-      raise 'Symptom value out of range' if value < 1 || value > 100
+      raise 'Symptom value out of range' if value < 0 || value > 100
       @symptoms[type][cause] = value
     end
 
--- a/lib/generic/context.rb
+++ b/lib/generic/context.rb
@@ -57,14 +57,7 @@ module Synthea
           # No satisfied condition or fallback transition.  Go to the default terminal state.
           return States::Terminal.new(self, 'Terminal')
         elsif @current_state.complex_transition
-          @current_state.complex_transition.each do |ct|
-            cond = ct['condition']
-            if cond.nil? || Synthea::Generic::Logic.test(cond, self, time, entity)
-              return pick_distributed_transition(ct['distributions'])
-            end
-          end
-          # No satisfied condition or fallback transition.  Go to the default terminal state.
-          return States::Terminal.new(self, 'Terminal')
+          return pick_complex_transition(@current_state.complex_transition, time, entity)
         else
           # No transition was specified.  Go to the default terminal state.
           return States::Terminal.new(self, 'Terminal')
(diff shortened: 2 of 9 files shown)


jawalonoski:
At Most example in wiki should be "maximum" not "minimum"

jawalonoski:
No treatment for Bronchitis?

dehall:
added OTC medication for bronchitis

bronchitis

PR #74, opened 2016-11-04
GMF validator

dehall:
Re-architects some of the internals of the Generic Module Framework such that we can easily validate modules with a unit test, and not have to duplicate logic between using and validating states. 
Includes:
 - new Transition and Component modules, and updates Logic module so that conditions are their own objects. (Component includes Range, Delay, and Code classes)
 - class Metadata, and the ability to build up objects from JSON; heavily inspired by FHIR's "Hashable"
 - objects within the GMF are now true objects instead of just hashes
 - automatic validation based on required fields
 - new "generic_modules_test" to verify that all modules are valid, including all test fixtures
 - renaming "class" within Encounter states to "encounter_class"

### DIFF
--- a/lib/generic/components.rb
+++ b/lib/generic/components.rb
@@ -1,0 +1,56 @@
+module Synthea
+  module Generic
+    module Components
+      # generic "Components" that don't fit anywhere else
+      class Component
+        include Synthea::Generic::Metadata
+        include Synthea::Generic::Hashable
+
+        def initialize(hash)
+          from_hash(hash)
+        end
+      end
+
+      class Range < Component
+        attr_accessor :low, :high
+        required_field and: [:low, :high]
+
+        def value
+          rand(low..high)
+        end
+      end
+
+      class RangeWithUnit < Range
+        attr_accessor :unit
+        required_field :unit
+
+        def value
+          rand(low.send(unit)..high.send(unit))
+        end
+      end
+
+      class Exact < Component
+        attr_accessor :quantity
+        required_field :quantity
+
+        def value
+          quantity
+        end
+      end
+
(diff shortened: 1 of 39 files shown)


jawalonoski:
Having maximums at 999 seems like a hack... you'd hope there wouldn't be that many codes attached to a single state, but <sigh>

jawalonoski:
The code that checks for valid transitions should not be in a unit test, but part of the module validation itself. This will be useful for GUIs, where the module needs to be validated during editing outside of a unit test.

cjduffett:
Perhaps allow a more realistic maximum of 10 codes for a given state?

cjduffett:
Based on my conversation with @dehall; I think the validator should also test the use of attributes.

For example, if a `ConditionEnd` state has `referenced_by_attribute` set the validator should check that that attribute actually exists and is set by a respective `ConditionOnset` in one of the modules. 

I recognize that this may be difficult to validate for attributes that are used across modules (e.g. opioid addiction). Perhaps a simple solution would be to build up a global hash table of all attributes set in the modules, then validate attributes in each module against that global attribute table?

dehall:
@cjduffett I like the idea of validating the attributes, but I'm not sure it's entirely feasible because the attributes can also be set by the ruby modules (ie, metabolic_syndrome, cardiovascular_disease, etc), so it's not as easy as building up a list of the attributes mentioned in the generic modules. We could break out "errors" and "warnings" but I'm not a huge fan of that idea.

@jawalonoski  I think I covered your concerns; I changed the 999 to Float::Infinity like what FHIR has, and moved all the validations so that the unit test only calls the validate function on the Context.

PR #79, opened 2016-11-07
Updated deceased datetime in FHIR exporter

cjduffett:
Updated FHIR exporter to handle the rare case when a patient's death date exists as an event but still occurs in the future (the patient hasn't actually died yet). This is typically seen for patients with lung cancer and is a rare occurence (1-3 in 10k records). Also added the deceasedBoolean field to the FHIR records that are exported to make it easier to query FHIR for living or deceased patients using the Patient?deceased=true/false query.

### DIFF
--- a/lib/records/ccda.rb
+++ b/lib/records/ccda.rb
@@ -59,7 +59,7 @@ module Synthea
         patient.race = { 'name' => entity[:race].to_s.capitalize, 'code' => RACE_ETHNICITY_CODES[entity[:race]] }
         patient.ethnicity = { 'name' => entity[:ethnicity].to_s.capitalize, 'code' => RACE_ETHNICITY_CODES[entity[:ethnicity]] }
 
-        if !entity.alive? && entity.record_synthea.patient_info[:deathdate] <= end_time
+        if !entity.alive?(end_time)
           patient.deathdate = entity.record_synthea.patient_info[:deathdate].to_i
           patient.expired = true
           # TODO: would like to put cause of death on the record, though different IGs seem to provide different templates
--- a/lib/records/ccda.rb
+++ b/lib/records/ccda.rb
@@ -59,7 +59,7 @@ module Synthea
         patient.race = { 'name' => entity[:race].to_s.capitalize, 'code' => RACE_ETHNICITY_CODES[entity[:race]] }
         patient.ethnicity = { 'name' => entity[:ethnicity].to_s.capitalize, 'code' => RACE_ETHNICITY_CODES[entity[:ethnicity]] }
 
-        if !entity.alive?(end_time)
+        unless entity.alive?(end_time)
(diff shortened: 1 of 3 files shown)


jawalonoski:
You cannot have both `deceasedBoolean` and `deceasedDateTime` -- it is consider a choice in the FHIR. You can have one or the other, but not both.

jawalonoski:
I don't actually this this is as simple as asking if the patient is dead as of `Time.now` because you'll also have to remove the cause of death from the record, if it is present, among other things.

You probably need to pass a date (default to `Time.now` if you want) into the export -- and passed down the stack -- that will filter out all items in the record that occur after the given date.

cjduffett:
Ah, I didn't know the `[x]` properties in the spec had that rule. Seems silly to me but I'll fix it. Reference: http://hl7.org/fhir/2016Sep/formats.html#choice

cjduffett:
I followed your recommended approach above and added an `end_date`argument to the FHIR and CCDA exports (defaults to `Time.now`) that is combined with `is_alive?` to test if a patient's deceasedDateTime property should be recorded. This is also used in `filter_for_export()`.

lib/records/ccda.rb
```diff
@@ -58,7 +58,8 @@ def self.basic_info(entity, ccda_record)
         # patient.effective_time
         patient.race = { 'name' => entity[:race].to_s.capitalize, 'code' => RACE_ETHNICITY_CODES[entity[:race]] }
         patient.ethnicity = { 'name' => entity[:ethnicity].to_s.capitalize, 'code' => RACE_ETHNICITY_CODES[entity[:ethnicity]] }
-        unless entity.alive?
+
+        if !entity.alive? && entity.record_synthea.patient_info[:deathdate] <= end_time
```
jawalonoski: FYI, you should be able to just write `if !entity.alive?(end_time)`, but this is also fine.

lib/records/exporter.rb
```diff
@@ -137,7 +139,7 @@ def self.should_keep_entry(e, attribute, record, cutoff_date)
         when :careplans
           return record.careplan_active?(e['type'])
         when :conditions
-          return record.present[e['type']] || (e['end_time'] && e['end_time'] > cutoff_date)
+          return record.present[e['type']] || (e['end_time'] && e['end_time'] > cutoff_date && e['end_time'] <= end_time)
```
jawalonoski: Not sure this is correct...

This keeps the condition of it is active (present) or if the end_time is after the cut-off and before the end... Theoretically, it should keep the condition if it ended after the cut-off... even if it was in the future.

cjduffett:
Based on conversation with @jawalonoski, if an entity in the patient's record ends in the future it's end time is now removed before that entity is exported, but that entity is still kept in the patient's record. I decided to do this for MedicationOrders as well so we're consistent with how we're handling things.

PR #81, opened 2016-11-15
Pregnancy updates

cjduffett:
Updated the pregnancy and birth modules to include encounters, procedures, conditions, and care plans. Integrated the previously separate birth module into the pregnancy module.


cjduffett:
pregnancy

cjduffett:
pregnancy

cjduffett:
After testing this with 100 patients we have a replacement rate of about 1.25 births / woman. This is a little low but I only think we should adjust after we test this with 10k patients.

cjduffett:
I snuck the pregnancy module in with my run of the injuries module last night. I found the following:

**1033 births** in a 10-year period
655 unique mothers
Average of **1.58 births per mother**

5113 total living women

jawalonoski:
Check that those women are of child-bearing years before you calculate "births per woman" because the birth rate seems reasonable (1.58 per mother)

cjduffett:
I am using this definition of replacement rate:

> The total fertility rate (TFR) of a population is the average number of children that would be born to a woman over her lifetime if:  
1. She were to experience the exact current age-specific fertility rates (ASFRs) through her lifetime
2. She were to survive from birth through the end of her reproductive life

So I take this to mean the denominator for our fertility rate is _every_ living woman .

PR #82, opened 2016-11-17
Added UTI Module

cjduffett:
Created a module to model urinary tract infections. 50% of women will get at least one UTI in their lifetime, with some getting 2 or 3. In rare cases women can also get recurring UTIs that happen every 2-4 months for life. UTIs are rare for males but about 1 in 20 will get 1.

### DIFF
--- a/lib/generic/modules/urinary_tract_infections.json
+++ b/lib/generic/modules/urinary_tract_infections.json
@@ -220,6 +220,31 @@
           "display": "Phenazopyridine hydrochloride 100 MG [Pyridium]"
         }
       ],
+      "direct_transition": "UTI_CarePlan"
+    },
+
+    "UTI_CarePlan": {
+      "type": "CarePlanStart",
+      "target_encounter": "UTI_Diagnosis",
+      "codes": [
+        {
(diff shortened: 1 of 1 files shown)


cjduffett:
For reference:

urinary tract infections

dehall:
1 - the first round of antibiotics doesn't get ended if it's effective. probably want to move the distributed transition from Take_Antibiotics to End_First_Round_of_Antibiotics.
2 - recurring UTIs happen for life?
3 - I'm not a huge fan of choosing in advance how many UTIs a person will get. I know I'm guilty of doing the same thing in a couple modules but for conditions that can occur repeatedly I feel like it's defining the model too tightly to the real prevalences / "overfitting". @jawalonoski I'd like to know your thoughts on that as well.

jawalonoski:
Yeah, not a huge fan of overfitting as Dylan describes... the prevalences will change based on the region.

cjduffett:
I updated the module to probabilistically give men and women additional UTIs (rather than determining that up-front). Also added a CarePlan to the module.

cjduffett:
urinary tract infections

PR #84, opened 2016-11-18
COPD module

dehall:
Adds a module for COPD (Chronic Obstructive Pulmonary Disease, #.4 on the GBD YLL Top 10)

Changes smoking % from a fixed 16% to variable by year.

copd


cjduffett:
Looks good and the numbers check out. The only thing I would add are CarePlans for patients living with COPD, and perhaps a separate CarePlan after a lung transplant surgery. For examples of CarePlans see the injuries module. Usually 1-2 activities as part of the CarePlan is enough.

PR #85, opened 2016-11-18
Fixed the Opioid Crisis

cjduffett:
The prevalence of opioid addicts is now realistic - about 1% of the population. The updated injuries module now drives about 75% of the opioid prescriptions. The remainder are still generated by the opioids module. In general I reduced the incidence of misuse and changed the time scale of the module to 1-week steps instead of 1-day. I adjusted the relevant probabilities accordingly.

The updated graphviz with new probabilities and time step:

opioid addiction

### DIFF
--- a/lib/generic/modules/opioid_addiction.json
+++ b/lib/generic/modules/opioid_addiction.json
@@ -1,52 +1,80 @@
 {
-  "name" : "Opioid Addiction",
-  "states" : {
+  "name": "Opioid Addiction",
+  "remarks": [
+    "Opioid addiction currently affects about 0.7% of the American population, or about 2.2M people. ",
+    "Since the 1990's prescription rates for opioids have increased exponentially. Today, doctors write ",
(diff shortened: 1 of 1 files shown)


cjduffett:
The one thing I'm not wild about still is the prescription rates. Since the 1990's opioid prescriptions have increased dramatically. This is not clear in our patient records. I ran a batch of 500 patients with 10-year history and found 92 MedicationOrders for opioids. The CDC's found that over 50 - 150 prescription for opioids are written **per 100 people**, **per year**. We're not even in the same ballpark.

What we don't take into account is the number of prescriptions written per person (e.g. refills, extensions, redundant prescriptions from multiple providers), or sharing of prescriptions (e.g. Jill gives her niece Theresa some Percocet for her back pain). However, to model those things would require a heavy redesign of the opioid addiction module (and any module that prescribes opioids, for the matter) and is probably not worth the time/effort right now.

PR #86, opened 2016-11-18
Added Activities to CarePlanStart States in Graphviz

cjduffett:
Added a list of Activities prescribed by the care plan to the CarePlanStart states generated by graphviz.


cjduffett:
Before:

before

cjduffett:
After:

after

PR #90, opened 2016-11-25
Arthritis

cjduffett:
This technically wasn't in the top 10 lists, but this was a big one. Most of the module is straightforward, but as a note: gout and rheumatoid arthritis (RA) have the exact same care path. Although the state names don't match on the RA path the record still turns out correctly.

arthritis


lib/generic/modules/arthritis.json
```diff
@@ -0,0 +1,737 @@
...
+        " | Fibromyalgia   |   0.01670  |    7:1    | ",
+        " | Gout           |   0.03900  |    1:3    | ",
+        " | Rheumatoid     |   0.01000  |    1:1    | ",
+        " | Lupus          |   0.00054  |    6:1    | ",
+        " ------------------------------------------- "
+      ],
+      "direct_transition": "Delay_Until_Arthritis"
+    },
+
+    "Delay_Until_Arthritis": {
```
dehall: In this model people can only have one of these 5 conditions. Is that true in reality? I would assume there's some level of comorbidity, in particular with gout which is heavily dependent on diet, or with osteoarthritis in multiple locations.

lib/generic/modules/arthritis.json
```diff
@@ -0,0 +1,737 @@
...
+        {
+          "system": "RxNorm",
+          "code": "835900",
+          "display": "cycloSPORINE, modified 100 MG [Neoral]"
+        }
+      ],
+      "direct_transition": "Terminal"
+    },
+
+    "Terminal": {
```
dehall: None of the conditions or medications ever end. Is that accurate?

lib/generic/modules/arthritis.json
cjduffett: There is a certain likelihood of comorbidity. I think @jawalonoski's suggestion to factor these out into individual modules makes sense and solves this.

lib/generic/modules/arthritis.json
cjduffett: As far as I can tell arthritis never really ends. It just maintains or gets worse as you age. If a patient's arthritis is bad enough to merit medication I don't think the condition or the medication to treat it should ever end.

cjduffett:
I refactored the arthritis types into several different modules. I ran a test of 100 patients and the prevalences seem reasonable. Here's the graphviz diagrams for the new modules. Any feedback is welcome, I started to go blind reading through them all. :P

The most notable changes are the use of the `total_joint_replacement` module to handle joint replacements for OA and RA. Also Fibromyalgia is kinda special - it wasn't officially a disease until 1987.

## Osteoarthritis

osteoarthritis

## Rheumatoid Arthritis

rheumatoid arthritis

## Gout

gout

## Fibromyalgia

fibromyalgia

## Lupus

lupus

## Total Joint Replacement

total joint replacement

jawalonoski:
I'd like to note that most of these are missing proper diagnosis procedures... for example, gout is usually diagnosed after a microscopic analysis of urate crystals in body fluid -- which should probably be in the record. We can ignore those for now.

Please add a short-term CarePlan to the Total Joint Replacement module. It should have something about post-operative physical therapy.

Otherwise, this looks good...

cjduffett:
Added a post-op careplan for TJR:

total joint replacement

PR #91, opened 2016-11-25
Added ADHD

cjduffett:
Added childhood ADHD, treated by either behavioral therapy or medication.

attention deficit disorder

### DIFF
--- a/lib/generic/modules/attention_deficit_disorder.json
+++ b/lib/generic/modules/attention_deficit_disorder.json
@@ -1,21 +1,50 @@
 {
   "name": "Attention Deficit Disorder",
   "remarks": [
-    "The CDC estimates that 11% of children are diagnosed with ADD. ",
+    "The CDC estimates that 11% of children are diagnosed with ADHD. Althought ADHD ",
+    "is possible in adults we only model this in children. ",
+
     "Source: http://www.cdc.gov/ncbddd/adhd/data.html"
   ],
   "states": {
 
     "Initial": {
       "type": "Initial",
-      "distributed_transition": [
+      "remarks": [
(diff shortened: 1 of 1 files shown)


dehall:
could you add some logic based on date? Ritalin wasn't introduced until 1955 and Strattera wasn't around until 2000 or so. Other than that it looks good

jawalonoski:
Based on what I know, children with ADHD can often get _both_ medication and behavior therapy. Also, behavior therapy treatments repeat and can go on for years.

cjduffett:
@dehall, good catch. I'll fix that.

@jawalonoski ideally I'd like the behavior treatment to go on for maybe 2-5 years. But since patients can be diagnosed with ADHD anywhere between 2 and 15 years old I have no way to regulate how long they stay in the behavior treatment loop other than age. I think this is a notable pain area in the GMF - looping for a set time frame after a state is processed and regardless of age.

The following modules could benefit from some sort of for loop or counter-like state:

* Pregnancy (how many times women go through the process)
* Allergic Rhinitis (how long patients get immunotherapy injections for)
* ADHD (the behavior treatment loop)
* UTIs (to regulate repeat UTIs and recurrent UTIs)

I'm sure there are other use cases I haven't thought of too. Thoughts?

jawalonoski:
I suppose the way you have the therapy looping is fine for now... Although, someone on medication should also have the possibility of concurrent therapy. You should change that.

For the larger issue... we can add a maximum loop count (number of times) or maximum loop duration (will only loop for a set amount of time). Put it in the project board.

cjduffett:
Updated the medications and allowed concurrent medication and behavioral treatments:

attention deficit disorder

PR #93, opened 2016-11-29
Encounter reason

dehall:
Adds a "reason" to the encounter state in GMF, and updates existing modules with that reason.
Reason may be either a previous ConditionOnset state name or an attribute set by a ConditionOnset.

### DIFF
--- a/lib/generic/modules/allergic_rhinitis.json
+++ b/lib/generic/modules/allergic_rhinitis.json
@@ -126,6 +126,7 @@
           "display": "Encounter for symptom"
         }
       ],
+      "reason" : "has_allergic_rhinitis",
       "remarks": [
         "I select one of several over-the-counter (OTC) medications to be prescribed",
         "to the patient. These are uniformly distributed since they are all equally",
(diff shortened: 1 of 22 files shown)


cjduffett:
In your new implementation the `reason` can reference an attribute. However, in the past I've found that the `reason` for a `MedicationOrder`, `Procedure`, or `CarePlanStart` can only be a named reference to a previous `ConditionOnset` state. Is that still an issue for those states?

dehall:
MedicationOrder, Procedure, CarePlanStart, and now Encounter can all accept attributes as reasons. I'll update the wiki. At some point I'd like to extract the reason field to an object, something like: ` "reason" : { "attribute" : "attr_name" } ` or ` "reason" : { "condition_onset" : "State_Name" } `

jawalonoski:
I actually just prefer `reason` to just be a named thing... makes writing and reading these things easier. You don't need to remember whether or not `diabetes` was an attribute or ConditionOnset state.

Issue #96, opened 2016-11-29
allergic_rhinitis reason error


jeffeastman:
The Wiki requires the reason of a Procedure to be: "the name of the ConditionOnset state which represents the reason for procedure. This ConditionOnset must come before the Procedure in the module".

In this module, there are two reason references to "reason": "has_allergic_rhinitis". While there is a state "Has_Allergic_Rhinitis" it is not a ConditionOnset.

jawalonoski:
Good catch. Thanks for the bug report.

PR #97, opened 2016-11-29
Fixed reasons

cjduffett:
Swapped the code-display properties that were incorrect in the injuries module. Removed the reason attribute in the Procedure states in the allergic rhinitis module since we can't conclusively point that reason to a single ConditionOnset state. It would probably be a meaningful enhancement to allow reasons to be attributes as well.

### DIFF
--- a/lib/generic/modules/allergic_rhinitis.json
+++ b/lib/generic/modules/allergic_rhinitis.json
@@ -284,7 +284,6 @@
     "Allergy_Screening_Test": {
       "type": "Procedure",
       "target_encounter": "Immunotherapy_Consultation",
-      "reason": "has_allergic_rhinitis",
       "codes": [
         {
           "system": "SNOMED-CT",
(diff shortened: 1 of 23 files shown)


dehall:
The reason being set to an attribute should be fine - this is how it's handled in the code: https://github.com/synthetichealth/synthea/blob/master/lib/generic/states.rb#L414  The real issue described in #96 is that we're inconsistent with how "reason" is used across states, and the wiki isn't clear on that.

cjduffett:
I added a `reason` to every `Encounter`, `Procedure`, `MedicationOrder`, and `CarePlanStart` state where a reason could possibly apply. I also changed the ordering of fields in a few of the GMF states for consistency.

Transitions are now the last field in every GMF state. All remarks are now lists, for consistency. I'm going to run a JSON formatter in a separate commit to clean things up further.

cjduffett:
Ready to merge

PR #102, opened 2016-12-09
Update to FHIR v1.8 and fix #89

### LINKED ISSUE #89: MedicationOrder is not used in DSTU3 anymore
MedicationOrder is not used in DSTU3 anymore, but synthea:fhirupload still uploads that resource to STU3 FHIR server.

jawalonoski:
This updates FHIR from version 1.6 (STU3 "Baltimore") to 1.8 (STU3 "San Antonio") and fixes issue #89 .


lib/records/fhir.rb
```diff
@@ -385,39 +388,31 @@ def self.medications(prescription, fhir_record, patient, encounter)
...
+                                                'dateWritten' => convert_fhir_date_time(prescription['start_time']),
+                                                'reasonReference' => [])
         reasons.each do |r|
           med_order.reasonReference << FHIR::Reference.new('reference' => r.fullUrl.to_s)
         end
-        if prescription['stop']
-          med_order.status = 'stopped'
-
-          event = FHIR::MedicationOrder::EventHistory.new('status' => 'stopped',
-                                                          'dateTime' => convert_fhir_date_time(prescription['stop']))
```
dehall: For posterity: MedicationRequest does still have an eventHistory but it seems like the Provenance resource requires more detail than we have to provide here (I assume `agents` & `roles` in the context of ending a medication likely refer to the prescribing physician which Synthea isn't modeling yet). Would be nice to have something here but this is fine for now.

PR #107, opened 2016-12-13
Dockerize jRuby

cjduffett:
This was just for posterity. While on bare metal our multithreaded jRuby implementation is 30% faster, on Docker - even with a full allocation of CPUs and memory - the jRuby version is orders of magnitude slower. If we're going to run Synthea in Docker it better be with many parallel instances of the C Ruby version...

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,26 @@
+# Synthea .dockerignore
+# The MITRE Corporation
+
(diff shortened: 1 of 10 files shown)


jawalonoski:
As a team, we've decided to merge the standard ruby/Docker pull request and not the JRuby version.

PR #116, opened 2016-12-14
Homelessness Module

dehall:
Adds a module that represents people experiencing homelessness. Targets around 0.18% of the population being homeless at any one time. People cycle in and out of homelessness at roughly accurate rates. Currently the fact that a person is homeless is only recorded as an attribute, the person's residential address is unchanged. People that are homeless may visit shelters, at which point they are given a form and observations are taken regarding certain risk factors. These observations are intentionally simplified - eventually these risk factors should be modeled and represented directly within the patient record itself.

Also includes a couple minor tweaks: 1 - adds the ability to record an observation with the value of an attribute, and 2 - adds a config setting to disable generating fingerprints (which can speed up generation)

homelessness

### DIFF
--- a/config/synthea.yml
+++ b/config/synthea.yml
@@ -20,6 +20,7 @@ synthea:
     birth_variance: 0.01
     daily_births_per_square_mile: <%= (ma_births_per_year = 73000.0)/(ma_square_miles=10554.0)/365.0 %>
     append_hash_to_person_names: true # If true, person names have a numeric hash appended to them to make them more obviously fake
+    generate_fingerprints: true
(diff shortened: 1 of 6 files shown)


jawalonoski:
Shouldn't HIV also be an attribute? Seems weird that it could change from visit to visit.

cjduffett:
Make sure you update the `Observation` state documentation for nominal observation units/values.

dehall:
Updated HIV question so that if they ever answer HIV+ they will continue to do so going forward.

hiv_question

PR #117, opened 2016-12-15
Minor changes

jawalonoski:
Fix synthetichealth/syntheticmass#28 and add configuration option for readable filenames (not UUIDs).

### DIFF
--- a/config/synthea.yml
+++ b/config/synthea.yml
@@ -12,6 +12,7 @@ synthea:
     location: './output/'  # should either start with . for relative path or / for absolute
     folder_per_city: false
     subfolders_by_id_substring: true # if true, creates subfolders based on the first characters of the patient guid. this reduces the # of files in any single folder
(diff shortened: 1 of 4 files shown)


dehall:
My only concern is that the new filename format isn't guaranteed to be unique (especially since the numbers at the end of generated people's names is the hash of the string and not totally random) and there's nothing to handle a collision. With the name scheme last_first_age and Faker's database of ~3000 first names, ~500 last names, and ~100 possible ages ( https://github.com/stympy/faker/blob/master/lib/locales/en.yml ), we have around 150 million possible combinations, but by the birthday paradox the chance of a collision is already 50% with only 15k generated patients. </overthinking it>

jawalonoski:
Well, that is why it isn't the default (the UUID filenames are still default). 

But it is super useful option when I'm watching the output scroll by in my terminal and I see a patient with 3 or 4 co-morbidities and I want to look at someone named "John Doe"... hard to find him when everything is alphanumeric folders and UUIDs. This is mostly a debugging feature.

Issue #118, opened 2016-12-20
Injuries Timing Issue


jeffeastman:
When a patient gets a Knee_Injury, the following sequence of states can be observed:

Knee_Injury [Simple]
ED_Visit_For_Knee_Injury [Encounter]
Torn_ACL [ConditionOnset]
Wait_For_Knee_Surgery [Delay]
Knee_Surgery_Encounter [Encounter]
Knee_Surgery_Procedure [Procedure]
Knee_Injury_Prescribe_Pain_Medication [Simple]
Knee_Injury_Prescribe_Non_Opioid [MedicationOrder]
Knee_Injury_CarePlan [CarePlanStart]

The problem I see is that the Knee_Injury_CarePlan state is referencing the ED_Visit_For_Knee_Injury state. Because the Wait_For_Knee_Surgery delay advances the clock by one week, the referenced ED visit violates the CarePlanStart requirement: "The CarePlanStart state must come after a target_encounter Encounter state in the module, but must have the same start time as that Encounter; otherwise it will not be recorded in the patient's record."

In my implementation, the problem manifests itself because the subsequent End_Injury_CarePlan state cannot find the CarePlan and throws an exception.

Here's the full state trace:

Knee_Injury [Simple]
ED_Visit_For_Knee_Injury [Encounter]
Torn_ACL [ConditionOnset]
Wait_For_Knee_Surgery [Delay]
Knee_Surgery_Encounter [Encounter]
Knee_Surgery_Procedure [Procedure]
Knee_Injury_Prescribe_Pain_Medication [Simple]
Knee_Injury_Prescribe_Non_Opioid [MedicationOrder]
Knee_Injury_CarePlan [CarePlanStart]
Knee_Injury_Recovery_Period [Delay]
End_Knee_Injury [ConditionEnd]
Conclude_Injury [Simple]
End_Injury_CarePlan [CarePlanEnd]

I tried, instead, to make the Knee_Injury_CarePlan reference the Knee_Surgery_Encounter. This worked for the Torn_ACL but results in another error in the Torn_Meniscus case where there is no Delay.

Knee_Injury [Simple]
ED_Visit_For_Knee_Injury [Encounter]
Torn_Meniscus [ConditionOnset]
Knee_Injury_CarePlan [CarePlanStart]
Knee_Injury_Recovery_Period [Delay]
End_Knee_Injury [ConditionEnd]
Conclude_Injury [Simple]
End_Injury_CarePlan [CarePlanEnd]

[...]

PR #119, opened 2016-12-21
Fixed Injury Care Plans

cjduffett:
All injury care plans are now guaranteed to be concurrent with their `target_encounter`. Fixes [Issue #118](https://github.com/synthetichealth/synthea/issues/118).

### DIFF
--- a/lib/generic/modules/injuries.json
+++ b/lib/generic/modules/injuries.json
@@ -332,6 +332,33 @@
           "display": "Emergency room admission"
         }
       ],
+      "direct_transition": "Spinal_Injury_CarePlan"
+    },
+
+    "Spinal_Injury_CarePlan": {
+      "type": "CarePlanStart",
+      "target_encounter": "Spinal_Injury_Treatment_Encounter",
+      "assign_to_attribute": "injury_careplan",
+      "reason": "spinal_injury",
+      "codes": [
+        {
+          "system": "SNOMED-CT",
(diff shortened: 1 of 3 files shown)


dehall:
This failed silently and it required a close review of the module to find the issue. Does it make sense to change the logic in CarePlanStart (and ConditionOnset, MedicationOrder, and Procedure) so that it raises an exception if not concurrent with the target encounter, instead of just doing nothing? Ideally I would prefer some kind of validation that these states will always be concurrent with the target encounter, but I'm not sure where that would fit in the existing validation framework.

jeffeastman:
+1 Its hard enough to debug the modules when everything thats supposed to happen occurs. The silent omission of an element because of a timing issue makes debugging much more difficult. Due to the complexity of the flows, static analysis to detect these situations seems challenging too. I’d be interested in a discussion of static validation but in the mean time an exception would be useful to force the resolution.

Jeff

cjduffett:
Good point @dehall, I'll add that to `concurrent_with_target_encounter()`.

cjduffett:
I added two checks to the states:

1. Only `ConditionOnset` states are allowed to come before their `target_encounter` state. All other clinical states now raise an exception if the `target_encounter` is not found in the module's history when the clinical state is reached.
2. If any clinical state other than a `ConditionOnset` is not recorded because it's not concurrent with its `target encounter`, an exception is raised.

PR #120, opened 2016-12-27
migrate allergies to GMF

dehall:
Adds new states to onset and end Allergies within the GMF, and migrates the ruby module "food_allergies" to a new generic module. Note that the current model keeps conditions and allergies in the same place; eventually we may want to break out allergies into its own space. Alternatively, since the semantics are largely the same it may make more sense to keep Conditions and Allergies combined and distinguish them with a "class" or similar flag.

For the wiki:

## AllergyOnset

The `AllergyOnset` state type indicates a point in the module where the patient acquires an allergy.  This is _not_ necessarily the same as when the allergy is diagnosed and recorded in the patient's record.  In fact, it is possible for an allergy to onset but never be discovered.

If the AllergyOnset state's `target_encounter` is set to the name of a future encounter, then the allergy will be diagnosed when that future encounter occurs.  If the `target_encounter` is set to the name of a previous encounter, then the condition will only be diagnosed if the AllergyOnset start time is the same as the encounter's start time.  See the Encounter section above for more details.

### Future Implementation Considerations

Although the generic module framework supports a distinction between an allergy's onset date and diagnosis date, currently only the diagnosis date is recorded.  In the future, the `Synthea::Output::Record::condition` method should be updated to support an onset date.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `type` | `string` | Must be `"AllergyOnset"`. |
| `target_encounter` | `string` | Either an `"attribute"` or a `"State_Name"` referencing a<br/>future or concurrent `Encounter` state. |
| `assign_to_attribute` | `string` | **(optional)** The name of the `"attribute"` to assign this state to. |
| `codes` | `[]` | One or more codes that describe the Allergy. Must be valid [SNOMED codes](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-Basics#snomed-codes). |

### Example

The following is an example of an `AllergyOnset` that should be diagnosed at the `"ED_Visit"` Encounter.

```json
"Peanut_allergy": {
  "type": "AllergyOnset",
  "target_encounter": "ED_Visit",
  "codes": [
  	{
	  "system": "SNOMED-CT",
	  "code": "91935009",
	  "display": "Allergy to peanuts"
    }
  ]
}
```

## AllergyEnd

The `AllergyEnd` state type indicates a point in the module where a currently active allergy should be ended, for example if the patient's allergy subsides with time. The `AllergyEnd` state supports three ways of specifying the allergy to end:

[...]

PR #121, opened 2016-12-27
Submodules

cjduffett:
## What is This Mess?
This was a rather large endeavor and admittedly my commit history is a little messy... here's a summary of the changes I made:

1. **I added the concept of "packages" to the organization of GMF modules.** This requires that modules in the same package specify which `package` they are in:

	```json
	{
		"name": "A GMF Module",
		"package": "package_name",
		"remarks": [],
		"states": {}
	}
	```
	
	The `package` should match the name of the directory that is used to organize the modules/submodules. Submodules may only be used within their package. Two examples of packages can be found in the `fixtures/` folder. For example, the `test_package` is organized as:
	
	```
	test_package/
	  test_package.json
	  submodule.json
	```
	
	The file for the main module in the package must match the name of the package directory. That's how we know it's the main module.

2. **I added a `CallSubmodule` state that may be used to invoke execution of a submodule.** For example, a `CallSubmodule` state may look like:

	```json
	"Pregnancy_Labs": {
		"type": "CallSubmodule",
		"submodule": "Pregnancy Labs Submodule",
		"args": {
		  "encounter": "12_Week_Visit"
		},
		"direct_transition": "15_Week_Visit"
	}
	```
	
	When first processed, a `CallSubmodule` state does the following:
	
	1. Initializes a new `Synthea::Generic::Context` for the submodule.
	2. Passes the current history of the parent module to the submodule's context.
	3. Passes any arguments to the submodule's context (more on arguments later).
	4. Begins executing the submodule from it's `Initial` state _in the same time step_ as the parent module.
	5. Records the `CallSubmodule` state's exit time as the current time and adds the `CallSubmodule` state to the history.

	If and when a submodule returns (by reaching a `Terminal` state), the parent module then processes the `CallSubmodule` state a second time. This does the following:
	
	1. Copies the up-to-date history of the submodule back to its parent module.
	2. Records the `Terminal` state in the history.
	3. Continues execution of the parent module _in the same time step_ as the submodule.

3. **I added the concept of passing "arguments" to submodules.** To keep submodules and parent modules loosely coupled, parent modules may pass key:value pairs of strings to a submodule as "arguments", similar to the way arguments are passed to a function call. An argument may be used to satisfy a `reason`, `target_encounter`, `medication_order`, `condition_onset`, or `careplan` property in a submodule that needs to reference a state in the parent module's history.

	For example, if a submodule performs an observation that should be recorded at the same time as an encounter in the parent module, the parent module may pass the name of that `Encounter` state to the submodule:
	
	```json
	"Pregnancy_Labs": {
		"type": "CallSubmodule",
		"submodule": "Pregnancy Labs Submodule",
		"args": {
		  "encounter": "12_Week_Visit"
		},
		"direct_transition": "15_Week_Visit"
	}
	```
	
	And the submodule may then use it:
	
	```json
	"Blood_Test_Result": {
		"type": "Observation",
		"target_encounter": "{encounter}",
		"codes": [],
		"direct_transition": "Next_State"
	}
	```
	
	Arguments must start and end with curly braces `{` `}` and at present may only be strings. Arguments may be passed several levels deep in the context stack. For example, an argument may be passed from one submodule to another:
	
	```json
	"Call_Submodule_From_Submodule": {
		"type": "CallSubmodule",
		"submodule": "Nested Submodule",
		"args": {
			"encounter": "{encounter}"
		},
		"direct_transition": "Submodule_Terminal_State"
	}
	```
	
	At runtime the argument is evaulated and its value is obtained from the submodule's context.
	
4. **I added support for displaying a `CallSubmodule` state in graphviz.** To avoid confusion and conflicts with the dot syntax (where curly braces have special meaning), a CallSubmodule state and its arguments (if any) are displayed as follows:

submodule

5. **Graphviz output is now organized by package.** When a graphviz is generated for a package, each module/submodule is generated as a separate image file. All of these generated images are placed in a single folder that matches the name of the package.

6. **I created a new `ContextRunner` object that manages a context stack.** Each time a `CallSubmodule` state is reached a new context is pushed onto the stack. When a submodule returns (by reaching a `Terminal` state), its context is popped from the stack. During each time step, the `ContextRunner` simply executes whatever context is on the top of the stack. An empty stack means the module or package is finished for the current entity.

7. **I moved logging responsibilities from `Context` to `ContextRunner`.** The most up-to-date history in a package is always in the context at the top of the stack. This history may be a composition of the history from several different contexts that are each partially executed. Therefore a single context cannot be responsible for logging the history of the _entire_ package.

8. **Unit tests galore.** I added comprehensive tests for each of the new items above.

## Why Bother With All of This?

Our GMF modules are growing increasingly complex and becoming less DRY. The use of submodules is intended to introduce some DRY-ness and organization into the GMF. My "Top 3" existing areas that I think will benefit from submodules:

* **Repeating the Same Sequence of States.** In the near future we need to add labs to the pregnancy module. These are the same or similar labs, performed at each visit to an OBGYN, 20-30 times during the course of a pregnancy. The only thing that varies each time is the target encounter that the labs are performed at. This screams DRY!

* **Complex Prescriptions.** What medications are available in the current year of the simulation? What doses should be used for adults? Children? What alternative drugs can be prescribed if the patient has a drug allergy? This complexity can be expressed in the GMF but will require dozens of states that realistically do not belong in the main flow of the module.

* **Massive Modules.** Think injuries. While the overall incidence of injuries can be handled by a single parent module, each injury could be placed in its own submodule. That way we could have a collection of 10-20 injury-specific submodules in a larger `injuries` package. Easier to read, easier to build, easier to maintain.

## Any Limitations?

**Validation**. Currently we validate modules on an individual basis. Since packages span across several modules, and may pass key pieces of context using arguments, it is impossible to statically validate each module in a package by itself. We'll need to beef up the validation framework to validate packages.

**Added complexity**. There is a fair amount of overhead that goes into managing multiple contexts and calls to submodules. This makes it a little harder for others to implement GMF interpreters, but not by much. Hopefully Jeff doesn't mind.


dehall:
I'm definitely in favor of the concept of submodules but I'm not sure about this implementation. My biggest concern is the additional complexity this adds to running modules. Context is already the most complicated class and this adds another level on top of that. 

The concept of passing arguments is interesting, but it feels too limited in what it can do.  Take a look at the Allergies module: #120   The main core of the module is 3 states repeated 7 times with different values. With a more powerful system this could be represented a submodule of those 3 different states, and the different codes and percentages passed in as parameters. But I realize something like that can get too complex for people to read and follow so I guess there are tradeoffs.

The use of curly brackets around the name of an argument feels like a hack instead of using more traditional OO methods. Maybe this isn't the will of the team but I'm in favor of being more strict and explicit in the models rather than having 1 field be potentially multiple things which all have to be checked at runtime. Instead of `"reason" : "xyz"` where xyz could be a state name, attribute, or argument,  I'd much rather see `"reason" : { "state" : "xyz" }` or `"reason" : { "attribute" : "xyz" }` or `"reason" : { "argument" : "xyz" }` . The author has to know which one it is when they write it, why not save some processing time, both for the software itself and for later readers when they try to figure out what they're looking at?

jeffeastman:
+1 I think the concept is very useful and I’m experimenting with an implementation. The idea of passing (binding) states from a parent module into a subModule makes sense to me, and Codes would be useful too. The curly brackets mechanism for denoting an argument is ok with me until we gain more experience with subModules. Its nice to be OO but Ruby’s ability to dynamically add fields&methods to objects makes my static Java world a little more challenging.

I’ll add more thoughts as I wade into this a bit more.

Jeff

jeffeastman:
I think its working for this limited test case. In this test, the first Encounter has the ConditionOnset and the MedicationOrder. From the trace, you can see that the subModule waits for the next wellness encounter (patient age = 0 so its not long). Context isn’t too hairy in my implementation as I’m just cloning the modules. 

Need more test data!
Jeff

0 TestPerson [0] Initial [Initial]
2016/12/29 11:50 TestPerson [0] Encounter [Encounter]
1 TestPerson [0] Encounter [Encounter]
  TestEncounterEvent [name=Encounter]
2017/01/05 10:50 TestPerson [0] Examplitis [ConditionOnset]
2017/01/05 10:50 TestPerson [0] CallSubmodule [CallSubroutine]
2 TestPerson [0] Initial [Initial]
2017/01/05 10:50 TestPerson [0] MedicationOrder [MedicationOrder]
  TestMedication [name=MedicationOrder]
2017/01/05 11:50 TestPerson [0] Wellness [Encounter]
3 TestPerson [0] Wellness [Encounter]
2017/01/28 20:50 TestPerson [0] Sub_Terminal [Terminal]
2017/01/28 20:50 TestPerson [0] Main_Terminal [Terminal]

jeffeastman:
Here’s a trace of the recursive test package. A little trickier but overall this still seems like a reasonable capability. It will be a huge effort to refactor the other modules to employ submodules, but they can be enhanced as needed for additional functionality. The lines beginning with small integers record the number of iterations the test has done. I jump the simulation date forward around the delays so you can see when the various events are recorded they have occurred at points in time past the delays.

Jeff

1 TestPerson [0] Initial [Initial]
2016/12/29 14:48 TestPerson [0] Example_Condition [ConditionOnset]
2016/12/29 14:48 TestPerson [0] CallSubmodule [submodule=Encounter Submodule]
2 TestPerson [0] Initial [Initial]
2017/12/29 14:48 TestPerson [0] Delay [Delay]
3 TestPerson [0] Delay [Delay]
2017/12/29 15:48 TestPerson [0] Encounter_In_Submodule [Encounter]
  TestEncounterEvent [name=Encounter_In_Submodule]
2017/12/29 15:48 TestPerson [0] CallSubmodule [submodule=MedicationOrder Submodule]
5 TestPerson [0] Initial [Initial]
2017/12/29 15:48 TestPerson [0] Examplitis_Medication [MedicationOrder]
  TestMedication [name=Examplitis_Medication]
2018/01/12 15:48 TestPerson [0] Delay_Yet_Again [Delay]
6 TestPerson [0] Delay_Yet_Again [Delay]
2018/01/12 15:48 TestPerson [0] End_Medication [MedicationEnd]
2018/01/12 15:48 TestPerson [0] Med_Terminal [Terminal]
2018/02/23 14:48 TestPerson [0] Delay_Some_More [Delay]
7 TestPerson [0] Delay_Some_More [Delay]
2018/02/23 14:48 TestPerson [0] Encounter_Terminal [Terminal]
2018/02/23 14:48 TestPerson [0] End_Condition [ConditionEnd]
2018/02/23 14:48 TestPerson [0] Terminal [Terminal]

Encounter 0 Events:
TestCondition [name=Example_Condition]
TestEncounterEvent [name=Encounter_In_Submodule]
TestMedication [name=Examplitis_Medication]

cjduffett:
I removed `args` from the `CallSubmodule` state. In 99% of meaningful use cases the `args` were just used to pass the name of the most recent `Encounter` state for a clinical state in the submodule to use as a `target_encounter`.

Now, Synthea instead infers the correct `target_encounter` for a clinical state by checking to see if an encounter was recently processed and is concurrent with the clinical state in question. This encounter is known as the `current_encounter` in each module's context. This makes `target_encounter` optional (and in fact, unused) for all clinical states except `ConditionOnset` and `AllergyOnset`. In a future release we should probably remove the `target_encounter` property from those states. For now, however, I left `target_encounter` as an optional property to maintain backwards compatibility with our existing modules.

In some cases, `target_encounter` can also be optional for `OnsetStates` if they directly follow an `Encounter` state. In practice this rarely occurs though. In all cases, if a `current_encounter` cannot be found for a clinical state an exception is raised.

jeffeastman:
+1 This does seem reasonable. Its only when ConditionOnset and AllergyOnset need to refer to future states that target_encounter is needed as you’ve noted. And, by eliminating the args on the Call_Submodule state it cleans up the {encounter} hygiene issues. Other arguments can be passed via attributes if they are needed. The changes to my implementation were minimal.

Jeff

lib/generic/states.rb
```diff
@@ -211,9 +240,14 @@ class Encounter < State
 
         metadata 'codes', type: 'Components::Code', min: 1, max: Float::INFINITY
 
+        def initialize(context, name)
+          super(context, name)
```
dehall: a constructor that just calls super with the same args doesn't need to be explicitly defined

lib/tasks/graphviz.rb
```diff
@@ -411,6 +442,14 @@ def self.find_referenced_type(logic)
         end
       end
 
+      def self.escape_arg(arg)
```
dehall: all this escape_arg stuff can be removed now

lib/generic/context_runner.rb
```diff
@@ -0,0 +1,103 @@
+module Synthea
```
dehall: Is there any way to simplify the code in the ContextRunner / Context / Package trio? It works but it feels like it's more complicated than it has to be. Maybe there's no clean way to do it without carving out massive changes but I'm going to keep thinking about that.

cjduffett: What in particular is still bothering you?

lib/tasks/graphviz.rb
cjduffett: Good catch, removed those references.

lib/generic/states.rb
cjduffett: Thanks for pointing this out too - removed those cases.

test/unit/generic_states_test.rb
```diff
@@ -286,59 +286,6 @@ def test_condition_onset_during_encounter
     @patient.record_synthea.verify
   end
 
-  def test_allergy_onset
```
cjduffett: Just noticed that these got left out when I did the rebase. I'll make sure I add them back in with the other fixes.

lib/generic/context_runner.rb
dehall: History is shared among all the modules in a stack. The new current_encounter is shared among all the modules in a stack. The only thing that really has to change when you call a submodule is that the current state and "module" have to be saved off in a stack somewhere.  There shouldn't have to be special logic in the `run` function of the Context/ContextRunner that checks if the last state is a CallSubmodule - there should be a new load_submodule function in the Context that gets called from CallSubmodule states. (and there can probably be something for Terminal too) Then the ContextRunner becomes unnecessary. Even though there's a 1-1 mapping between Context-Module right now there doesn't have to be. "Context" really just means the environment where things are happening so it's still the appropriate term if it encapsulates the whole Module-Submodule relationship.

And package is just a folder, we don't need a whole separate class to represent a folder. Also I don't see why modules have to specify what package they are in. The folder name is the package name, why does it have to be specified again within the file itself? And if the "main" module in a package was instead in the "modules" folder instead of the package folder, there doesn't have to be a distinction between a "module" and a "package" at all.

cjduffett:
## That's All Folks!

@dehall, thanks for pushing me to simplify and revise this. I think you'll be happy with the final result. In summary:

1. I squashed all commits into one, comprehensive commit. With all of the iterations my commit history got **ugly**.

2. I removed the `Package` and `ContextRunner` classes and placed everything under `Context`. A lot of redundant stuff disappeared in the process.

3. The `CallSubmodule` state now uses `Context.call_submodule()` to invoke a submodule and continue processing states.

4. The `Terminal` state now uses `Context.return_from_submodule()` to terminate a submodule and continue processing states in the parent module.

5. I created a global module lookup, `Synthea::MODULES`, that all modules/submodules are loaded into before execution begins. Since submodules are frequently reused, this ensures that each and every module and submodule is loaded only _once_.

6. A `Context` is now initialized using a key to a module in `Synthea::MODULES`. For example:
    ```ruby
    ctx = Synthea::Generic::Context.new('allergies')
    ```
    Where the `allergies` module is already loaded into `Synthea::MODULES['allergies']`. 

7. I had to disable static validation of references to states.

## Static Validation Issues
With the introduction of submodules, modules may now reference states that are not necessarily in the same file. Since we statically validate only one module file at a time, our approach for validating those references no longer works.

However, we are still able to validate:

1. That every state in the module is reachable
2. That every state's required fields are satisfied

This still holds true for every module and submodule.

## Future Considerations

In the near future (but in separate PRs) I recommend we also do the following:

#### Upgrade static validation of modules to support submodules
In my opinion, any module that uses submodules should also have those submodules validated with it. For example, if the `ear_infections` module uses the `otc_pain_reliever` submodule and the `injuries` module _also_ uses the `otc_pain_reliever` submodule, each pair should be validated separately. This means the `otc_pain_reliever` module may be validated repeatedly, but in different contexts.

#### Upgrade graphviz to support submodules
Graphviz should make a distinction between modules and submodules and group them appropriately in separate directories. As it stands now, all modules/submodules are rendered and placed in the `output` directory.

#### Use submodules!
For starters, I think a few medication submodules would be a good first step. I think common, shared submodules for over-the-counter pain relievers, over-the-counter antihistamines, opioid pain relievers, and antibiotics would be immediately useful.

PR #125, opened 2017-01-04
GMF Vital Signs

dehall:
Adds Vital Signs to the Generic Module Framework. Vital Sign states can be used to store information which can then be captured by Observations or drive conditional logic. Vital Signs represent the physical state of the patient, in contrast to Observations which are the recording of that physical state. For instance, height and weight are physical properties of a person which can be observed. Note that not all observations are direct recording of physical properties, so Observations can still specify a range or exact value. For instance, MMSE (Mini–Mental State Examination, a diagnostic evaluation used for tracking dementia, ex Alzheimer's) is not a physical quantity but a score representing an evaluation. There may not be a clear distinction between the Vital Sign and Observation in all cases, so I leave it to the author to decide which is more appropriate. In general, the Vital Sign should be used if the value directly affects the patient's physical condition. For example, high blood pressure directly increases the risk of heart attack so any conditional logic that would trigger a heart attack should consider the Vital Sign instead of the Observation. On the other hand, if the value affects the patient's care, the Observation would be more appropriate. For example, it is the Observation of high blood pressure that leads a doctor to prescribe medication, not the high blood pressure itself.

Also added DiagnosticReport and MultiObservation states to group observations.

Also upgraded the PriorState logic to allow checking within a certain timeframe or since a previous state. For example, you may want to check whether a patient has had a certain test within the past 5 years.

Adds a "Wellness Encounters" generic module that records various observations at each wellness encounter. This replaces the recording of observations from the ruby modules. See graphviz below

For the Wiki:
# States
## VitalSign
The `VitalSign` state type indicates a point in the module where a patient's vital sign is set. Vital Signs represent the physical state of the patient, in contrast to Observations which are the recording of that physical state.

### Notes
There may not be a clear distinction between the Vital Sign and Observation in all cases, so it is left to the module author to decide which is more appropriate. In general, the Vital Sign should be used if the value directly affects the patient's physical condition. For example, high blood pressure directly increases the risk of heart attack so any conditional logic that would trigger a heart attack should reference a Vital Sign instead of an Observation. On the other hand, if the value affects the patient's care, the Observation would be more appropriate. For example, it is the Observation of high blood pressure that leads a doctor to prescribe medication, not the high blood pressure itself.

### Supported Properties
| Attribute | Type | Description |
|:----------|:-----|:------------|
| `type` | `string` | Must be `"VitalSign"`. |
| `vital_sign`| `string` | The name of the Vital Sign, which may be referenced later
| `unit` | `string` | The unit of measure in which the vital sign is measured (e.g. `"cm"`). |

And one of: 

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `exact` | `{}` | The exact value to be recorded for this Observation |
| `range` | `{}` | A range of values from which one value will be recorded |

##### `exact`:

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `quantity` | `numeric` | The exact value of `unit` observed. |

##### `range`:

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `low` | `numeric` | The lowest (inclusive) numeric value of `unit` observed. |
| `high` | `numeric` | The greatest (inclusive) numeric value of `unit` observed. |

### Example
The following is an example of a VitalSign state that sets the patient's Systolic Blood Pressure to 120 mmHg:

```
"SetBP" : {
   "type" : "VitalSign",
   "vital_sign" : "Systolic Blood Pressure",
   "exact" : { "quantity" : "120" },
   "unit" : "mmHg"
},
```

## Observation
The `Observation` state type indicates a point in the module where an observation is recorded. Observations include clinical findings, vital signs, lab tests, etc.

If the Observation state's `target_encounter` is set to the name of a future encounter, then the observation will be recorded when that future encounter occurs.  If the `target_encounter` is set to the name of a previous encounter, then the observation will only be recorded if the Observation start time is the _same_ as the encounter's start time.  See the Encounter section above for more details.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `type` | `string` | Must be `"Observation"`. |
| `target_encounter` | `string` | Either an `"attribute"` or a `"State_Name"` referencing a concurrent or future `Encounter` state. |
| `unit` | `string` | The unit of measure in which the observation is recorded (e.g. `"cm"`). |
| `codes` | `[]` | One or more codes that describe the Observation. Must be valid [LOINC codes](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-Basics#loinc-codes). |

And one of: 

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `exact` | `{}` | The exact value to be recorded for this Observation |
| `range` | `{}` | A range of values from which one value will be recorded |
| `attribute` | `string` | The name of an attribute that contains the value of some Observation |
| `vital_sign`| `string` | The name of a Vital Sign that contains a value

##### `exact`:

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `quantity` | `numeric` | The exact value of `unit` observed. |

##### `range`:

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `low` | `numeric` | The lowest (inclusive) numeric value of `unit` observed. |
| `high` | `numeric` | The greatest (inclusive) numeric value of `unit` observed. |

The actual value for the observation will be chosen randomly from this range.

### Example

The following is an example of an Observation that should be taken at the `"Checkup"` Encounter, that will result in an observation of body height between 40 and 60 cm:

```json
"Height_Measurement": {
  "type": "Observation",
  "target_encounter": "Checkup",
  "unit": "cm",
  "range": {
    "low": 40,
    "high": 60
  },
  "codes": [
    {
      "system": "LOINC",
      "code": "8302-2",
      "display": "Body Height"
    }
  ]
}
```

## MultiObservation
The `MultiObservation` state indicates that some number of prior Observation states should be grouped together as a single observation. This can be necessary when one observation records multiple values, for example in the case of Blood Pressure, which is really 2 values, Systolic and Diastolic Blood Pressure. This state must occur directly after the relevant Observation states, otherwise unexpected behavior can occur.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `type` | `string` | Must be `"MultiObservation"`. |
| `target_encounter` | `string` | Either an `"attribute"` or a `"State_Name"` referencing a concurrent or future `Encounter` state. |
| `number_of_observations` | `integer` | The number of observations to group within this `MultiObservation`
| `codes` | `[]` | One or more codes that describe the Observation. Must be valid [LOINC codes](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-Basics#loinc-codes). |

### Example
The following example shows a `MultiObservation` which groups the 2 previous observations into 1 observation for Blood Pressure:

```
"Record_BP" : {
   "type" : "MultiObservation",
   "number_of_observations" : 2,
   "codes" : [{
     "system" : "LOINC",
     "code" : "55284-4",
     "display" : "Blood Pressure"
   }],
   "target_encounter" : "Wellness_Encounter"
 }
```

## DiagnosticReport
The `DiagnosticReport` state indicates that some number of prior Observation states should be grouped together within a single Diagnostic Report. This can be used when multiple observations are part of a single panel.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `type` | `string` | Must be `"DiagnosticReport"`. |
| `target_encounter` | `string` | Either an `"attribute"` or a `"State_Name"` referencing a concurrent or future `Encounter` state. |
| `number_of_observations` | `integer` | The number of observations to group within this `DiagnosticReport`
| `codes` | `[]` | One or more codes that describe the DiagnosticReport. Must be valid [LOINC codes](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-Basics#loinc-codes). |

### Example
The following example shows a `DiagnosticReport` which groups the 8 previous observations into a Metabolic Panel Diagnostic Report:

```
"Record_MetabolicPanel" : {
   "type" : "DiagnosticReport",
   "number_of_observations" : 8,
   "codes" : [{
     "system" : "LOINC",
     "code" : "51990-0",
     "display" : "Basic Metabolic Panel"
   }],
   "target_encounter" : "Wellness_Encounter"
 }
```

# Logic
## Vital Sign
The `Vital Sign` condition type tests a patient's current vital signs. Synthea tracks vital signs in order to drive a patient's physical condition, and are recorded in observations. See also the [[Symptom|Generic Module Framework: States#VitalSign]] state.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `condition_type` | `string` | Must be `"Vital Sign"`. |
| `vital_sign` | `string` | The name of the vital sign to test against. |
| `operator` | `string` | One of [`"=="`, `"!="`, `"<"`, `">"`, `"<="`, `">="`]. |
| `value` | `numeric` | The value to test the vital sign against. |

### Example

The following `Vital Sign` condition will return `true` if the patient's Systolic Blood Pressure vital sign is greater than 120:

```json
{
    "condition_type" : "Vital Sign",
    "vital_sign" : "Systolic Blood Pressure",
    "operator" : ">",
    "value" : 120
  },
```

## PriorState

The `PriorState` condition type tests the progression of the patient through the module, and checks if a specific state has already been processed (in other words, the state is in the module's state history). The search for the state may be limited by time or the name of another state.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `condition_type` | `string` | Must be `"PriorState"`. |
| `name` | `string` | The name of the state to check for in the module's history. |
| `since` | `string` | The name of a state at which this logic will stop checking for the target state.
| `within` | `{}` | An exact timespan to use to limit the search for the target state

### Example

The following `PriorState` condition will return `true` if the patient has already passed through a state called `"Emergency_Encounter"`, `false` otherwise:

```json
{
  "condition_type": "PriorState",
  "name": "Emergency_Encounter"
}
```

The following `PriorState` condition will return `true` if the patient has passed through a state called `"Emergency_Encounter"` within the last 3 years, or `false` if the patient has never passed through the state, or last passed through the state more than 3 years ago:

```json
{
  "condition_type": "PriorState",
  "name": "Emergency_Encounter",
  "within": { "quantity": 3, "unit" : "years"}
}
```

The following `PriorState` condition will return `true` if the patient has already passed through a state called `"Emergency_Encounter"` since last passing through a state called "Followup", `false` otherwise. (In other words, if the patient hit state "Emergency_Encounter" more recently than state "Followup". Note that it is not necessary for the state "Followup" to have been hit for this to return true.)

```json
{
  "condition_type": "PriorState",
  "name": "Emergency_Encounter",
  "since" : "Followup"
}
```

wellness encounters

### DIFF
--- a/lib/entity/entity.rb
+++ b/lib/entity/entity.rb
@@ -4,7 +4,7 @@ module Synthea
     attr_accessor :events
 
     def initialize
-      @attributes = {}
+      @attributes = { vital_signs: {} }
       @events = Synthea::EventList.new
 
       # We represent symptoms as a hash of hashes; the first level key is the symptom name (ie :fatigue), the
@@ -35,6 +35,20 @@ module Synthea
 
     def alive?(time = nil)
       event(:birth) && !had_event?(:death, time)
+    end
+
+    def vital_sign(vital_sign_name)
+      vital_sign_name = vital_sign_name.to_s.gsub(/\s+/, '_').downcase.to_sym
+      @attributes[:vital_signs][vital_sign_name]
+    end
+
+    def get_vital_sign_value(vital_sign_name)
+      vital_sign(vital_sign_name)[:value]
+    end
+
+    def set_vital_sign(vital_sign_name, value, unit)
+      vital_sign_name = vital_sign_name.to_s.gsub(/\s+/, '_').downcase.to_sym
+      @attributes[:vital_signs][vital_sign_name] = { value: value, unit: unit }
     end
 
     #-----------------------------------------------------------------------
--- a/lib/generic/logic.rb
+++ b/lib/generic/logic.rb
@@ -192,6 +192,16 @@ module Synthea
         end
       end
 
+      class VitalSign < Condition
+        attr_accessor :vital_sign, :value, :operator
+        required_field and: [:vital_sign, :operator] # value is allowed to be omitted if operator is 'is nil'
+
+        def test(_context, _time, entity)
+          vs = entity.vital_sign(vital_sign) || {}
+          compare(vs[:value], value, operator)
+        end
+      end
+
       class ActiveCondition < Condition
         attr_accessor :codes, :referenced_by_attribute
         required_field or: [:codes, :referenced_by_attribute]
@@ -217,13 +227,20 @@ module Synthea
       end
 
       class PriorState < Condition
-        attr_accessor :name
+        attr_accessor :name, :since, :within
         required_field :name
 
         metadata 'name', reference_to_state_type: 'State', min: 1, max: 1
+        metadata 'since', reference_to_state_type: 'State', min: 0, max: 1
+        metadata 'within', type: 'Components::ExactWithUnit', min: 0, max: 1
 
-        def test(context, _time, _entity)
-          !context.most_recent_by_name(name).nil?
+        def test(context, time, _entity)
+          context.history.reverse_each do |h|
+            return false if within && h.exited && h.exited < (time - within.value)
+            return true if h.name == name
+            return false if h.name == since
+          end
+          false
         end
       end
 
--- a/lib/generic/modules/wellness_encounters.json
+++ b/lib/generic/modules/wellness_encounters.json
@@ -1,0 +1,462 @@
+{
+  "name": "Wellness Encounters",
+  "remarks": [
+    "migrated observations from Lifecycle and Metabolic Syndrome ruby modules"
+  ],
+  "states": {
+
+    "Initial" : {
+      "type" : "Initial",
+      "direct_transition" : "Wellness_Encounter"
+    },
+
+    "Wellness_Encounter" : {
+      "type" : "Encounter",
+      "wellness" : true,
+      "direct_transition" : "Record_Height"
+    },
+
+    "Record_Height" : {
+      "type" : "Observation",
+      "target_encounter" : "Wellness_Encounter",
+      "vital_sign" : "Height",
+      "codes" : [{
+        "system" : "LOINC",
+        "code" : "8302-2",
+        "display" : "Body Height"
+      }],
+      "unit" : "cm",
+      "direct_transition" : "Record_Weight"
+    },
+
+    "Record_Weight" : {
+      "type" : "Observation",
+      "target_encounter" : "Wellness_Encounter",
+      "vital_sign" : "Weight",
+      "codes" : [{
+        "system" : "LOINC",
+        "code" : "29463-7",
+        "display" : "Body Weight"
+      }],
+      "unit" : "kg",
+      "conditional_transition" : [
+        {
+          "condition" :{
+            "condition_type": "Age",
+            "operator": "<",
+            "quantity": 2,
+            "unit": "years"
+          },
+          "remarks" : ["BMI is not particularly meaningful for children under 2"],
+          "transition" : "Record_Systolic_BP"
+        },
+        {
+          "transition" : "Record_BMI"
+        }
+      ]
+    },
+
+    "Record_BMI" : {
+      "type" : "Observation",
+      "target_encounter" : "Wellness_Encounter",
+      "vital_sign" : "BMI",
+      "codes" : [{
+        "system" : "LOINC",
+        "code" : "39156-5",
+        "display" : "Body Mass Index"
+      }],
+      "unit" : "kg/m2",
+      "direct_transition" : "Record_Systolic_BP"
+    },
+
+    "Record_Systolic_BP" : {
+      "type" : "Observation",
+      "target_encounter" : "Wellness_Encounter",
+      "vital_sign" : "Systolic Blood Pressure",
+      "codes" : [{
+        "system" : "LOINC",
+        "code" : "8480-6",
+        "display" : "Systolic Blood Pressure"
+      }],
+      "unit" : "mmHg",
+      "direct_transition" : "Record_Diastolic_BP"
+    },
+
+    "Record_Diastolic_BP" : {
+      "type" : "Observation",
+      "target_encounter" : "Wellness_Encounter",
+      "vital_sign" : "Diastolic Blood Pressure",
+      "codes" : [{
+        "system" : "LOINC",
+        "code" : "8462-4",
+        "display" : "Diastolic Blood Pressure"
+      }],
+      "unit" : "mmHg",
+      "direct_transition" : "Record_BP"
+    },
+
+    "Record_BP" : {
+      "type" : "MultiObservation",
+      "number_of_observations" : 2,
+      "codes" : [{
+        "system" : "LOINC",
+        "code" : "55284-4",
+        "display" : "Blood Pressure"
+      }],
+      "target_encounter" : "Wellness_Encounter",
+      "direct_transition" : "Lab_1_HA1C"
+    },
+
+    "Lab_1_HA1C" : {
+      "type" : "Simple",
+      "conditional_transition" : [
+        {
+          "condition" : {
+            "condition_type" : "Or",
+            "conditions" : [
+              {
+                "condition_type": "Attribute",
+                "attribute": "diabetes",
+                "operator": "is not nil"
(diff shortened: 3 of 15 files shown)


jawalonoski:
Need to fix: "The value between 0 and 100 to test the symptom against." The value can be anything, even negative.

jawalonoski:
Not sure I like multi observations and diagnostic reports... possibly too much FHIR bleeding through into the GMF.

Also, I'm unconvinced on VitalSigns. Why not just use attributes?

jawalonoski:
JSON error (missing `:` before `3`):

`"within": { "quantity" 3, "unit" : "years"}`

dehall:
These are really just enhanced attributes, so I think the value of having the VitalSign state + logic vs just using pure attributes is that the module author's intent is more explicitly expressed. To me the most important thing is having a more clear distinction between patient and record. If you're against having the Vital Sign state & logic I can rework this to use plain attributes.


Re: MultiObservations & DiagnosticReports - I personally would have preferred a new field on Observation that would indicate the name of a grouping, so that observations with the same group name & time would get grouped automatically. But that seemed much more difficult to implement so I went with this. Another option would be to just use a single "ObservationGroup" state instead of the straight-FHIR states.

cjduffett:
The distinction between `VitalSign` and `Observation` makes sense to me. I think @dehall's description makes this clear enough. I also recognize why vital signs really can't be stored as attributes (at least, as we represent attributes now). They must encapsulate both a quantity and a unit describing the vital sign, which is more than one "value".

But considering @jawalonoski's point - if you want to store vital signs as attributes, maybe we can just beef up the attribute API to allow us to store hashes? So a vital sign could be stored as:

```json
"Set_Systolic_Blood_Pressure": {
  "type": "SetAttribute",
  "attribute": "systolic_bp",
  "value": {
    "quantity": 120,
    "unit": "mmHg"
  }
}
```

This would of course make conditions comparing attributes more challenging since it would be unclear which field in the hash to use for the comparison. Perhaps you could require the condition to also specify a key if the attribute's value is a hash? e.g:

```json
{
  "condition_type": "Attribute",
  "attribute": "systolic_bp",
  "key": "quantity",
  "operator": ">",
  "value": 120
}
```

To keep things simple we would not allow hash attributes to be nested. Thoughts?

***

I agree with @jawalonoski that `MultiObservation` and `DiagnosticReport` seem too FHIR-ish. If they both just simply group a set of Observations, why are there two different ways to do that? I'm not clearly seeing that distinction.

Also why do we need to group the observations at all? I don't see any changes to the way exports are done, and I can't think of any other reason why we'd bother grouping them.

dehall:
The DiagnosticReport and MultiObservation are just direct translations of these 2 calls from the ruby modules:

https://github.com/synthetichealth/synthea/blob/master/lib/modules/metabolic_syndrome.rb#L491
https://github.com/synthetichealth/synthea/blob/master/lib/modules/metabolic_syndrome.rb#L511

MultiObservation groups 2 or more observations into a single result. Ex, Blood Pressure is recorded in FHIR as a single Observation with 2 components, systolic and diastolic. DiagnosticReport is a list of references to many distinct but related Observations, ie a test panel:  https://en.wikipedia.org/wiki/Test_panel.  They have different semantics so if we want both to be available in the GMF there has to be a distinction somehow. I'm ok with using a single ObservationGroup type state but then it would have to specify something like "class: panel" or "class: multi-observation".

cjduffett:
Could we allow the `Observation` state to have one or more LOINC codes as multiple, related observations? That way we could infer that it is a multi-observation simply by checking the length of `codes`.

Then perhaps we have a separate `DiagnosticReport` state, since that creates both different FHIR and CCDA resources?

jawalonoski:
Once these conflicts are resolved I'll merge this pull request.

dehall:
@jawalonoski @cjduffett  PR has been rebased. I wanted to refactor the MultiObs/DiagnosticReport stuff based on our discussions, but it would be a nontrivial effort and take time away from higher priorities. I do need some pieces of this in order to convert the diabetes model to GMF

PR #134, opened 2017-01-23
Encounter and procedure duration

dehall:
Adds the ability for Encounters and Procedures to have realistic durations. Currently all encounters and procedures are recorded to take 15 minutes, which is obviously not true in reality. This PR adds:
#### 1.
 a "duration" field on Procedure which takes a range of time that the procedure may take, and
#### 2. 
 an "EncounterEnd" state which signifies the end of the current Encounter and accepts an optional discharge disposition. The time between Encounter and EncounterEnd states is recorded in the final output (CCDA and FHIR).  The switch to having a start + end allows us to use Delays in the middle to represent the amount of time taken. Note that Encounters and EncounterEnds do not have to exist in a 1-1 relationship; multiple Encounters could proceed into a single EncounterEnd, or a single Encounter could branch out and end in various different EncounterEnds.

Because an Encounter can now span a time range from minutes to months, Procedures, MedicationOrders, etc, do not necessarily have to be concurrent with the Encounter in which they occur.

Also updated the modules such that every module now has EncounterEnds, and added durations for various procedures as well.

### DIFF
--- a/lib/modules/encounters.rb
+++ b/lib/modules/encounters.rb
@@ -126,11 +126,18 @@ module Synthea
                else
                  :age_senior
                end
-        entity.record_synthea.encounter(type, time, reason)
+        entity.record_synthea.encounter(type, time, reason: reason)
+        # TODO: wellness encounters need their duration defined by the activities performed
+        # the trouble is those activities are split among many modules
+        entity.record_synthea.encounter_end(type, time + 1.hour)
       end
 
       def self.emergency_encounter(entity, time, reason = nil)
-        entity.record_synthea.encounter(:emergency, time, reason)
+        entity.record_synthea.encounter(:emergency, time, reason: reason)
+        # TODO: emergency encounters need their duration to be defined by the activities performed
+        # based on the emergencies given here (heart attack, stroke)
+        # assume people will be in the hospital for observation for a few days
+        entity.record_synthea.encounter_end(:emergency, time + 4.days)
       end
     end
   end
--- a/lib/records/fhir.rb
+++ b/lib/records/fhir.rb
@@ -11,9 +11,11 @@ module Synthea
         patient = basic_info(entity, fhir_record, end_time)
         synthea_record.encounters.each do |encounter|
           curr_encounter = encounter(encounter, fhir_record, patient)
+          encounter_end = encounter['end_time'] || synthea_record.patient_info[:deathdate] || end_time
+          # if an encounter doesn't have an end date, either the patient died during the encounter, or they are still in the encounter
           [:conditions, :observations, :procedures, :immunizations, :careplans, :medications].each do |attribute|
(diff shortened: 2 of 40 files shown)


lib/modules/encounters.rb
```diff
@@ -126,11 +126,18 @@ def self.encounter(entity, time, reason = nil)
...
+        entity.record_synthea.encounter_end(type, time + 1.hour)
       end
 
       def self.emergency_encounter(entity, time, reason = nil)
-        entity.record_synthea.encounter(:emergency, time, reason)
+        entity.record_synthea.encounter(:emergency, time, reason: reason)
+        # TODO: emergency encounters need their duration to be defined by the activities performed
+        # based on the emergencies given here (heart attack, stroke)
+        # assume people will be in the hospital for observation for a few days
+        entity.record_synthea.encounter_end(:emergency, time + 4.days)
```
jawalonoski: Probably need something that looks to see what was added in the encounter, sum all the durations (assumes that nothing happens concurrently), and add some padding....

lib/records/fhir.rb
```diff
@@ -182,19 +184,32 @@ def self.encounter(encounter, fhir_record, patient)
...
         if reason_data
           fhir_encounter.reason = FHIR::CodeableConcept.new('coding' => [{
                                                               'code' => reason_data[:codes]['SNOMED-CT'][0],
                                                               'display' => reason_data[:description],
                                                               'system' => 'http://snomed.info/sct'
                                                             }])
         end
+
+        if encounter['discharge']
+          fhir_encounter.hospitalization = FHIR::Hospitalization.new('dischargeDisposition' => {
```
jawalonoski: I think I'm confused... is `Hospitalization` a FHIR resource? Which version of FHIR?

jawalonoski: Oh, I see... that should be `FHIR::Encounter::Hospitalization.new`

jawalonoski:
In general, I like having a `duration` on Procedure. I expected to see the same thing on Encounter... but instead, it is just another state, possibly with Delays mixed in there... Not what I expected... but, OK. The trick is to figure out how much time should have passed based on what happened in the encounter...

dehall:
Originally I did have a `duration` on Encounter as well, but then I realized that the encounter duration is really just a sum of the durations of the events that happened in that encounter. This may violate the "principle of least astonishment" but I think this is the cleanest way to model it. If we specify the encounter duration directly there are a lot of ways weirdness could sneak in. For example we would either have to call `encounter_end` with a future date, or add new logic to the context that would look back at the encounter to end it once its time was up.

lib/modules/encounters.rb
dehall: I'm planning to just leave this part as-is, on the assumption that these are going to be replaced by generic modules eventually

PR #135, opened 2017-01-25
Diagnosed vs. Active Condition

cjduffett:
## Conceptual Changes

I created a new `Diagnosed Condition` condition and revised the meaning of `Active Condition`.

"Active" conditions are now conditions that the patient currently has, but _may not_ be in the patient's record yet if the patient hasn't had that condition's `target_encounter`. This new condition is useful for controlling program flow _before_ the diagnosis. For example, consider the following scenario:

A patient has a peanut allergy. They go to an allergist. The allergist runs a series of tests. If the patient has an _active_ peanut allergy, the test will return positive for a peanut allergy (even though the patient's record does not yet list a peanut allergy) and the allergist can then _diagnose_ that allergy.

This could also be useful for separating disease progression and treatment. An `Active Condition` could influence a disease to progress even though the condition is not yet diagnosed.

"Diagnosed" conditions must be both _active_ (the patient is currently affected by the condition) and recorded in the patient's record.

There is one caveat: If a patient has a condition that is no longer active, it will remain in the patient's record as a previously diagnosed condition, but the `Diagnosed Condition` condition will now return `false` since that condition is no longer active.

If the need arises to check if a patient _ever_ had a condition, perhaps we would need a new logical condition to test for that. However, in practice none of our modules need that sort of logic.

## Code Changes

I was able to remove the `present` attribute from the `Record` class. I expanded the `Person` entity to include the `onset_condition`, `end_condition`, `active_condition?`, and  `diagnosed_condition?` methods. These need to be evaluated for the _entity_ rather than just the record since the notion of a condition now spans both the entity (active conditions) and the record (diagnosed conditions).

I also refactored the record API with more consistent method names and some convenience methods that make manipulating the record a little more intuitive. All of the new `find_*` methods (e.g. `find_medication` or `find_condition`) now do a reverse scan of the patient's record and find the _most recent_ item.

### DIFF
--- a/lib/records/record.rb
+++ b/lib/records/record.rb
@@ -1,15 +1,12 @@
 module Synthea
   module Output
     class Record
-      attr_accessor :patient_info, :encounters, :observations, :conditions, :present, :procedures, :immunizations, :medications, :careplans
+      attr_accessor :patient_info, :encounters, :observations, :conditions, :procedures, :immunizations, :medications, :careplans
       def initialize
         @patient_info = { expired: false, uuid: SecureRandom.uuid }
         @encounters = []
         @observations = []
-        # store condition info
         @conditions = []
-        # check presence of condition
-        @present = {}
         @procedures = []
         @immunizations = []
         @medications = []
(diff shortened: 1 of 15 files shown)


cjduffett:
One thought I have now is that perhaps the `ConditionEnd` state should also specify a `target_encounter` state. That way a `ConditionEnd` state ends the patient's _active_ condition, but doesn't formally end the condition in the patient's record until they have another encounter and that condition is confirmed to be "ended". What do you think?

jawalonoski:
Ending the condition should already coincide with an Encounter. After all, nothing happens to the record outside of encounters. Specifying the `target_encounter` would only say it has to happen at this specific encounter, rather than the next encounter.

jawalonoski:
Also, what is this for? Precursor to allergy pull request? I want to stop infrastructure changes unless they are required to support module requirements.

dehall:
I vote no more `target_encounter`s.

cjduffett:
Precursor to some allergy stuff.

lib/records/record.rb
```diff
@@ -103,21 +127,27 @@ def medication_start(type, time, reasons, rx_info = {})
         @medications << med
       end
 
-      def medication_active?(type)
-        !@medications.find { |x| x['type'] == type && x['stop'].nil? }.nil?
+      def find_medication(type)
+        # Returns the most recent medication found matching the referenced type
+        @medications.reverse.find { |x| x['type'] == type }
```
jawalonoski: Performance

lib/records/record.rb
```diff
@@ -133,18 +163,24 @@ def careplan_start(type, activities, time, reason)
         }
       end
 
-      def careplan_active?(type)
-        !@careplans.find { |x| x['type'] == type && x['stop'].nil? }.nil?
+      def find_careplan(type)
+        # Returns the most recent careplan found matching the referenced type
+        @careplans.reverse.find { |c| c['type'] == type }
```
jawalonoski: Performance. Try `Array.reverse_each`

cjduffett:
As it turns out, I was able to model my Atopic Triad of allergies without needing this distinction. I am going to retire this idea for now. We can revisit it later.

PR #137, opened 2017-01-30
GMF diabetes

dehall:
Migrates the current Diabetes / Metabolic Syndrome ruby module to the Generic Module Framework. This is modeled as 2 separate modules - 1 for disease progression and 1 for care. The structure of the original ruby module rules made this approach simpler than the "sequence" model we normally see. There is still a lot of room for realism fixes in this; upon review of the ruby module there was a lot of detail in the vital sign #s but not a lot of realism in transition/prevalence/incidence %s, so some of those have been updated where possible.  Unfortunately a lot of the vital signs require calculations, so those could not be fully migrated to GMF yet. For now they remain a single rule within the Lifecycle module.

Module images (Sorry for the eye charts)

## Disease Progression
metabolic syndrome disease progression

## Standards of Care
metabolic syndrome standards of care


lib/generic/modules/metabolic_syndrome_disease.json
```diff
@@ -0,0 +1,651 @@
...
+        {
+          "transition" : "Potential_Diabetes"
+        }
+      ]
+    },
+
+    "Chance_to_Onset_Hypertension" : {
+      "type" : "Simple",
+      "distributed_transition" : [
+        { "distribution" : 0.594, "transition" : "Onset_Hypertension" },
```
jawalonoski: The probability that someone 18 or older has hypertension is `0.296`

The probability is `0.594` given they have diabetes. They do not have diabetes at this point.

lib/generic/modules/metabolic_syndrome_disease.json
```diff
@@ -0,0 +1,651 @@
...
+      "type" : "SetAttribute",
+      "attribute" : "hypertension",
+      "value" : true,
+      "direct_transition" : "Potential_Diabetes"
+    },
+
+    "No_Hypertension" : {
+      "type" : "SetAttribute",
+      "attribute" : "hypertension",
+      "value" : false,
```
jawalonoski: Setting hypertension to `false` is correct, but Graphviz shows `nil` -- did you change this after generating the images?

Setting the value to `nil` would recalculate whether or not they had hypertension every time step, which would be very wrong.

lib/generic/modules/metabolic_syndrome_disease.json
```diff
@@ -0,0 +1,651 @@
...
+    },
+
+    "Chance_to_Onset_Hypertension_at_Diabetes_Onset" : {
+      "type" : "Simple",
+      "complex_transition" : [
+        {
+          "condition" : {
+            "condition_type" : "Attribute",
+            "attribute" : "hypertension",
+            "operator" : "is not nil"
```
jawalonoski: I think you want to transition here if hypertension is `true` otherwise you want to recalculate whether or not they have hypertension (their probability increased)

lib/generic/modules/metabolic_syndrome_disease.json
```diff
@@ -0,0 +1,651 @@
...
+      "type" : "SetAttribute",
+      "attribute" : "hypertension",
+      "value" : true,
+      "direct_transition" : "Nephropathy_Progression"
+    },
+
+    "No_Hypertension_With_Diabetes" : {
+      "type" : "SetAttribute",
+      "attribute" : "hypertension",
+      "value" : false,
```
jawalonoski: Again, setting this to `false` is correct, but the image shows `nil`

lib/generic/modules/metabolic_syndrome_disease.json
```diff
@@ -0,0 +1,651 @@
...
+      "type" : "Symptom",
+      "symptom" : "Thirst",
+      "range" : { "low" : 50, "high" : 100 },
+      "direct_transition" : "Chance_of_Death_From_End_Stage_Kidney_Disease"
+    },
+
+    "Chance_of_Death_From_End_Stage_Kidney_Disease" : {
+      "type" : "Simple",
+      "distributed_transition" : [
+        { "distribution" : 0.0001, "transition" : "Death" },
```
jawalonoski: Not sure this will work correctly based on the configurable time-step... Perhaps we should just figure out how long people can live with ESRD (4-10 years if they are on dialysis, 1-4 weeks without dialysis) and set a future death date.

lib/generic/modules/metabolic_syndrome_care.json
```diff
@@ -0,0 +1,1857 @@
...
+      "codes" : [{
+        "system" : "SNOMED-CT",
+        "code" : "46177005",
+        "display" : "End stage renal disease (disorder)"
+      }],
+      "direct_transition" : "Check_Retinopathy"
+    },
+
+    "End_End_Stage_Renal_Disease" : {
+      "type" : "ConditionEnd",
```
jawalonoski: I'm pretty sure the only way this ends is that you die or get a kidney transplant.

jawalonoski:
In general this is great.

Can we adjust some of the unknown probabilities? Progression of disease still has a lot of 1% and 99% transitions in there. I know these are verbatim from the original Ruby module, but they are not really correct.

What is the percentage of diabetics with retinopathy, blindness, ESRD, neuropathy, etc?

I don't think this will end up with the right numbers for these sub conditions.

lib/generic/modules/metabolic_syndrome_disease.json
cjduffett: I encountered this when working on my allergies modules. Setting an attribute to `false` is interpreted as setting it to something false-ey, or `nil`. Consider using the strings `"true"` and `"false"` instead.

jawalonoski: Nah, we should use the booleans. If you have correctly identified the issue, then I think this means we should fix the Graphviz renderer.

lib/generic/modules/metabolic_syndrome_disease.json
cjduffett: Is there really a >50% chance _every week_ of acquiring hypertension?

lib/generic/modules/metabolic_syndrome_disease.json
dehall: I'll make the fix, it's trivial

lib/generic/modules/metabolic_syndrome_disease.json
dehall: That state should only be hit once, when the patient is 18 and the 'hypertension' attribute is nil. It sets the flag to true or false, and then the nil check checks explicitly for nil not just a false-y value so it's not hit again

cjduffett:
Agree with @jawalonoski, overall this looks great. I especially like the clear separation of disease progression from standards of care.

With this new model what is the overall prevalence of diabetes? It should still be about 10% and that's one aspect that will be immediately apparent on SyntheticMass if it's off.

dehall:
Updated the disease module with more realistic percentages (and fixed the graphviz display issue)

New module:

metabolic syndrome disease progression

jeffeastman:
According to my brother, Richard C. Eastman MD (Ret):

“Note that "pre-diabetes" is a state of dysglycemia that is not associated with classic symptoms of diabetes due to hyperglycemia (polyuria, polydipsia, polyphagia, blurred vision).  Not sure why the prediabetes track shows and scores these symptoms.  If a person has those symptoms and hyperglycemia A1c>6.5 they have diabetes.  A1c >5.7-6.5 is asymptomatic.  Those symptoms don't really occur until the renal threshold for glucose is reached, about 200 mg/ml FBS.  As I recall this equates with an A1c of about 9 (approx).  Above A1c of 6.7 there is end organ damage but usually no classic symptoms of diabetes.  

If the track were from undiagnosed diabetes coding for classic symptoms would make more sense, but only if blood glucose is elevated."

Dick developed a quantitative Monte Carlo simulation of Diabetes while at the NIH in the 1990s. It was an Excel spreadsheet implementation that was predictive of morbidity, mortality, treatment costs and quality of life years saved by comparing various treatment options. But it was quite slow. I reimplemented it in Smalltalk in 1998 to speed it up and recoded it into Java in 2000. This was the first model I implemented in PatientGen and we’ve extended the approach to include other conditions as well.

It’s good to see this model coming into the GMF set. The transition percentages need some work and need to be Race and Age specific if predictive accuracy is to be achieved. Looping Dick into this thread to see if he has other observations to contribute.

Jeff

jawalonoski:
I just noticed that none of the medications reference the reason.

dehall:
Changed the timescale of disease progression from weekly to monthly: 

metabolic syndrome disease progression

jawalonoski:
@jeffeastman Thanks for the detailed feedback. I suppose this starkly reveals the limits of our clinical knowledge. I'm not sure what to do. We could just remove those symptoms from pre diabetes. They only thing the symptoms do is push a person to scheduling a doctors appointment.

jeffeastman:
As Dick indicted in his note, the classical symptoms don't begin to present until the patient has a quite high A1c (~9). This is why the disease is so difficult to detect in its early stages without regular lab tests. Fortunately the standard metabolic lab panel will test for fasting blood sugar and, if elevated, the PCP can also test for A1c. 

Since its really the A1c that correlates with the morbidities, one thing you might do with the GMF model is use and describe that VitalSign explicitly. 

Removing the symptoms from the PreDiabetes state but elevating the fasting glucose & A1c vitals slowly during its lifetime could trigger detection in a wellness encounter down the road.

Dick also indicated that a track for UndiagnosedDiabetes (A1c>9) could present those symptoms and that might then lead to a more in-depth scheduled PCP visit. But at that A1c there has already been some likely organ damage.

Issue #151, opened 2017-02-17
Undefined local variable or method "dead"


cjduffett:
Encountered this error during a multithreaded sequential run seeded with `towns.json`.

Error trace:

```
undefined local variable or method `dead' for #<Synthea::World::Sequential:0x902a7b6>
/Users/cduffett/github/synthea/lib/world/sequential.rb:225:in `block in process_person'
org/jruby/RubyKernel.java:1290:in `loop'
/Users/cduffett/github/synthea/lib/world/sequential.rb:207:in `process_person'
/Users/cduffett/github/synthea/lib/world/sequential.rb:194:in `block in process_city'
/Users/cduffett/github/synthea/lib/world/sequential.rb:325:in `block in run_task'
/Users/cduffett/.rvm/gems/jruby-9.1.2.0/gems/concurrent-ruby-1.0.2-java/lib/concurrent/executor/java_executor_service.rb:94:in `run'
/Users/cduffett/.rvm/gems/jruby-9.1.2.0/gems/concurrent-ruby-1.0.2-java/lib/concurrent/executor/java_executor_service.rb:29:in `post'
/Users/cduffett/github/synthea/lib/world/sequential.rb:323:in `run_task'
/Users/cduffett/github/synthea/lib/world/sequential.rb:191:in `block in process_city'
org/jruby/RubyFixnum.java:297:in `times'
/Users/cduffett/github/synthea/lib/world/sequential.rb:190:in `process_city'
/Users/cduffett/github/synthea/lib/world/sequential.rb:79:in `block in run'
/Users/cduffett/github/synthea/lib/world/sequential.rb:325:in `block in run_task'
/Users/cduffett/.rvm/gems/jruby-9.1.2.0/gems/concurrent-ruby-1.0.2-java/lib/concurrent/executor/java_executor_service.rb:94:in `run'
```

PR #152, opened 2017-02-17
Fixed undefined 'dead' error

cjduffett:
Fixes https://github.com/synthetichealth/synthea/issues/151


jawalonoski:
Don't do that. You already do `person.had_event?(:death)` like 5 lines up.

Assign that to a variable and use it twice. Save yourself the lookup cost.

PR #153, opened 2017-02-20
Add CSV exporter to Synthea.

jawalonoski:
Exports Patients, Encounters, Observations, and so forth into separate CSV files. Each row in the files is linked to a patient and encounter when appropriate. The data for all patients is contained in one set of files. For example, all patients are contained in a single `output/csv/patients.csv` file -- one per row. The same pattern holds for the other resource types.

This pull request also fixes the allergies in the `text` exporter.

### DIFF
--- a/lib/tasks/tasks.rake
+++ b/lib/tasks/tasks.rake
@@ -21,9 +21,11 @@ namespace :synthea do
 
   desc 'generate'
   task :generate, [] do |_t, _args|
+    clear_output
     start = Time.now
     world = Synthea::World::Population.new
     world.run
+    Synthea::Output::CsvRecord.close_csv_files
     finish = Time.now
     minutes = ((finish - start) / 60)
     seconds = (minutes - minutes.floor) * 60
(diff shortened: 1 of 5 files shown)


lib/tasks/tasks.rake
```diff
@@ -123,6 +119,18 @@ namespace :synthea do
...
 
+  def clear_output
+    if Synthea::Config.sequential.clean_output_each_run
+      puts 'Clearing output folders...'
+      %w(html fhir CCDA text csv).each do |type|
+        out_dir = Synthea::Output::Exporter.get_output_folder(type)
+        FileUtils.rm_r out_dir if File.exist? out_dir
+        FileUtils.mkdir_p out_dir
+      end
+      Synthea::Output::CsvRecord.open_csv_files
```
dehall: this call should be outside the `if Synthea::Config.sequential.clean_output_each_run` and should instead be conditional based on `Synthea::Config.exporter.csv.export`

also the effect of opening files isn't implied by the name "clear_output". Not a huge deal but I'd rather the call be directly inside the generate and sequential tasks, so it's clear what's happening and it's symmetric with the call to `close_csv_files` (which should also be conditional based on `Synthea::Config.exporter.csv.export` if the open call is)

PR #164, opened 2017-02-27
Observation Categories

dehall:
Adds a mandatory `category` field to Observations, in anticipation of SHR FHIR profiles which require Observation.category.

Category is required for all Observations, and must come from the following value set: http://hl7.org/fhir/2017Jan/valueset-observation-category.html

For the wiki:

## Observation

The `Observation` state type indicates a point in the module where an observation is recorded. Observations include clinical findings, vital signs, lab tests, etc. The Observation state must come after an Encounter state in the module, but must have the same start time as that Encounter; otherwise it will not be recorded in the patient's record.  See the Encounter section above for more details.

### Observation Categories
Common observation categories include:

- `"vital-signs"` :  Clinical observations measure the body's basic functions such as such as blood pressure, heart rate, respiratory rate, height, weight, body mass index, head circumference, pulse oximetry, temperature, and body surface area.
- `"procedure"` : Observations generated by other procedures. This category includes observations resulting from interventional and non-interventional procedures excluding lab and imaging (e.g. cardiology catheterization, endoscopy, electrodiagnostics, etc.). Procedure results are typically generated by a clinician to provide more granular information about component observations made during a procedure, such as where a gastroenterologist reports the size of a polyp observed during a colonoscopy.
- `"laboratory"` : The results of observations generated by laboratories. Laboratory results are typically generated by laboratories providing analytic services in areas such as chemistry, hematology, serology, histology, cytology, anatomic pathology, microbiology, and/or virology. These observations are based on analysis of specimens obtained from the patient and submitted to the laboratory.
- `"exam"` : Observations generated by physical exam findings including direct observations made by a clinician and use of simple instruments and the result of simple maneuvers performed directly on the patient's body.
- `"social-history"` : The Social History Observations define the patient's occupational, personal (e.g. lifestyle), social, and environmental history and health risk factors, as well as administrative data such as marital status, race, ethnicity and religious affiliation.

### Supported Properties

| Attribute | Type | Description |
|:----------|:-----|:------------|
| `type` | `string` | Must be `"Observation"`. |
| `category` | `string` | The type of Observation being made, for example . Must be a valid [HL7 Observation Category](http://hl7.org/fhir/2017Jan/valueset-observation-category.html). |
| `unit` | `string` | The unit of measure in which the observation is recorded (e.g. `"cm"`). |
| `codes` | `[]` | One or more codes that describe the Observation. Must be valid [LOINC codes](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-Basics#loinc-codes). |

### DIFF
--- a/lib/generic/modules/allergies/allergy_panel.json
+++ b/lib/generic/modules/allergies/allergy_panel.json
@@ -52,6 +52,7 @@
 
     "Peanut_Test_Positive": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -69,6 +70,7 @@
 
     "Peanut_Test_Negative": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -113,6 +115,7 @@
 
     "Nut_Test_Positive": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -130,6 +133,7 @@
 
     "Nut_Test_Negative": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -174,6 +178,7 @@
 
     "Fish_Test_Positive": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -191,6 +196,7 @@
 
     "Fish_Test_Negative": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -237,6 +243,7 @@
 
     "Shellfish_Test_Positive": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -254,6 +261,7 @@
 
     "Shellfish_Test_Negative": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
@@ -298,6 +306,7 @@
 
     "Wheat_Test_Positive": {
       "type": "Observation",
+      "category" : "laboratory",
       "codes": [
         {
           "system": "LOINC",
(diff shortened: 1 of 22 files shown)


cmoesel:
In SHR, Observation does *not* require `category`.  In some elements based on Observation (like VitalSign), it specifies that one of the categories must be a code meaning "vital sign" -- so in that case lower card is `1` because one must be the code for vital sign.  But in general, Observation shouldn't require category.

It appears there may be a bug in the profile exporter that misses the "includes code" constraint -- so the use case I mentioned above (Vital Sign) will show a lower card of `1` in the profile, but won't indicate that one value should be the code for vital sign.  This will require a slice to specify it correctly, and right now the exporter has very limited support for slices.

cmoesel:
BTW -- I just logged a bug on the exporter for this: https://github.com/standardhealth/shr-fhir-export/issues/4

jawalonoski:
Interesting. But right now category is `1..1`... and the PR satisfies that constraint. So...

jawalonoski:
Dylan, it looks like you removed a lot of target encounters... can you comment?

dehall:
Removing target_encounters was intentional - those are optional/deprecated everywhere except ConditionOnset at this point, so I removed them from the Observations that I touched here. Sorry I probably should have mentioned that in the description.

Re: category not being truly required on Observation - even if it's not strictly required on every Observation it's at least optional so having the category there won't break anything. That said though this implementation does mean that a new module could use a code and specify the wrong category for it.  An alternative approach would be to parse through the SHR StructureDefinitions and produce a mapping of code => category, and then when exporting FHIR lookup the category. Would that be a better option?

jawalonoski:
No. This approach allows us to be ahead of the profiles. And unit tests (TBD) that focus on those profiles (matching by code, possibly) can catch those errors as they are introduced.

cmoesel:
@jawalonoski : Just to be clear, `Observation.category` is not `1..1` now.  See the [Observation struct def](https://github.com/synthetichealth/synthea/blob/shr_fhir_profiles/resources/shr_profiles/StructureDefinition-shr-observation-Observation.json#L538-L545).  That said, it _is_ `1..*` on _some_ elements (notably, the vital signs).

@dehall: I'm not sure I completely understand what you're suggesting, but it sounds complicated to me.  I'm not suggesting that GFM Observation shouldn't support `category` at all -- I'm just suggesting that it shouldn't be _required_ for _all_ Observations in GFM.

PR #176, opened 2017-03-22
Added 'Download Data and Concepts' section

cjduffett:
This is all I really wanted to add. I mainly wanted to make sure the download and concept links were together in at least one place. I would refer AMIA Datathon participants to this gh pages site for more info on Synthea/the synthetic data.

Anything else you think we should add?


dehall:
If we're linking to the full 1 million patients we should probably indicate the file size and provide more clarity about what that file actually includes. People might click out of curiosity but not actually want to download a multi-GB file, or they may not want FHIR, etc.

cjduffett:
How do you think we should present these links then? Maybe more download buttons on the sidebar for each file? Maybe we have links to download 1k, 10k, and 100k patients as well?

jawalonoski:
You don't actually have any files to download... don't you think those should be linked before we merge this?

cjduffett:
The following are currently live and valid links:

**concepts.csv**:
https://syntheticmass.mitre.org/downloads/2017_03_20/concepts.csv

**1k patients (25.7kB)**:
https://syntheticmass.mitre.org/downloads/2017_04_06/synthea_1k.tar.gz

**1M patients (29GB)**:
https://syntheticmass.mitre.org/downloads/2017_02_27/synthea_1M.tar.gz

I'm not sure what happened to the 10k and 100k files but I'm re-generating them now. I'll post an update when those files are linked.

jawalonoski:
The files only need to be as up-to-date as the data on the FHIR server. Which we don't update continuously. They do not to be as up-to-date as `master`.

jawalonoski:
If you look at the links that Dylan added to the branch, you'll see that they link to `href="#"` which isn't going to be helpful. Please update those links in the code.

cjduffett:
Closing this now that the new version of the site is up.

PR #177, opened 2017-03-22
Add Osteoporosis module

dehall:
Based on feedback that fractures should trigger an osteoporosis workup in the elderly:

 - Adds a simple module that onsets Osteoporosis at realistic prevalences by age
 - Adds a section in the Injuries module following broken bones that can result in a diagnosis of osteoporosis.

Injuries snippet:
injuries_snippet

Osteoporosis module:
osteoporosis

### DIFF
--- a/lib/generic/modules/injuries.json
+++ b/lib/generic/modules/injuries.json
@@ -1225,8 +1225,119 @@
           ]
         }
       ],
-      "direct_transition": "End_Broken_Bone_Encounter"
+      "direct_transition": "Consider_Osteoporosis"
     },
+
+    "Consider_Osteoporosis" : {
+      "type" : "Simple",
+      "complex_transition" : [
+        {
+          "condition" : {
+            "condition_type" : "Active Condition",
+            "codes" : [{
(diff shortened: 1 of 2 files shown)


cjduffett:
Could you also update the injury incidence to make broken bones more frequent for those with osteoporosis? It helps that you already scale the likelihood of osteoporosis with age, so injuries can just follow it.

dehall:
Ok. It will make the start of the injuries module even more complex but I'll do that. We should think about splitting up the injuries from that module into separate module so that we can cleanly define more targeted incidence #s. (ex, osteoporosis makes fractures more likely, but shouldn't affect the likelihood of any of the other injuries we model. At a glance I don't see anything else that would be significantly affected by condition but there could be)

cjduffett:
The incidence updates look good, thank you. The original reason I didn't put the injuries in separate modules is that I didn't want acute injuries to occur simultaneously, e.g. patient's shouldn't get a broken leg while suffering from a spinal injury, etc. Maybe that's an incorrect assumption.

PR #181, opened 2017-03-28
Growth from birth-20 driven by cdc growth chart data

dehall:
Height and weight from birth to age 20 are now driven by CDC growth chart information. At birth patients are assigned a percentile for both height and weight which they track until age 20.  In reality people will move up or down in the percentiles as they age but modeling that and keeping the #s realistic would be much more difficult.

Also adds a rake task that condenses the charts from multiple CSVs into a single JSON


dehall:
Turns out this doesn't work on JRuby with the statistics2 gem.

PR #182, opened 2017-04-04
Growth charts

dehall:
Introduces data from CDC growth charts to drive people's height and weight from birth to age 20. 
Data source: https://www.cdc.gov/growthcharts/percentile_data_files.htm

This impl uses the LMS parameters described at the data source page to allow for height and weight from the 3rd-97th percentile.

Note: using that LMS formula required bringing back the `distribution` gem to convert percentile to z-score. After taking a second look into why we initially removed it, it really wasn't causing any problems (besides a warning at startup)


jawalonoski:
I think the gem didn't work with JRuby -- did you verify that?

dehall:
I was never convinced it didn't work with JRuby before, but I didn't feel it was worth it to push back on removing it. I guess there's a lesson in that.

There **is** a warning at startup: 
```
/home/dehall/.rvm/gems/jruby-9.1.6.0/gems/distribution-0.7.3/lib/distribution/math_extension/incomplete_beta.rb:178: warning: `*' after local variable or literal is interpreted as binary operator
/home/dehall/.rvm/gems/jruby-9.1.6.0/gems/distribution-0.7.3/lib/distribution/math_extension/incomplete_beta.rb:178: warning: even though it seems like argument prefix
```
But if you actually look at that line, the only * operators there are multiplication anyway. So interpreting that as a binary operator is correct.  Further, we don't even use that class so if it was broken we wouldn't see it. We only use Normal.rng and now Normal.p_value, which are mostly self-contained and don't call other classes in the Distribution library.

For a quick check, I ran 100 patients on JRuby and on MRI 2.0.0. Results for adult (age > 20) height in inches follow. Height would be a prime spot to see weirdness because the height value are based on Normal.rng (to assign a percentile at birth) and Normal.p_value (to calculate height based on age and percentile)

Run1:
```
   "adult_height_inches": {
    "M": {
      "64": 6,
      "66": 1,
      "67": 7,
      "68": 8,
      "69": 9,
      "70": 5,
      "71": 3,
      "72": 3,
      "74": 6,
    },
    "F": {
      "59": 2,
      "61": 3,
      "62": 3,
      "63": 7,
      "64": 11,
      "65": 7,
      "66": 6,
      "67": 3,
      "69": 4,
    }
  },
```

Run2:
```
 "adult_height_inches": {
    "M": {
      "64": 2,
      "65": 2,
      "66": 4,
      "67": 4,
      "68": 8,
      "69": 14,
      "70": 10,
      "71": 4,
      "72": 4,
      "73": 3,
      "74": 4
    },
    "F": {
      "59": 6,
      "60": 2,
      "61": 2,
      "62": 9,
      "63": 6,
      "64": 13,
      "65": 6,
      "66": 5,
      "67": 2,
      "68": 4,
      "69": 3
    }
  },
```

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: ruby
 rvm:
   - 2.0.0
+  - jruby-9.1.5.0
```
cjduffett: I take it you fixed the jruby warnings by using a different jruby version? Cool.

cjduffett: Are there still the warnings at startup?

dehall: Unfortunately the warning at startup is still there. But it refers to a file that we're not using so it's really only a tiny annoyance not a problem with the code.

jawalonoski:
So, growth is an annual event right now. But with the CDC growth charts having monthly data, we could increase the resolution here. This would especially increase the realism with infants, where pediatric encounters are very frequent and little kids grow continuously and not as annual step functions. Thoughts?

jeffeastman:
+1 I’m using the same data and aging all Vitals monthly, including height & weight, using Expressions in a Vital_Signs module. Of course, Ive had to add a Matrix primitive to Expression to store the CDC data etc. 

Jeff

dehall:
I don't see any issue with switching it to monthly.  It would be a slight performance hit but I wouldn't expect anything significant.  I assume we'd only want monthly resolution for age < 20 but we could scale our adult weight gain #s to be monthly as well. I can add that in this PR if you'd like.

cjduffett:
+1 for monthly

PR #183, opened 2017-04-07
Submodules directory for Graphviz

cjduffett:
All submodules rendered by graphviz are now placed in a separate 'submodules' directory in the same export path. For example:

```
output/
    |_ graphviz/
        |_ submodules/
        |   |_ <submodule_files>
        |_ <main_module_files>
```

One other design change I'd suggest, but haven't implemented yet: I think we should only add the main module files to the index file for the GitHub pages site carousel. That way users aren't viewing submodules, which are often meaningless without context. Thoughts?

### DIFF
--- a/lib/tasks/graphviz.rb
+++ b/lib/tasks/graphviz.rb
@@ -10,9 +10,7 @@ module Synthea
 
       def self.generate_graphs
         folder = Synthea::Config.graphviz.output
-        submodule_folder = File.join(folder, 'submodules')
         FileUtils.mkdir_p folder unless File.exists? folder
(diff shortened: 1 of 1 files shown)


cjduffett:
And regarding the index file for the GitHub pages site?

jawalonoski:
Removing the submodules from the carousel? Fine. When we get a better way to browse the modules having the submodules visible will be important.

cjduffett:
The directory structure now matches the structure of the modules directory, e.g.:
```
output/graphviz/Allergies and Treatment.png
output/graphviz/allergies/Allergy Incidence.png
```

Issue #114, opened 2016-12-14
Observations Feedback


cjduffett:
- [ ] Hemoglobin A1c should be reported to only 1 decimal place. Also our max possible value for A1c is too high - "incompatible with life", in fact.

- [ ]

PR #184, opened 2017-04-07
Typo

jawalonoski:
Fixes a misspelling in the dementia module and changes EGRF to an integer because the FHIR exporter only handles `Observation.valueQuantity` and `'>60'` is a String which would produce invalid resources.


dehall:
For context, the ">60" value was suggested based on clinician feedback - #114 . Ideally we'd be able to keep that but not if it breaks the export

PR #185, opened 2017-04-12
Detailed & configurable prevalence data for all conditions+medications

dehall:
Adds tracking of prevalence data for all combinations of age/race/gender/precondition. 

This implementation uses a multi-level nested hash to track the various combinations. While ideally we would use a database to keep patients in, and then run queries to get whatever stats we want, that would involve a major refactoring of the application.

This approach uses a template in the /resources/ directory: 
[prevalence_template.csv.txt](https://github.com/synthetichealth/synthea/files/916489/prevalence_template.csv.txt)

Which is then populated with statistics following a run:
[targeted_prevalences.csv.txt](https://github.com/synthetichealth/synthea/files/916492/targeted_prevalences.csv.txt)

Also exported is a csv with prevalence for all conditions and medications for living patients.
[all_prevalences.csv.txt](https://github.com/synthetichealth/synthea/files/916496/all_prevalences.csv.txt)

There is a small performance hit with this level of data collection, so it can be turned off if not actively tracking the generated stats.

### DIFF
--- a/config/synthea.yml
+++ b/config/synthea.yml
@@ -32,6 +32,7 @@ synthea:
     # population is scaled to accurately represent MA (using a "true" population of 6,794,422)
     population: 100
     real_world_population: 6794422 #sum of the jsons in the config folder
+    track_detailed_prevalence: true
     multithreading: false
     thread_pool_size:
     # numbers picked by experimentation on 2013 i5 mbp, may need to be tweaked on less powerful systems
--- a/lib/world/sequential.rb
+++ b/lib/world/sequential.rb
@@ -118,7 +118,7 @@ module Synthea
         end
 
         puts 'Generated Demographics:'
-        puts JSON.pretty_unparse(@stats.except(:occurrences)) # occurrences is way too unwieldy to print out
+        puts JSON.pretty_unparse(@stats.except(:occurrences, :all_occurrences, :living_occurrences, :dead_occurrences)) # occurrences is way too unwieldy to print out
 
         if Synthea::Config.generic.log_state_statistics
           puts 'State Statistics:'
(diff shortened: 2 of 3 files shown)


cjduffett:
@dehall: This may seem a little odd, but what about Redis? Since all you're really doing is atomically incrementing integers at some key:value pair, Redis would do that really fast and scale really easily. 

Also, as I augment Synthea to run in parallel on AWS I was also thinking that would be a good way to track stats across multiple Synthea instances.

Yes, Redis is (sort of) a database, and yes this would make Redis an (optional?) dependency of Synthea, but I think it would meet your desire for simplicity and speed.

jawalonoski:
"No" on adding external dependencies at this point.

dehall:
@jawalonoski I've been thinking about this a bit, and with the restriction on new dependencies the only alternative I can think of is to calculate category totals at the end instead of at each iteration.
Right now every possible combination of sums is stored and incremented individually, which is kind of a waste since we don't use most of those sums.

Ex.
```
        statuses = [living, '*']
        ages = [age, '*']
        genders = [gender, '*']
        races = [race, '*']
        conditions = top_level_conditions + ['*']

        # this looks like a massive loop, but each list will only have a couple items in it
        statuses.each do |s|
          ages.each do |a|
            genders.each do |g|
              races.each do |r|
                conditions.each do |c|
                  @stats[:occurrences][category][s][a][g][r][c][code].increment
                end
              end
            end
          end
        end
      end
```

The new approach would change the collection to just:
```
conditions = top_level_conditions + ['*']
conditions.each do |c|
  @stats[:occurrences][category][status][age][gender][race][c][code].increment
end
```

and then the other loops would move into the retrieve_prevalence function to calculate arbitrary sums.  Below is an example, though this abstracts away some details and there are short-circuit and caching possibilities.

```
def get_prevalence
  sum = 0
  statuses = # all status keys 
  genders = # all gender keys
  # etc
   statuses.each do |s|
     ages.each do |a|
       genders.each do |g|
         races.each do |r|
           conditions.each do |c|
             sum += @stats[:occurrences][category][s][a][g][r][c][code]
           end
         end
       end
     end
   end
```

Issue #196, opened 2017-05-16
Last names can still have Hash when append_hash_to_person_names is false


gteichrow:
In lifecycle.rb, there is a section of the code dealing with married women where the last_name was reset but the reset logic did not check the append_hash_to_person_names flag to see if the hash should be appended.  

I have a fix in my fork and will commit, once I figure out git, lol.  I'm an svn fart and I think I have it, but want to get thru the pull request, etc.

jawalonoski:
Great. Looking forward to the pull request!

gteichrow:
Shazam, baby's first pull-request.  w00t.

jawalonoski:
Thanks for the contribution!

PR #197, opened 2017-05-16
Fixed an issue on LastName generation where hash added innappropriately

gteichrow:
Fixed an issue on LastName generation where in one case the name was still getting the hash added if if the config flag was turned off.


gteichrow:
Corresponds to issue #196

jawalonoski:
Thanks for the pull request!

Issue #200, opened 2017-05-30
Master head `rake ..` tasks infinitely recurse on `require 'fhir_client'`. (synthea.rb:12)


preston:
When running `rake <anything>` off master as noted in the instructions, a gnarly recursive stack overflow is emitted. This happens on `rake`, `rake -T`, `bundle exec rake synthea:generate` etc when synthea.rb `requires 'fhir_client'` on line 12.

(additional recursive calls omitted)

> /Users/preston/.rvm/gems/ruby-2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:131:in `block (2 levels) in <class:Numeric>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:131:in `block (2 levels) in <class:Numeric>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:131:in `block (2 levels) in <class:Numeric>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/addressable-2.5.1/lib/addressable/version.rb:27:in `join'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/addressable-2.5.1/lib/addressable/version.rb:27:in `<module:VERSION>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/addressable-2.5.1/lib/addressable/version.rb:22:in `<module:Addressable>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/addressable-2.5.1/lib/addressable/version.rb:21:in `<top (required)>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/addressable-2.5.1/lib/addressable/uri.rb:19:in `require'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/addressable-2.5.1/lib/addressable/uri.rb:19:in `<top (required)>'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/fhir_client-3.0.1/lib/fhir_client/client.rb:3:in `require'
/Users/preston/.rvm/gems/ruby-2.4.0/gems/fhir_client-3.0.1/lib/fhir_client/client.rb:3:in `<top (required)>'
[...]

Issue #199, opened 2017-05-30
Update ruby.


preston:
Ruby 2.1.0 is very old, at least in the Ruby community. Any good reason why this gem should be pegged to such an old release? If not.. could I get a merge in if I submit the PR? 😄

jawalonoski:
No strong reason, other than we were interested in maintaining backwards compatibility. Which is a good practice in general. However, you could add a newer version to the Travis file. In general, yes, we take pull requests so long as they make sense.

jawalonoski:
See #200

jawalonoski:
Fixed by #205

PR #202, opened 2017-06-06
Ruby v2.4 support for #199 and #200.

preston:


preston:
CC @jawalonoski

jawalonoski:
So, it doesn't look like your fork of health data standards is building correctly, based on this PR: https://github.com/projectcypress/health-data-standards/pull/501

Can you verify?

preston:
@jawalonoski The travis logs show everything is actually passing. It appears to be rubocop style errors that are causing the build to fail.

preston:
Will recreate this to update the example mongoid config file using the correct v5 structure.

PR #204, opened 2017-06-08
Add support to Synthea so it can now either generated 100 alive patie…

AmartC:
…nts or 100 dead patients. In addition, ensure that causes of death are always based on modifying some of the modules

### DIFF
--- a/config/synthea.yml
+++ b/config/synthea.yml
@@ -31,6 +31,7 @@ synthea:
     generate_fingerprints: false
   sequential: # sequential population generation... one at a time...
     # population is scaled to accurately represent MA (using a "true" population of 6,794,422)
+    only_dead_patients: false
     population: 100
     real_world_population: 6794422 #sum of the jsons in the config folder
     track_detailed_prevalence: true
--- a/lib/world/sequential.rb
+++ b/lib/world/sequential.rb
@@ -45,6 +45,7 @@ module Synthea
         }
 
         @population_count = Synthea::Config.sequential.population
+        @only_dead_patients = Synthea::Config.sequential.only_dead_patients
         @generate_count = Concurrent::AtomicFixnum.new(0)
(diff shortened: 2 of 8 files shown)


dehall:
The `run_random_dead` function is very similar to the `run_random` function so it might make more sense to combine those and have the check on `Synthea::Config.sequential.all_dead` be internal to that combined function, so that if we want to make changes in the future we only have to do it in one spot.

Also, the config setting only applies to running a random simulation, not the simulation based on targeted city populations (ex `bundle exec rake synthea:generate['./config/Suffolk_County.json'] ` to generate patients in that one county, with realistic but scaled populations and demographics ) so I think the setting name should indicate that somehow.

AmartC:
Yeah, I had them as one function but there were a lot of conditionals I added and it seemed a bit messier so I decided to make it as two functions. I can go back and change the variable names since the focus was to have Synthea generate the patients one by one rather than through city information.

config/synthea.yml
```diff
@@ -31,6 +31,7 @@ synthea:
     generate_fingerprints: false
   sequential: # sequential population generation... one at a time...
     # population is scaled to accurately represent MA (using a "true" population of 6,794,422)
+    random_all_dead: false
```
jawalonoski: Maybe `only_dead_patients` is a better description.

lib/world/sequential.rb
```diff
@@ -45,6 +45,7 @@ def initialize(datafile = nil)
         }
 
         @population_count = Synthea::Config.sequential.population
+        @random_all_dead_bool = Synthea::Config.sequential.random_all_dead
```
jawalonoski: Again, something like `@only_dead_patients` is more descriptive/readable.

jawalonoski:
Some variable name changes and we're ready to go here.

Issue #130, opened 2017-01-15
DSTU2 Exporter


jawalonoski:
Most production implementations to be used within the United States that will be certified for use in MU 2015 are going to use DSTU2.

jawalonoski:
Fixed by #205

PR #203, opened 2017-06-06
Updated Ruby 2.4 support patch.

preston:
Related to #199 and #200. Obsoletes #202.

PR #205, opened 2017-06-13
DSTU2 FHIR exporter

dehall:
Adds a DSTU2 FHIR exporter. The DSTU2 exporter is just cloned from the regular [STU3] FHIR exporter with a few field names changed to the DSTU2 versions. By default the DSTU2 exporter is turned off.

Note that the HDS dependency had to be changed because of version issues with nokogiri. (HDS master requires 1.6, dstu2_models requires 1.7)


jawalonoski:
Fixes #130

jawalonoski:
Can you change this to the `ruby2_4` branch of HDS? I'm trying to get this and PR #203 to do both.

jawalonoski:
Fixes #199 and #200

PR #210, opened 2017-06-29
Providers and Hospitals

mdk6jd:
## Overview
Adds providers to Synthea.  Currently supported services and their respective facility types include:
 - ambulatory (outpatient) service - acute hospital, non acute hospital, hospital inpatient satellite, hospital satellite
 - inpatient service - acute hospital, non acute hospital, hospital inpatient satellite 
 - emergency service - acute hospital, satellite emergency facility 

Hospitals provide different services based on the type of facility.  The services above were already built into Synthea (called encounter_class in most cases).

Healthcare facility information is contained in the following locations:
- healthcare_facilities.json - All acute hospitals, non-acute hospitals, hospital inpatient satellites, satellite emergency facilities and hospital satellites registered in the state of Massachusettes.  Synthea will grab hospital information from this file by default.
- test_healthcare_facilities - four fabricated hospitals for testing purposes
- test_single_healthcare_facilities - one fabricated hospital for testing purposes 

## Implementation
Patients are assigned to their closest service providers for each type of service upon creation, which is stored in `entity`'s attributes as `preferred_provider`.  Distance is determined using the Haversine formula in GeoRuby and accounts for the curvature of the Earth.  `current_provider` is a list of the providers for active encounters so that procedures and medication prescriptions occurring during encounters would know which provider that encounter occurred at.

To keep a reference to all the providers in Synthea and support different services, the provider class includes the following global attributes:
- `provider_list` - Array of all providers that are imported
- `services` - Hash of services to providers that provide them 

Similarly, the hospital class includes:
- `hospital_list `- Array of all hospitals that are imported 

Provider keeps a count of the utilization of each provider, including encounters, procedures, labs, and prescriptions.  Encounters and procedures occur at the closest hospitals that provide that service.  Prescriptions associated with an encounter are counted as a prescription administered by the encounter's provider.  Otherwise, standalone prescriptions are counted as given by the patient's default hospital.  Labs are also assumed to be administered by the patient's default hospital.  The counts of utilization match those reported by the csv output.  (The only exception is that death certifications are not counted as an encounter for the provider but is included in the csv output.)

In regards to output, hospital utilization counts are currently output as text under the file named hospital_information.txt.  Each patient's text file states their default hospital in the basic info section. In addition, hospital information is output in fhir and fhir dstu2 formats.  In each patient's fhir output, the patient's default hospital is assigned as the `generalPractitioner`.  For each encounter, the respective hospital for that encounter is assigned as the `serviceProvider`. 

## Further Considerations
Finding the closest service provider currently loops through all the providers with the service and computes and compares distance.  This occurs three times for each patient for the three service types currently supported by this model.  This can be optimized using a voronoi diagram or quad 
tree.


config/test_healthcare_facilities.json
```diff
@@ -0,0 +1,58 @@
+{
```
dehall: Since this file is only used by the test suite it would be more appropriate to put it in the test/fixtures folder.

lib/generic/states.rb
```diff
@@ -285,10 +285,19 @@ def perform_encounter(time, entity, record_encounter = true)
           end
 
           if record_encounter
+            # find closest service provider
+            service = @encounter_class
+            provider = Synthea::Provider.find_closest_service(entity, service)
+            # hash below is added so that procedures during encounters will have a reference to provider
+            entity[:current_provider] = provider
```
dehall: I know I suggested this approach but upon further consideration I think it should be slightly modified. Nothing in Synthea prevents a person from being in 2 different encounters at once(in 2 different modules), so if that happens then both modules are setting to this same variable and could stomp on each other. The easiest way to prevent that would be making `entity[:current_provider]` a hash, which takes the name of the current module as the key, ex. `entity[:current_provider][@context.name] = provider` . It feels like more of a hack to do it that way, but I think the ideal way would require a lot of changes.

test/fixtures/generic/encounter.json
```diff
@@ -8,6 +8,11 @@
         "Annual_Physical": {
             "type": "Encounter",
             "wellness": true,
+            "codes": [{
```
dehall: Was this added because something broke otherwise? We don't require codes for wellness encounters

lib/records/fhir.rb
```diff
@@ -56,6 +59,8 @@ def self.basic_info(entity, fhir_record, end_time = Time.now)
                                              'text' => { 'status' => 'generated',
                                                          'div' => '<div xmlns="http://www.w3.org/1999/xhtml">Generated by <a href="https://github.com/synthetichealth/synthea">Synthea</a>. '\
                                                                    "Version identifier: #{Synthea::Config.version_identifier}</div>" },
+                                             # hospital
+                                             'managingOrganization' => { 'reference' => prov.fullUrl.to_s },
```
dehall: Not sure `managingOrganization` is the best field for this. The description is "Organization that is the custodian of the patient record" but you're really trying to represent the fact that this is where they usually go for health care services. I think the `generalPractitioner` field is more appropriate, even though we're putting a hospital in there not a single doctor.

lib/records/fhir.rb
```diff
@@ -37,6 +37,9 @@ def self.basic_info(entity, fhir_record, end_time = Time.now)
         end
```
dehall: There should be similar changes in the FHIR DSTU2 exporter as well. Take a look at the CCDA exporter as well but we've let that fall significantly behind already so updating that isn't critical.

lib/world/provider.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      person_location = entity.attributes[:coordinates_address].to_coordinates
+      person_point = GeoRuby::SimpleFeatures::Point.from_x_y(person_location[0], person_location[1])
+      closest_distance = 100_000_000
+      closest_service = entity.hospital
+
+      # if service is nil or not supproted by simulation, patient goes to default hospital
+      if Provider.services.key?(service.to_sym)
+        Provider.services[service.to_sym].each do |h|
+          service_location = h.attributes[:coordinates]
```
dehall: Rather than looping through all providers at every call, an alternative approach could be a voronoi diagram (or multiple diagrams, 1 per service/type) generated once when the app starts. Then when you look up "closest provider" it's more like O(1) than O(N). Can you run a quick test to see how much this impacts performance? Run 100 patients a few times with the latest from master and a few times with this branch and record the time it takes?

Just a note - I do see this library to create voronoi diagrams but it uses C extensions so I don't think we can use it. https://github.com/abscondment/rubyvor

config/test_healthcare_facilities.json
mdk6jd: Good point, I've moved both test files into the folder.

test/fixtures/generic/encounter.json
mdk6jd: Yes, in providers_test.rb test_utilization_encounter, I call perform_encounter in states.rb.  All the other tests I looked at call perform_encounter with record_encounter = false, but I want to test the increment_encounters, which requires record_encounter to be true.  Somewhere during the record process, it throws an error because codes is null.  Is there a better way to test this without adding codes to the wellness encounter?

lib/records/fhir.rb
mdk6jd: That seems more appropriate, will change.

lib/records/fhir.rb
mdk6jd: The FHIR DSTU2 exporter has been modified

lib/generic/states.rb
```diff
@@ -284,11 +284,21 @@ def perform_encounter(time, entity, record_encounter = true)
             cond = entity[@reason].to_sym if cond.nil? && entity[@reason]
           end
 
+          # find closest service provider
+          service = @encounter_class
+          provider = Synthea::Provider.find_closest_service(entity, service)
```
jawalonoski: Seems like you don't need this, because you already assigned providers to each Entity and Person...

jawalonoski:
It seems like each `Entity` has a set of `current_provider`'s and each `Person` (which is a subclass of `Entity`) also has a duplicative set of `hospital`'s....

Doesn't seem like we need both.

lib/records/hospital_exporter.rb
```diff
@@ -0,0 +1,131 @@
...
+      end
+
+      def self.convert_to_fhir(hospital_list)
+        fhir_record = FHIR::Bundle.new
+        fhir_record.type = 'collection'
+        resource_id = SecureRandom.uuid
+
+        hospital_list.each do |h|
+          hospital_resource = FHIR::Organization.new(
+            'id' => resource_id,
```
jawalonoski: Each Organization in the list shouldn't have the same `resource_id` -- you need to generate a new UUID on each loop.

lib/records/hospital_exporter.rb
```diff
@@ -0,0 +1,131 @@
...
+              'coding' => [{
+                'code' => 'prov',
+                'display' => 'Healthcare Provider',
+                'system' => 'http://hl7.org/fhir/ValueSet/organization-type'
+              }],
+              'text' => 'Healthcare Provider'
+            }
+          )
+
+          entry = FHIR::Bundle::Entry.new
```
jawalonoski: Add `FHIR::Extensions` on the Bundle for the number of encounters, procedures, etc, like you did for the text exporter.

lib/records/hospital_exporter.rb
```diff
@@ -0,0 +1,131 @@
...
+        dirs = [base, folder_name]
+
+        folder = File.join(*dirs)
+
+        FileUtils.mkdir_p folder unless File.exist? folder
+
+        folder
+      end
+
+      def self.fhir_upload(bundle, fhir_server_url, fhir_client = nil)
```
jawalonoski: Did you try executing this `fhir_upload` on any FHIR servers such as HAPI?

lib/world/provider.rb
```diff
@@ -0,0 +1,92 @@
...
+      if service == 'outpatient' || service == :outpatient
+        service = 'ambulatory'
+      end
+
+      # if service is nil or not supproted by simulation, patient goes to default hospital
+      return entity.hospital[:ambulatory] if service.nil?
+
+      case service.to_sym
+      when :ambulatory
+        entity.assign_ambulatory_provider if entity.hospital[:ambulatory].nil?
```
jawalonoski: `entity.hospital` will only work on `Patient` objects, right?

It should fail on `Entity` objects that aren't also `Patients`.

jawalonoski: Same thing for `assign_ambulatory_provider` method... that is on `Person` and not `Entity`... so, you're argument is named correctly for you're lucking out that the only type of entities at the moment are People.

Hospitals could also be entities, although you didn't implement them this way.

lib/world/sequential.rb
```diff
@@ -72,6 +72,13 @@ def initialize(datafile = nil)
 
         @city_populations = JSON.parse(datafile) if datafile
 
+        # import hospitals
+        file = File.read './config/healthcare_facilities.json'
+        providers = JSON.parse(file)
+        providers.each do |_provider_name, provider_stats|
```
jawalonoski: What is the point of this `each` loop?

You aren't actually doing anything with the `Synthea:Hospital` after you create it.

jawalonoski: Oh, I see... you are caching it in the constructor.

Why not just have the class load the definitions? Then you can do it in one place and have it self-contained like the modules.

It seems like you this... maybe so you can override the list in the unit tests. Blah.

lib/world/provider.rb
mdk6jd: Yes, because the `hospital` attribute is in `Person`.  Should I add some sort of catch for `Entity` objects that aren't also `Person`?

lib/records/hospital_exporter.rb
mdk6jd: Good catch, thank you!

jawalonoski:
The ID that gets generated for the Athol hospital (for example) within a Patient record should be the same ID that gets generated for Athol hospital within `hospital_information.json`.

You may want to generate the UUID for each hospital a priori, so no matter where and when it gets exported it is consistent.

lib/world/sequential.rb
mdk6jd: That makes more sense; I've moved the loading into `Hospital` class

lib/records/hospital_exporter.rb
mdk6jd: `FHIR::Extensions` have been added

mdk6jd:
In regard to the comment about duplicate `hospital`'s:
`hospitals` is a hash of the closest providers to the entity for each of the service types while `current_provider` is a list of the providers for active encounters.  `current_provider` was added so that procedures/medication prescriptions occurring during encounters would know which provider that encounter occurred at (and increment the counts for utilization for the correct provider).  This is confusing so in the following commits, I have renamed `hospital` to `preferred_provider` and moved it into the `Entity` class.

kindofluke:
@mdk6jd can I work with you to document this feature? it would be something we would want to use but didn't even know it was there.

In other words, I will write the documentation for the wiki but just would love you to review it. Would you be able to send me some of the sample healthcare facility json files you used? 

thanks

jawalonoski:
@kindofluke That particular file is now in the `master` branch. See https://github.com/synthetichealth/synthea/blob/master/config/healthcare_facilities.json.

After running `bundle exec rake synthea:generate`, some of the information will also be exported (in FHIR STU3) as `./output/fhir/hospital_information.json` -- including some utilization numbers.

kindofluke:
Ok great, I will update to master and try it out. and write it out.

Issue #208, opened 2017-06-23
First names often don't correspond with patient's gender


gotdan:
I think this may be related to the following 2016 pull request for Faker: https://github.com/stympy/faker/pull/551

jawalonoski:
That looks like a really old pull request on faker.... not sure that will ever get merged for us to take advantage of.

gotdan:
Yup, that was my assessment too. What are your thoughts on how the best way to handle it - I suppose Synthea could fork faker, patch it, or use a custom name generation function based off of the pull request?

jawalonoski:
Well, here are my thoughts...

- I'm not really interested in maintaining our own fork of faker.
- We could use the existing forked repo that the pull request is based on, but it doesn't appear maintained...
- I'd be willing to entertain a pull request on a better name generation function...

jawalonoski:
Fixed with #212

PR #212, opened 2017-07-05
support gender specific first names

gotdan:
Lightweight fix for #208 to match patient's first names to their gender


jawalonoski:
Thanks for the PR!

PR #216, opened 2017-07-17
Quality of life

mdk6jd:
This PR introduces calculations for DALYs (disability adjusted life years) and QALYs (quality adjusted life years) for patients.  Disability weights are taken from the Global Burden of Disease Study 2015 (http://ghdx.healthdata.org/gbd-2015) and are used to calculate DALY.  These disability weights are mapped to conditions using SNOMED codes and stored in `resources/gbd_disability_weights.json`. 

DALY is calculated on a yearly timestep, after the patient has been generate, by qol.rb.  The calculate method also can take an `end_time`, which is the time during a patient’s life that the quality of life values should be calculated.  The `end_date` defaults to the end of the simulation.

> Disability-Adjusted Life Year = DALY = YLL + YLD
>  Years of Life Lost = YLL = (1) * (standard life expectancy at age of death in years)
> Years Lost due to Disability = YLD = (disability weight) * (average duration of case)
> 
> Quality-Adjusted Life Year = QALY = age - YLD

The resulting values are stored as observations occurring at the `end_date`.  They are exported by the fhir, fhir_dstu2, test, and csv exporters.  ‘Quality adjusted life year’ SNOMED code is used for QALY and ‘Disability rating scale’ SNOMED code is used for DALY since specific codes do not exist.

### DIFF
--- a/lib/records/fhir.rb
+++ b/lib/records/fhir.rb
@@ -426,30 +426,7 @@ module Synthea
       end
 
       def self.quality_of_life_observation(observation, fhir_record, patient)
-        qol_data = QOL_CODES[observation['type']]
-        entry = FHIR::Bundle::Entry.new
-        resource_id = SecureRandom.uuid
-        entry.fullUrl = "urn:uuid:#{resource_id}"
-
-        entry.resource = FHIR::Observation.new('id' => resource_id,
-                                               'status' => 'final',
-                                               'code' => {
-                                                 'coding' => [{ 'system' => 'http://snomed.info/sct', 'code' => qol_data[:codes]['SNOMED-CT'][0], 'display' => qol_data[:description] }],
-                                                 'text' => qol_data[:description]
-                                               },
-                                               'category' => {
-                                                 'coding' => [{ 'system' => 'http://hl7.org/fhir/observation-category', 'code' => observation['category'] }]
-                                               },
-                                               'subject' => { 'reference' => patient.fullUrl.to_s },
-                                               'effectiveDateTime' => convert_fhir_date_time(observation['time'], 'time'),
-                                               'issued' => convert_fhir_date_time(observation['time'], 'time'))
-        entry.resource.valueQuantity = FHIR::Quantity.new('value' => observation['value'], 'unit' => qol_data[:unit], 'code' => qol_data[:unit], 'system' => 'http://unitsofmeasure.org/')
-
-        if Synthea::Config.exporter.fhir.use_shr_extensions
-          entry.resource.meta = FHIR::Meta.new('profile' => ["#{SHR_EXT}shr-observation-Observation"]) # all Observations are Observations
-        end
-
-        fhir_record.entry << entry
+        Synthea::Output::FHIRVersionIndependent.quality_of_life_observation(observation, fhir_record, patient, FHIR)
       end
 
       def self.multi_observation(multi_obs, fhir_record, patient, encounter)
--- a/lib/records/fhir.rb
+++ b/lib/records/fhir.rb
@@ -16,12 +16,6 @@ module Synthea
           # if an encounter doesn't have an end date, either the patient died during the encounter, or they are still in the encounter
           [:conditions, :observations, :procedures, :immunizations, :careplans, :medications].each do |attribute|
             entry = synthea_record.send(attribute)[indices[attribute]]
-
-            # edge case when encounter is active at end_date of synthea, don't want DALY/QALY calculation to be associated with encounter
-            if entry && attribute == :observations
-              next if entry['type'] == :DALY || entry['type'] == :QALY
-            end
-
             while entry && entry['time'] <= encounter_end
               method = entry['fhir']
               method = attribute.to_s if method.nil?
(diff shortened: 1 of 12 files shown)


lib/records/fhir.rb
```diff
@@ -2,6 +2,7 @@ module Synthea
   module Output
     module FhirRecord
       SHR_EXT = 'http://standardhealthrecord.org/fhir/StructureDefinition/'.freeze
+      SYNTHEA_EXT = 'http://synthetichealth.github.io/synthea/'.freeze
```
eedrummer: I see `SYNTHEA_EXT` declared, but I don't see it used in any of the code in this PR. Am I missing it?

lib/records/fhir_dstu2.rb
```diff
@@ -1,6 +1,8 @@
 module Synthea
   module Output
     module FhirDstu2Record
+      SYNTHEA_EXT = 'http://synthetichealth.github.io/synthea/'.freeze
```
eedrummer: Again, not sure if this is used anywhere.

lib/records/fhir_dstu2.rb
```diff
@@ -332,6 +353,29 @@ def self.observation(observation, fhir_record, patient, encounter)
         fhir_record.entry << entry
       end
 
+      def self.quality_of_life_observation(observation, fhir_record, patient)
```
eedrummer: I realize you are following the same pattern between the FhirRecord and FhirDstu2Record, but the code here is mostly the same. It would be great if you could refactor this code so that it reused most of what is in FhirRecord. That pattern could then be repeated across the rest of the code in this file.

lib/records/qol.rb
```diff
@@ -0,0 +1,82 @@
...
+        yld = 0
+        age = person[:age]
+        birth_time = person.events.events[:birth][0].time
+        age = ((end_date - birth_time) / 31_557_600).round if end_date != Synthea::Config.end_date
+
+        if person.had_event?(:death)
+          # life expectancy equation derived from IHME GBD 2015 Reference Life Table
+          # 6E-5x^3 - 0.0054x^2 - 0.8502x + 86.16
+          # R^2 = 0.99978
+          l = (0.00006 * age * age * age) - (0.0054 * age * age) - (0.8502 * age) + 86.16
```
eedrummer: Please use the `**` operator instead of repeatedly multiplying the variable.

lib/records/qol.rb
```diff
@@ -0,0 +1,82 @@
...
+      def conditions_in_year(conditions, year_start, year_end)
+        conditions_in_year = []
+        conditions.each do |condition|
+          if @disability_weights.key?(condition['type'].to_s)
+            condition_start_time = condition['time']
+            condition_end_time = condition['end_time']
+            condition_end_time = Synthea::Config.end_date if condition_end_time.nil?
+            conditions_in_year << condition if year_start >= condition_start_time && condition_start_time < year_end && condition_end_time > year_start
+          end
+          next
```
eedrummer: Does this `next` do anything?

mdk6jd: No, it does not.  I added this because it was a Rubocop suggestion/error.  Is there a work-around for it?

eedrummer: Rubocop probably wants you to do `next unless @disability_weights.key?(condition['type'].to_s)` at the start of the loop. You could do that, or I would support getting rid of that Rubocop rule because I think it is crazy.

lib/records/fhir.rb
mdk6jd: I used this for a previous implementation to add the values into the fhir record, but it is no longer used.  I will remove them.

lib/records/fhir_dstu2.rb
mdk6jd: Do you mean calling the FhirRecord method from fhir_dstu2? The code is the same, but `entry`, `entry.resource`, and `entry.resource.valueQuantity` are all `FHIR::DSTU2` types as oppose to just `FHIR`.

eedrummer: Right. In ruby, module names are just constants. They can be assigned to variables and passed around. So you could create a new class, maybe something like `FHIRVersionIndependentRecord`. In the class, you can make your method `self.quality_of_life_observation(observation, fhir_record, patient, fhir_package)`. Then, in the code you can do things like `entry = fhir_package::Bundle::Entry.new`. Finally, in FhirRecord, you could have:

```ruby
def self.quality_of_life_observation(observation, fhir_record, patient)
  FHIRVersionIndependentRecord.quality_of_life_observation(observation, fhir_record, patient, FHIR)
end
```

You could repeat this for all of the duplicated code in FhirRecord and FhirDstu2Record. Although all of those changes would be for another pull request.

jawalonoski: Except the names of the fields are different in the versions, with different cardinalities sometimes. So, that is a big problem.

eedrummer: Yeah. I figured, otherwise, you would just collapse this down to one class. That said, for code where you can, pull it out into the version independent file. For places where you need variation, you could leave the independent code alone or do something like:

```ruby
def self.do_something(thing, fhir_record, patient)
  entry = FHIRVersionIndependentRecord.do_something(thing, fhir_record, patient, FHIR)
  entry.tweak_thing_for_this_version
  fhir_record.entry << entry
end
```

I don't know how you feel about this approach, but it just struck me how similar the two files are.

Issue #206, opened 2017-06-20
Overall US Demographics


Erstwild:
Is there a best way to change the default settings to represent the US national overall demographics?  Would I just modify a provided county to the national demographics in a pinch?

jawalonoski:
Yes, one way would be to copy and modify one of the county files.

However, if no county file is provided, default demographics come from `lib/world/demograhics.rb`

If you model an area outside of Massachusetts, and you care about the accuracy of generated addresses, then you'll probably also have to modify `lib/world/city_zip.json` for zip codes and `lib/world/MA_geo.rb` for latitude and longitudes.

jakejang91:
I am trying to build up the model that can represents whole county of the Indiana. 
I have gathered and put all required files in "resource" folder, and changed a code in taks.rake to use those files. However, I am getting error messages saying that "rake undefiend method'[]' for nil:NilClass."  

Is this because data is conflicting with those in lib/world/MA_geo.rb and lib/world/city_zip.json? 

Can you suggest me a way to generate IN_geo.rb just like MA_geo.rb?

jawalonoski:
Not sure. Can you post a stack trace? Maybe even the input files too.

Question -- do you require lat/lon for each address?

jakejang91:
I was trying to use Synthea:census to make json files of counties in Indiana, and wanted to further broaden the boundary to all counties in the USA. I do not require lat/lon for each address at this point. 
I am posting this trace photo, and also files that I was using. 

![error](https://user-images.githubusercontent.com/32156141/33380447-1fc9db66-d4e9-11e7-919d-c878a110ffbe.JPG)
[resources.zip](https://github.com/synthetichealth/synthea/files/1514161/resources.zip)

jawalonoski:
Huh... Indianapolis census data is strange. Looks like towns can exist in multiple counties? This makes filtering out rollup data difficult... looking into it....

jawalonoski:
@jakejang91 I created a new branch called `other_usa_states` that should work. You'll get data for Indiana (if you drop your files in the `./resources` folder).

Some remaining issues -- you won't get proper zip codes (they'll show up as `"XXXXX"`) without further changes and you won't get lat/lons in the `Patient` resources.

https://github.com/synthetichealth/synthea/tree/other_usa_states

nyquist212:
Can you direct me to some resource describing how to create these County census files?

[...]

Issue #131, opened 2017-01-15
Instructions for Other Geographic Areas


jawalonoski:
If you wanted to configure Synthea to run for an area in the Netherlands or England, what would you need to do? 
- How to configure the area demographics? 
- How to change prevalences and incidence? 
- How to turn off US-only features such as race and ethnicity?

jawalonoski:
This should be fixed with PR #245

PR #245, opened 2017-12-01
Other usa states

jawalonoski:
See #206 

You can now drop in new 2010 census data files in `./resources` for states other than Massachusetts.

Remaining issues: you won't get proper zip codes (they'll show up as `"XXXXX"`) without further changes and you won't get lat/lons in the `Patient` resources.


jawalonoski:
I tested this with data for Indiana posted here https://github.com/synthetichealth/synthea/files/1514161/resources.zip from #206

jawalonoski:
These changes should more or less address #131, #206, and #246 by including a `resources/demographics.csv` file that includes all demographics for the entire United States.

If you want to generate another country (e.g. Netherlands or the United Kingdom), you can do it by replacing `resources/demographics.csv` with a CSV file. The structure of the CSV file will be documented in the wiki here.

dehall:
Why are so many cities in the demographics.csv file missing the income and education columns? We have that data for every city in MA in config/towns.json so it must be in the census data

dehall:
A couple other notes, that I don't think can or should be fixed immediately now but should be pointed out:

 - All provider names are still MA provider names. The provider addresses don't seem to be exported in the patient FHIR so it's not obvious but it's still only MA providers. Not a major issue. 
 - A good number of the incidence and prevalence numbers that we use are based on MA data, so the statistics will not necessarily match up to the state's real data. (Ex. diabetes prevalence - MA is 8.0%, CO is 6.4%, MS is 13.6% -- https://gis.cdc.gov/grasp/diabetes/DiabetesAtlas.html )   We'll need to keep this in mind as we update the modules and engine.

jawalonoski:
Regarding missing data: it seems to  do with regional variation in the census areas. I can get the data for all of those cities, just not easily. Currently I'm working of a single geography entitled "All Places within United States and Puerto Rico" which allows me to download a giant file.. to get the missing information (e.g. Acton, MA), I'd have to look at each State and merge them in.

dehall:
Ok. Overall this looks good, with the noted limitations. There are also publicly available data sources with zip codes and city boundaries, though the GeoJSON for the boundaries of all cities in the US is > 200 MB and maybe not something we want to include directly. (For comparison the MA GeoJSON we include currently is ~3 MB)

PR #261, opened 2018-02-09
WIP: SPEW data into Synthea

Abdulrahims:
Generating people based on open-source demographic data from SPEW (Synthetic Populations and Ecosystems of the World). This PR adds the following data to the synthetic population in Synthea:
* Household size
* Household income
* Person relationship to household
* Hispanic origin
* Nativity
* Employment status
* School enrollment
* Grade level currently attending, if in school
* Occupation
* Worldwide birthplaces

Other SPEW variables to fully incorporate:
* Location based on coordinates (latitude and longitude)
* Workplaces
* Schools (Public or private, names, grades, and size)


src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,293 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+		  person.attributes.put(Person.SCHOOL_ENROLLMENT, "no");
+	  } else if (school_enrollment.equals("2")){
+		  person.attributes.put(Person.SCHOOL_ENROLLMENT, "public_school_or_public_college");
+	  } else if (school_enrollment.equals("3")){
+		  person.attributes.put(Person.SCHOOL_ENROLLMENT, "private_school_or_college_or_home_school");
+	  } 
+
+	  String grade_level = spewPerson.get(rand_spew).get("SCHG");
+
+	  if(grade_level.equals("NA")) {
```
dehall: this giant if-statement is kind of hard to follow, could be extracted to some other kind of map, or at least a switch statement

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,293 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
   }
 
   private long setDemographics(Person person, Demographics city) {
-    person.attributes.put(Person.CITY, city.city);
```
dehall: Because the full SPEW dataset is too large to upload to github, I think we want to be able to choose to run either using SPEW or using the existing demographics format. At a glance I don't think this should be too hard, just have a config setting and choose between the 2

jawalonoski: Agreed.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,293 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
-    return birthdate;
+	  // Create map and read in the sampled SPEW csv file for Massachusetts 
+
+	  List<LinkedHashMap<String, String>> spewPerson;
+
+	  try {
+		  spewPerson = SimpleCSV.parse(Utilities.readResource("people25.csv"));
+	  } catch (IOException e) {
+		  e.printStackTrace();
+		  return (Long) null;
```
jawalonoski: What is the behavior if this actually happens?

E.g. if you delete `people25.csv` and run this program, what happens?

I'm not convinced swallowing the exception and returning `null` makes sense.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,244 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
-    return birthdate;
+	  // Create map and read in the sampled SPEW csv file for Massachusetts 
+
+	  List<LinkedHashMap<String, String>> spewPerson;
+
+	  try {
+		  spewPerson = SimpleCSV.parse(Utilities.readResource("people_25.csv"));
+	  } catch (IOException e) {
+		  e.printStackTrace();
+		  return (Long) null;
```
jawalonoski: What is the behavior of the program if this actually happens?

E.g. if you delete `people_25.csv` and run the program what happens?

I'm not convinced that swallowing the exception and returning `null` makes sense.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,244 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+	  //TODO hispanic ethnicities that are in SPEW but not in synthea
+
+	  List<LinkedHashMap<String, String>> hispanic_codes;
+
+	  try {
+		  
+		  hispanic_codes = SimpleCSV.parse(Utilities.readResource("hispanic.csv"));
+	  } catch (IOException e) {
+		  e.printStackTrace();
+		  return (Long) null;
```
jawalonoski: Again... not sure this behavior makes sense. What would happen here?

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,244 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+		  person.attributes.put(Person.NATIVITY, "foreign_born");
+	  }
+	
+	  List<LinkedHashMap<String, String>> birthplaces;
+	  
+	  try {
+		  birthplaces = SimpleCSV.parse(Utilities.readResource("birthplaces.csv"));
+	  } catch (IOException e) {
+		  e.printStackTrace();
+		  return (Long) null;
```
jawalonoski: Again.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,244 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+	  } 
+	  
+	  List<LinkedHashMap<String, String>> grade_level;
+	  
+	  try {
+		  grade_level = SimpleCSV.parse(Utilities.readResource("grade_level.csv"));
+	  } catch (IOException e) {
+		  e.printStackTrace();
+		  return (Long) null;
+	  }
```
jawalonoski: Again.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +344,244 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+		  } 
+	  }
+	  
+	  List<LinkedHashMap<String, String>> relationship;
+	  
+	  try {
+		  relationship = SimpleCSV.parse(Utilities.readResource("relationship.csv"));
+	  } catch (IOException e) {
+		  e.printStackTrace();
+		  return (Long) null;
```
jawalonoski: Again. Also, reading these files in for each person doesn't make a whole lot of sense performance wise. You should be caching these files.

src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -164,7 +164,7 @@ public static void birth(Person person, long time) {
       // should never happen in practice, but can happen in unit tests
       location.assignPoint(person, city);
       person.attributes.put(Person.ZIP, location.getZipCode(city, person.random));
-      attributes.put(Person.BIRTHPLACE, location.randomCityName(person.random));
+      //attributes.put(Person.BIRTHPLACE, location.randomCityName(person.random));
```
jawalonoski: To account for when we run in "demographics mode" versus "household mode" we may prefer this:
```java
if (!person.attributes.containsKey(Person.BIRTHPLACE) {
  person.attributes.put(Person.BIRTHPLACE, location.randomCityName(person.random));
}
```

src/main/java/org/mitre/synthea/engine/Generator.java
dehall: If the data's bad then I'd say halt the program. We'll release it with good data, so the only reason it should fail is if the user changes it. My suggestion is let the program crash, let them know why, and let them fix it.

jawalonoski: Why not just throw an exception then instead of returning null?

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -335,51 +351,257 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
-      gender = "F";
-    }
-    person.attributes.put(Person.GENDER, gender);
+	  if(Config.get("generate.households.mode").equals("true")) {
+		  System.out.println("Running in households mode");
+
+		  // Create map and read in the sampled SPEW csv file for Massachusetts 
+
+		  @SuppressWarnings("rawtypes")
+		  ArrayList<List> spewList = Demographics.getSpewList();
```
dehall: I realize there's probably no way you could have known to use Checkstyle, but from the diff here it's easy to spot that this isn't consistent with the code style in the rest of the code. I'll add this info to the contributors guide as well, but please set up Checkstyle using the config file located at ./config/checkstyle/checkstyle.xml . From there you'll see warnings in your IDE and can set up a formatter that will automatically format the code in the right way. I can help you get all that setup.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -334,52 +351,250 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
-    person.attributes.put(Person.ETHNICITY, ethnicity);
-    String language = city.languageFromEthnicity((String) person.attributes.get(Person.ETHNICITY),
-        person);
-    person.attributes.put(Person.FIRST_LANGUAGE, language);
-
-    String gender = city.pickGender(person.random);
-    if (gender.equalsIgnoreCase("male") || gender.equalsIgnoreCase("M")) {
-      gender = "M";
+  private long setDemographics(Person person, Demographics city) throws IOException {
+    if (Demographics.householdsMode() == "true") {
```
dehall: Compare strings using .equals() not == . This comparison might work out sometimes but it's generally not going to work how you expect. Also consider creating a new boolean field for this; take a look at the "onlyDeadPatients" field for an example.

dehall: Also, please add a unit test where householdsMode is set to true, just to make sure we have some unit test coverage. Again the "only dead patients" setting is a good example to copy - take a look at GeneratorTest.testGenerateOnlyDeadPatients()

src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -165,6 +165,11 @@ public static void birth(Person person, long time) {
       // should never happen in practice, but can happen in unit tests
       location.assignPoint(person, city);
       person.attributes.put(Person.ZIP, location.getZipCode(city));
+      
+      if (!person.attributes.containsKey(Person.BIRTHPLACE)){
+        person.attributes.put(Person.BIRTHPLACE, location.randomCityName(person.random));
```
dehall: Looks like these added lines are duplicates of the lines above and below

src/main/java/org/mitre/synthea/world/geography/Demographics.java
```diff
@@ -39,7 +41,49 @@
   private RandomCollection<String> incomeDistribution;
   public Map<String, Double> education;
   private RandomCollection<String> educationDistribution;
+  
+
+  public static List<LinkedHashMap<String, String>> getSpewPeople() throws IOException {
+    List<LinkedHashMap<String, String>> spewPerson = SimpleCSV
```
dehall: If these are going to be called for each person, we don't want to be reading and parsing the CSV each time, that would be really slow. It would be better to store them in a constant so that they can be re-used many times. Take a look at LifecycleModule.growthChart for an example. (though the name is actually a bad example, it's a constant so it should be ALL_UPPER_CASE)

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -214,8 +230,9 @@ public Person generatePerson(int index) {
    * @param personSeed
    *          Seed for the random person
    * @return generated Person
+ * @throws Throwable 
    */
-  public Person generatePerson(int index, long personSeed) {
+  public Person generatePerson(int index, long personSeed) throws Throwable {
```
dehall: Generally it's not recommended to say "throws Throwable" because that doesn't really convey any information about what could go wrong. Can you make this a more specific Exception?    (That said, I do think in general we've been bad about exceptions and should do a global cleanup)

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -304,7 +367,14 @@ public Person generatePerson(int index, long personSeed) {
     } catch (Throwable e) {
       // lots of fhir things throw errors for some reason
       e.printStackTrace();
-      throw e;
+      try {
+        return generatePerson(index, personSeed);
```
dehall: I don't think this was meant to be here - if you catch an exception and call the function again you'll hit the same exception and wind up in an infinite loop

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -334,52 +404,241 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+      // Longitude and latitude
+      // TODO a look up to assign address/city/town/zip from lat and long
+      // look into using FIPS codes
+      
+      String latitude = spewPerson.get(randSpew).get("latitude");
+
+      String longitude = spewPerson.get(randSpew).get("longitude");
+       
+      String coordinates = new StringBuilder(latitude).append(",").append(longitude).toString();
+      person.attributes.put(Person.COORDINATE, coordinates);
```
dehall: the COORDINATE attribute is expected to be a "DirectPosition2D" object - see Person.getLatLon(). This is getting overwritten by LifecycleModule.birth() before anything uses it though so that's why it's not crashing with a ClassCastException

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -334,52 +404,241 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
-      gender = "M";
+  private long setDemographics(Person person, Demographics city) throws IOException {
+    if (householdsMode.equals("true")) {
+      // get a random spew person
+
+      int[] range = new int[] { 1, spewPerson.size() + 1 };
+
+      int randSpew = (int) person.rand(range);
+
+      String spewSerial = spewPerson.get(randSpew).get("SERIALNO");
```
dehall: `spewPerson.get(randSpew)` is repeated a dozen times here - it should probably be extracted to a local variable

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -81,21 +91,57 @@ public Generator(int population) {
...
+ * @throws IOException 
    */
-  public Generator(GeneratorOptions o) {
+  public Generator(GeneratorOptions o) throws IOException {
     String state = o.state == null ? DEFAULT_STATE : o.state;
     init(o.population, o.seed, state, o.city);
   }
+  
+  @SuppressWarnings("rawtypes")
+  private static Map relationships() {
```
dehall: Don't think this was intended to be here - it's called relationships() but references the cdc_growth_charts file

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -334,52 +404,241 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+
+      if (nativity.equals("1")) {
+        person.attributes.put(Person.NATIVITY, "native");
+      } else if (nativity.equals("2")) {
+        person.attributes.put(Person.NATIVITY, "foreign_born");
+      }
+
+      //Birthplace
+      for (int i = 0; i <= birthplaces.size() - 1; i++) {
+        if (spewPerson.get(randSpew).get("POBP").equals(birthplaces.get(i).get("Code"))) {
```
dehall: A lot of these strings aren't super clear what they mean, ex what does "POBP" mean? Could they be extracted to constants with more meaningful names?

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -334,52 +382,242 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+        if (randSpew.get("OCCP").equals(occupations.get(i).get("Code"))) {
+          person.attributes.put(Person.OCCUPATION, occupations.get(i).get("Occupation"));
+        }
+      }
+
+      String employmentStatus = randSpew.get("ESR");
+
+      if (employmentStatus.equals("NA")) {
+        person.attributes.put(Person.EMPLOYMENT_STATUS, "na_under_16");
+      } else if (employmentStatus.equals("")) {
```
dehall: Just realized these strings are all empty and need to be filled in with the right codes.

The last thing I'll mention is that these big if/else checks can be kind of ugly. There are a couple different ways to do these but fundamentally it's a judgment call so do whatever you prefer.

One way is to use a `switch` statement which you can do on Strings since Java 7 I think:

```java
      String employmentStatusCode = spewPerson.get(randSpew).get("ESR");
      String employmentStatusDesc;
      switch (employmentStatusCode) {
      case "NA":
        employmentStatusDesc = "na_under_16";
        break;
      case "1":
        employmentStatusDesc = "civilian_employed_at_work";
        break;
      case "2":
        employmentStatusDesc = "civilian_employed_with_job_but_not_at_work";
        break;
        // ... etc
        default:
          throw new IllegalArgumentException("Unexpected SPEW employment code: " + employmentStatusCode);
      }
      person.attributes.put(Person.EMPLOYMENT_STATUS, employmentStatusDesc);
```

Another way is to define the lookup table somewhere else and just call it, for an example see LifecycleModule.DIABETES_DRUG_HBA1C_IMPACTS . Then you just can do something along the lines of
```java
    String employmentStatusCode = spewPerson.get(randSpew).get("ESR");
    String employmentStatusDesc = SPEW_EMPLOYMENT_STATUS_CODES.get(employmentStatusCode);
    person.attributes.put(Person.EMPLOYMENT_STATUS, employmentStatusDesc);
```

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -334,52 +402,231 @@ private synchronized void writeToConsole(Person person, int index, long time, bo
...
+      person.attributes.put(Person.CITY, city.city);
+
+      // TODO spew location changes
+      person.attributes.put(Person.STATE, city.state);
+
+      String race = randSpew.get("RAC1P");
+      String hisp = randSpew.get("HISP");
+
+      // race codes, hispanic is a different variable
+      if (race.equals("1")) {
```
dehall: I had one last thought. There's an odd combination of translations/code mappings used here. For example this is a map from (1..9) to the various races, using an if-else branch. But other places read from CSV files and iterate over the possible values, and certain others use regular lookup tables. It would be a lot cleaner if all of these were moved into individual CSVs and then converted into regular lookup tables. For example here it would be a lot easier if you had a Map<String, String> containing `{ "1" => "white", "2" => "black", "3" => "native", etc... }` and then you just say `person.attributes.put(Person.RACE, map.get(race));` in one line. 

My suggestion for a CSV-to-lookup table is a Utilities function like the following:
```
  /**
   * Convert the parsed CSV into a map of key column => value column.
   * Given a List of Map&lt;String,String&gt;, this method "rotates" it such that
   * a lookup can be done from the key column -> value column.
   * For example, given the following CSV: <pre>
   * CODE,DEFINITION
   * 1,AAA
   * 2,BBB
   * 3,CCC</pre>
   * This function may be called with keyColumn="CODE" and valueColumn="DEFINITION" to produce
   * the following: { "1" => "AAA", "2" => "BBB", "3" => "CCC" }
   * Any other columns in the given map are ignored.
   * @param parsedCsv List of data parsed from a CSV file
   * @param keyColumn Title of the column containing the "key"
   * @param valueColumn Title of the column containing the "value"
   * @return Map of keyColumn => valueColumn
   */
  public static final Map<String, String> createMapFromCsv(
      List<LinkedHashMap<String, String>> parsedCsv, String keyColumn, String valueColumn) {
    Map<String, String> map = new HashMap<>();

    for (Map<String, String> line : parsedCsv) {
      String key = line.get(keyColumn);
      String value = line.get(valueColumn);
      map.put(key, value);
    }

    return map;
 }
```

And a free unit test:
```
  @Test
  public void testCreateMap() throws IOException {
    List<LinkedHashMap<String,String>> parsedCsv =
        SimpleCSV.parse(Utilities.readResource("sampleCsv.csv"));

    Map<String,String> map = Utilities.createMapFromCsv(parsedCsv, "CODE", "DESCRIPTION");

    assertEquals("Examplitis", map.get("123"));
    assertEquals("Examplitol", map.get("456"));
    assertEquals("Examplotomy Encounter", map.get("ABC"));
    assertEquals("Examplotomy", map.get("789"));
 }
```

With a new test resource:  sampleCsv.csv
```
CODE,DESCRIPTION
123,Examplitis
456,Examplitol
ABC,Examplotomy Encounter
789,Examplotomy
```

jawalonoski:
Thank you for the pull request, but it is being closed due to change in project direction.

PR #262, opened 2018-02-11
Performance improvements

dehall:
Improves generation performance by about 50% (or maybe more accurately, reduces the amount of time generation takes by 50%). Note that this only affects _generation_ time not _export_ time. That will be reviewed later. This is done with the following changes:

- Removed the Faker library. The problem with Faker is that it reads from a YML file every time the constructor is called. We were using a Faker per-person to ensure consistent and repeatable random results, and there doesn't seem to be a way to share that resource in a way that doesn't break our seeded randomness. Replaced Faker with a smaller set of random names & addresses. We actually used a similar approach in the Ruby version to get gender-specific names, so I used that same file and added a list of address types from wikipedia.

- Cached VitalSign names. Looking up VitalSigns was slow because it had to loop over every item and do reflection to get the name from the annotation. Now it's still a loop but it's a cached string.

- Cached Biometrics Config entries that are used by LifecycleModule.diabeticVitalSigns


src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -175,6 +193,50 @@ public static void birth(Person person, long time) {
...
+      choices = (List<String>) names.get("english.family");
+    }
+    // pick a random item from the list
+    return choices.get(random.nextInt(choices.size()));
+  }
+  
+  @SuppressWarnings("unchecked")
+  private static String fakeAddress(boolean includeLine2, Random random) {
+    int number = random.nextInt(1000) + 100;
+    List<String> n = (List<String>)names.get("english.family");
```
jawalonoski: We'll have to fix this long-term. Doesn't make sense. There are so many Main Streets and "Oak St" and stuff like that... not too many "Johnson Lanes" out there, is my guess.

dehall: Agreed, but note that this is actually roughly the approach faker uses as well:
https://github.com/DiUS/java-faker/blob/master/src/main/resources/en.yml#L34

They also include first names, but still they only use the set of people's names and nothing street-specific.

PR #265, opened 2018-02-13
Add some more unit tests.

jawalonoski:
Adds a unit test on `App.java`, adds a disabled unit test on `Graphviz.java`, and a handful of logic tests.

### DIFF
--- a/src/main/java/Graphviz.java
+++ b/src/main/java/Graphviz.java
@@ -48,8 +48,8 @@ public class Graphviz {
 
     Path inputPath = null;
     if (args != null && args.length > 0) {
-      URI uri = new URI("file://" + args[0]);
-      inputPath = Paths.get(uri);
+      File file = new File(args[0]);
+      inputPath = file.toPath();
     } else {
       URL modulesFolder = ClassLoader.getSystemClassLoader().getResource("modules");
       inputPath = Paths.get(modulesFolder.toURI());
--- a/src/main/java/Graphviz.java
+++ b/src/main/java/Graphviz.java
@@ -18,6 +18,8 @@ import guru.nidi.graphviz.model.Node;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Files;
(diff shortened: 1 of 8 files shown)


jawalonoski:
Huh... it looks like one build passed and the other failed. Can't figure out why that assertion is failing on Travis but not locally...

dehall:
Have you tried restarting the failed build? I've noticed our tests can be kind of flaky sometimes and fail once but pass the next 10 times.

jawalonoski:
Manual merge with master builds and tests successfully locally.

src/main/java/Graphviz.java
```diff
@@ -37,29 +39,39 @@
...
+   * @param args Optional path of modules to render. If not provided,
+   *   the default modules will be loaded using the ClassLoader.
+   * @throws URISyntaxException on failure to load modules.
    */
-  public static void main(String[] args) {
+  public static void main(String[] args) throws URISyntaxException {
     File folder = Exporter.getOutputFolder("graphviz", null);
 
+    Path inputPath = null;
+    if (args != null && args.length > 0) {
```
dehall: I like this idea of passing in an argument to specify where to load the modules from, especially if we can eventually extend it to only load 1 module at a time, but:
1. It's not clear to me how to call it outside of an IDE. Gradle needs more config to pass command line arguments to a task, for example see the "Params" thing on the run task
2. I'm not sure what format to use to pass in a folder in Windows. I tried both "C:\folder1\etc" and "/c/folder1/etc" and neither worked.
I'd suggest using this approach which should work just as well for Mac paths and accept Windows-style paths as well:
```
      File file = new File(args[0]);
      inputPath = file.toPath();
```

jawalonoski: Regarding 1, it is difficult. I didn't modify the gradle task, figuring we could do it later if when we wanted to enable that. I was running it through the Eclipse run configuration, but the principle should be the same as the `run_synthea` script.

Regarding 2, it is a trying to create a URI, so `file://...` So, for example, this worked for me:

```
/Users/jwalonoski/git/synthea/src/main/resources/modules/medications
```

Regardless, I'll make the suggested change.

PR #266, opened 2018-02-16
Costs, v2

dehall:
Updated cost data, based on HCUP costs for procedures and medications. This includes only costs for codes that are currently used in Synthea, so this will need to be updated as we add more concepts. Costs were determined by taking the latest version of concepts.csv and manually mapping code descriptions against the closest match from HCUP data. The procedure costs are orders of magnitude higher than the medication cost, but those are the numbers they report.

Primary source for medication costs:
(note that this data file is ~ 2 GB)
https://data.cms.gov/Medicare-Part-D/Medicare-Provider-Utilization-and-Payment-Data-201/465c-49pb

Primary source for procedure costs:
https://hcupnet.ahrq.gov/#query/eyJBTkFMWVNJU19UWVBFIjpbIkFUX1EiXSwiWUVBUlMiOlsiWVJfMjAxNCJdLCJTVEFURVMiOlsiU1RfTUEiXSwiQ0FURUdPUklaQVRJT05fVFlQRSI6WyJDVF9DQ1NQIl0sIlFVSUNLVEFCTEVfVFlQRSI6WyJRVFRfREVUIl0sIkRBVEFTRVRfU09VUkNFIjpbIkRTX1NJRCJdfQ==


jawalonoski:
This is a good start. I thought the HCUP data included min, max, and average costs by region. Is the number you used the average national cost?

Eventually, we'll want to vary the costs by region and we may want to choose a random value between the min/max, to reflect the wild disparity in healthcare costs.

jawalonoski:
Also, we should probably have a different base cost for emergency encounters. So you may want to add another row there after I merge this in.

PR #275, opened 2018-03-07
Providers and Encounters

jawalonoski:
- Rework EncounterType a bit in preparation for other Provider types (e.g. REHAB, DIALYSIS).


src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
```diff
@@ -225,7 +225,11 @@ public BigDecimal cost() {
   }
 
   public enum EncounterType {
-    WELLNESS, EMERGENCY, INPATIENT, AMBULATORY
+    WELLNESS, EMERGENCY, INPATIENT, AMBULATORY;
+
+    public boolean equals(String type) {
```
dehall: This mixing of strings and enums doesn't feel great to me, especially this equals method. It's generally expected that for non-null x and y, x.equals(y) should be the same as y.equals(x), and that won't be the case here.

Could we completely switch over to just the enums and not pass strings around?

jawalonoski:
This PR is hopeless out of date.

PR #277, opened 2018-03-19
Use RxNorm codes only for Medications.

jawalonoski:
Use RxNorm codes only for Medications. Fixes dermatitis module prescription.


jawalonoski:
Possibly, but none left in the SNOMED category though.

PR #279, opened 2018-03-20
Add modules to concepts list.

jawalonoski:
This reworks the `Concepts` class a little bit. Drops the use of `Table` to use `Map`:

```java
Map<Code,Set<String>> concepts;
```

This allows us to easily track which concepts appear in which modules. Most codes only appear in a single module, but some appear in many.

I also added `Comparable<Code>` to `Code` so that codes could be sorted alphabetically by system and then by code.

### DIFF
--- a/src/main/java/org/mitre/synthea/helpers/Concepts.java
+++ b/src/main/java/org/mitre/synthea/helpers/Concepts.java
@@ -1,7 +1,5 @@
 package org.mitre.synthea.helpers;
 
-import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Table;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mitre/synthea/helpers/Concepts.java
```diff
@@ -50,7 +51,7 @@ public static void main(String[] args) throws Exception {
     
     Files.write(outFilePath, output, StandardOpenOption.CREATE);
```
dehall: There's an existing bug here where if you don't have the output directory (ex, if you clone the repo and immediately run `gradle concepts`), it will throw a FileNotFoundException. We should probably also use the configuration property `exporter.baseDirectory`. I'll clean that up in a later PR.

PR #280, opened 2018-03-20
Claims Revisited

jawalonoski:
This pull request fixes some oddities in Claims representations in FHIR and adds costs to CSV export.

The CSV costs are added directly into the encounters, immunizations, medications, and procedures csv tables. There is no separate claims table with aggregated costs.

The FHIR costs were also changed, as I noticed some weird things while looking at claims in general. Foremost is that now FHIR claims also include charges for immunizations. Second, is that DSTU2 claims were not linked to an Encounter... which doesn't make a whole lot of sense. The individual entries were also using example bindings, where it makes more sense to bind directly to the codes of the services provided (e.g. the procedure code).

### DIFF
--- a/src/main/java/org/mitre/synthea/export/CSVExporter.java
+++ b/src/main/java/org/mitre/synthea/export/CSVExporter.java
@@ -9,6 +9,7 @@ import java.nio.file.Path;
 import java.util.UUID;
 
 import org.mitre.synthea.world.agents.Person;
+import org.mitre.synthea.world.concepts.Costs;
 import org.mitre.synthea.world.concepts.HealthRecord;
 import org.mitre.synthea.world.concepts.HealthRecord.CarePlan;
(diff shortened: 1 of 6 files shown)


src/main/java/org/mitre/synthea/export/CSVExporter.java
```diff
@@ -438,7 +441,9 @@ private void procedure(String personID, String encounterID,
 
     s.append(coding.code).append(',');
     s.append(clean(coding.display)).append(',');
-    
+
+    s.append(String.format("%.2f", Costs.calculateCost(procedure, true))).append(',');
```
dehall: Rather than calculating the cost in the exporters, I'm wondering if it makes more sense to calculate the cost upfront and then just store it as a field in the Entry. But we can hold off on that until our cost model becomes more robust.

Issue #283, opened 2018-04-02
CSV Export


SimonUzL:
Hello,
I have created a PostgreSQL database to import the csv files produced by synthea. I used the schema given by the wiki entry "CSV File Data Dictionary". 
In Observations the entry value should be numeric and in CarePlans the reasoncode is not allowed to be null according to the wiki article. But i have found csv records which are violating these constraints.
In Observations there is a string "Natural death with unknown cause" for the column value which is numeric and in CarePlans there is a null reasoncode. 
Am I doing something wrong? Could that be a bug?

I can reproduce it with the command 

> ./run_synthea -p 100

 every time.

jawalonoski:
It is a bug... Either in the wiki or the exporter. Or both.

We'll have to investigate the CarePlan issue -- a patient shouldn't be assigned a careplan for no reason... although `CarePlanStart.reason` is optional (see https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-States#careplanstart) so maybe that is something we need to change to required.

Re: observations. Hmm. We'll need to think about that. Our primary exporter is FHIR, which technically allows a choice of datatypes for `Observation.value` -- it seems we have some that aren't just numerics, but it looks like the FHIR exporter code handles that correctly.

Suggestions?

jawalonoski:
Some fact finding.... Technically, the value for that observation is a code:
```json
"valueCodeableConcept": {
  "coding": [{
    "system": "http://snomed.info/sct",
    "code": "9855000",
    "display": "Natural death with unknown cause"
  }],
  "text": "Natural death with unknown cause"
}
```
When converted to CSV, only the `text` gets exported.

SimonUzL:
The FHIR standard allows different types for the value in Observations:

[https://www.hl7.org/fhir/observation-definitions.html#Observation.value_x_](url)

But it is not possible in a relational schema to have different data types in one column. My idea would be to have three columns to solve this problem: value, unit and additionally type. For example it could look like this (I have omitted the other columns in this example):

```
... other columns |      value       | unit  |  type   
------------------+------------------+-------+---------
...               | unexpected death |       | string
...               | 32.5             | kg/m2 | decimal
...               | 82.9             | mmHg  | decimal

 Column | Type | Modifiers 
--------+------+-----------
... other columns
 value  | text | not null  
 unit   | text |           
 type   | text | not null  
```

Maybe this is a solution? 

[...]

PR #285, opened 2018-04-03
issue #283 possible CSV export fix

SimonUzL:
This is a possible fix for issue #283. As far as I can see there are no more types than text and decimal. I ran the tests without problems.


jawalonoski:
Ideally we'd have some unit tests in there, which is what codecov is complaining about, but this looks sufficient.

PR #287, opened 2018-04-06
Add ImagingStudy State to Synthea

cjduffett:
## Hello There! 👋 

This PR adds the FHIR-inspired `ImagingStudy` state to the synthea runtime, GMF, and exporters.

Most IDEs (including mine) automatically trim trailing whitespace on save. As a result, some of this PR's delta is just whitespace. Viewing the non-whitespace changes gets much easier if you add `&w=1` to the query string at the end of these Github URLs, for example:

```URL
https://github.com/Ciitizen/synthea/compare/master...Ciitizen:carlton/add-fhir-imaging-study?expand=1&w=1
```
## The ImagingStudy State

An ImagingStudy represents a **study** that consists of one or more **series** of images.

Each series is a set of images taken of a specific `body_site` (for example, "Right Knee") using a specific `modality` (for example "Magnetic Resonance").

Each image taken is represented as an **instance** in the series, with a `title` and `sop_class` describing the DICOM Subject-Object Pair (SOP) that constitutes the image.

Altogether, an ImagingStudy state may look like:

```json
"Knee_MRI": {
  "type": "ImagingStudy",
  "series": [
    {
      "body_site": {
        "system": "SNOMED-CT",
        "code": "6757004",
        "display": "Right knee"
      },
      "modality": {
        "system": "DICOM-DCM",
        "code": "MR",
        "display": "Magnetic Resonance"
      },
      "instances": [
        {
          "title": "Image of right knee",
          "sop_class": {
            "system": "DICOM-SOP",
            "code": "1.2.840.10008.5.1.4.1.1.4",
            "display": "MR Image Storage"
          }
        }
      ]
    }
  ],
  "direct_transition": "End_ED_Visit"
}
```

## The Graphviz

A simple module containing an ImagingStudy state. An ImagingStudy **must** occur during an Encounter. For now, the state simply displays its "primary" body site and modality, taken from the first series in the study.

imaging_study

## The randomDicomUid

When exported, each study, series, and instance in an imaging study **must** have its own, unique DICOM UID. These IDs conform to the DICOM specification and follow the formats:

| Entity | Template | Example |
|:------|:---------|:----------|
| study | `1.2.840.<org_id>.<current_unix_time>` | `1.2.840.99999999.1522291594299` |
| series | `1.2.840.<org_id>.<series_no>.<current_unix_time>` | `1.2.840.99999999.1.1522291594300` |
| instance | `1.2.840.<org_id>.<series_no>.<instance_no>.<current_unix_time>` | `1.2.840.99999999.1.1.1522291594300` |

Where the `<org_id>` was arbitrarily chosen to be `99999999` and `<series_no>` and `<instance_no>` monotonically increase with the current series or instance, starting at 1.

## FHIR STU3

When exported in FHIR STU3, an ImagingStudy may look like:

```json
{
  "fullUrl": "urn:uuid:19f8c4ae-f921-4d27-9d65-0ff57fd7c826",
  "resource": {
    "resourceType": "ImagingStudy",
    "id": "19f8c4ae-f921-4d27-9d65-0ff57fd7c826",
    "uid": "urn:oid:1.2.840.99999999.1522027833636",
    "patient": {
      "reference": "urn:uuid:f5b8c7af-3b6c-4598-913f-f49f5c27234d"
    },
    "context": {
      "reference": "urn:uuid:d46cc23d-2523-45ed-ae73-1a2da9e3832f"
    },
    "started": "2015-03-21T00:11:50-07:00",
    "numberOfSeries": 1,
    "numberOfInstances": 1,
    "series": [
      {
        "uid": "urn:oid:1.2.840.99999999.1.1522027833636",
        "number": 1,
        "modality": {
          "system": "http://dicom.nema.org/resources/ontology/DCM",
          "code": "MR",
          "display": "Magnetic Resonance"
        },
        "numberOfInstances": 1,
        "availability": "UNAVAILABLE",
        "bodySite": {
          "system": "http://snomed.info/sct",
          "code": "6757004",
          "display": "Right knee"
        },
        "started": "2015-03-21T00:11:50-07:00",
        "instance": [
          {
            "uid": "urn:oid:1.2.840.99999999.1.1.1522027833636",
            "number": 1,
            "sopClass": "urn:oid:1.2.840.10008.5.1.4.1.1.4",
            "title": "Image of right knee"
          }
        ]
      }
    ]
  }
}
```

## FHIR DSTU2

When exported in FHIR DSTU2, an ImagingStudy may look like:

```json
{
  "resourceType": "ImagingStudy",
  "id": "71e41c44-3629-4586-a221-8b8fc3aff3e9",
  "started": "2013-08-22T06:12:42-07:00",
  "patient": {
    "reference": "urn:uuid:72fc7c7c-1f6e-4f81-97e4-04a774821b79"
  },
  "uid": "urn:oid:1.2.840.99999999.1522291594299",
  "numberOfSeries": 1,
  "numberOfInstances": 1,
  "series": [
    {
      "number": 1,
      "modality": {
        "system": "http://dicom.nema.org/resources/ontology/DCM",
        "code": "MR",
        "display": "Magnetic Resonance"
      },
      "uid": "urn:oid:1.2.840.99999999.1.1522291594300",
      "numberOfInstances": 1,
      "availability": "UNAVAILABLE",
      "bodySite": {
        "system": "http://snomed.info/sct",
        "code": "6757004",
        "display": "Right knee"
      },
      "started": "2013-08-22T06:12:42-07:00",
      "instance": [
        {
          "number": 1,
          "uid": "urn:oid:1.2.840.99999999.1.1.1522291594300",
          "sopClass": "urn:oid:1.2.840.10008.5.1.4.1.1.4",
          "title": "Image of right knee"
        }
      ]
    }
  ]
}
```

## C-CDA Diagnostic Imaging Report

In CDA, ImagingStudies are represented in a Diagnostic Imaging Report section, template ID `2.16.840.1.113883.10.20.22.1.5`. For example:

```xml
<component>
  <!--Imaging Studies-->
  <section classCode="DOCSECT" moodCode="EVN">
    <templateId root="2.16.840.1.113883.10.20.22.1.5"/>
    <code code="18748-4" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" displayName="Diagnostic Imaging Report"/>
    <title>Imaging Studies</title>
    <entry typeCode="COMP">
      <!-- Study -->
      <act classCode="ACT" moodCode="EVN">
        <templateId root="2.16.840.1.113883.10.20.6.2.6"/>
        <id root="1.2.840.99999999.1522300574147"/>
        <code code="113014" codeSystem="1.2.840.10008.2.16.4" codeSystemName="DCM" displayName="Study"/>
        <!-- Series -->
        <entryRelationship typeCode="COMP">
          <act classCode="ACT" moodCode="EVN">
            <id root="1.2.840.99999999.1.1522300574147"/>
            <code code="113015" codeSystem="1.2.840.10008.2.16.4" codeSystemName="DCM" displayName="Series">
              <qualifier>
                <name code="121139" codeSystem="1.2.840.10008.2.16.4" codeSystemName="DCM" displayName="Modality"> </name>
                <value code="MR" codeSystem="1.2.840.10008.2.16.4" codeSystemName="DCM" displayName="Magnetic Resonance"> </value>
              </qualifier>
            </code>
            <!-- Instances -->
            <entryRelationship typeCode="COMP">
              <observation classCode="DGIMG" moodCode="EVN">
                <templateId root="2.16.840.1.113883.10.20.6.2.8"/>
                <id root="1.2.840.99999999.1.1.1522300574147"/>
                <code code="1.2.840.10008.5.1.4.1.1.4" codeSystem="1.2.840.10008.2.6.1" codeSystemName="DCMUID" displayName="MR Image Storage"> </code>
                <effectiveTime value="20151003193818"/>
              </observation>
            </entryRelationship>
          </act>
        </entryRelationship>
      </act>
    </entry>
  </section>
</component>
```
References to these template elements:

* DICOM Object Catalog:  
http://bccda.art-decor.org/bccdapilot-html-20161231T015049/tmp-2.16.840.1.113883.10.20.6.1.1-2016-10-20T062037.html

* Study Act:  
http://bccda.art-decor.org/bccdapilot-html-20161231T015049/tmp-2.16.840.1.113883.10.20.6.2.6-2016-12-16T223748.html

* Series Act:  
http://bccda.art-decor.org/bccdapilot-html-20161231T015049/tmp-2.16.840.1.113883.10.20.22.4.63-2015-08-13T000000.html

* SOP Instance Observation:  
http://bccda.art-decor.org/bccdapilot-html-20161231T015049/tmp-2.16.840.1.113883.10.20.6.2.8-2015-08-13T000000.html

## CSV Export

Reflects only the "primary" body site and modality codes, for example:

| ID       | DATE       | PATIENT  | ENCOUNTER | BODYSITE_CODE | BODYSITE_DESCRIPTION | MODALITY_CODE | MODALITY_DESCRIPTION | SOP_CODE                  | SOP_DESCRIPTION  | 
|----------|------------|----------|-----------|---------------|----------------------|---------------|----------------------|---------------------------|------------------| 
| 15ff2c08 | 2009-10-12 | ebfbae04 | a7998499  | 6757004       | Right knee           | MR            | Magnetic Resonance   | 1.2.840.10008.5.1.4.1.1.4 | MR Image Storage | 

## Text Export

Similarly, reflects only the primary body site and modality codes:

```
--------------------------------------------------------------------------------
IMAGING STUDIES:
2014-02-27 : Magnetic Resonance, Right knee
--------------------------------------------------------------------------------
```


cjduffett:
👋  Hey @jawalonoski @dehall, I love what you've done with the place! Synthea's a helluva lot quicker!

The codecov report dropped because we don't yet have a GMF module with an ImagingStudy state. This leave the exporters without any ImagingStudies to export. We test deserialization of states well already (GMF JSON => synthea), but not the other way around (synthea => fhir/ccda/csv/text).

I think we should expand the exporter tests to validate that a patient's full record (or at least, each individual component of it) gets exported to fhir/ccda/csv/text correctly. That way we can validate that the exporters work correctly without needing to run many patients through the process. Should improve coverage and speed up tests nicely.

jawalonoski:
@cjduffett, this is awesome. There is a lot here to review here, but at first glance this looks like a great contribution.

Regarding tests -- we can add exporter tests as you suggest, but unless we were convinced we could cover all the corner cases (and I'm not sure we could), I would be inclined to keep the random generation exporter tests intact as a weak type of fuzz testing.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1223,6 +1223,33 @@ public boolean process(Person person, long time) {
     }
   }
 
+  /**
+   * The ImagingStudy state indicates a point in the module when an imaging study was performed.
+   * An ImagingStudy consists of one or more Studies, where each Study containes one or more
```
jawalonoski: "containes" should be "contains"

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
```diff
@@ -176,6 +177,41 @@ public CarePlan(long time, String type) {
...
+    public ImagingStudy(long time, String type) {
+      super(time, type);
+      this.series = new ArrayList<Series>();
+    }
+
+    /**
+     * ImagingStudy.Series represents a series of images that were taken of
+     * a specific part of the body.
+     */
+    public class Series {
```
dehall: My gut feeling is that this fits better as a "Component" rather than in the health record, but it's not critical to change it.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
```diff
@@ -176,6 +177,41 @@ public CarePlan(long time, String type) {
...
+      /* A SNOMED-CT body structures code */
+      @SerializedName("body_site")
```
dehall: Are these `@SerializedName`s necessary? The Gson setting is specified that field names get changed from camelCase variables to lower case with underscores in the serialized json.  https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/helpers/Utilities.java#L223  True it's more explicit but I don't think it's required in this case.


Also, re: the comments here, please make those javadocs.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
cjduffett: I took "Components" to be more generic JSON elements that are shared by multiple states. That seems true of all of the existing ones anyway. Since `Series` and `Instance` are really specific to just an `ImagingStudy` and no other states, I think it's best to leave them where they are.

src/main/java/org/mitre/synthea/export/FhirDstu2.java
```diff
@@ -1115,9 +1126,101 @@ private static Entry careplan(Entry personEntry, Bundle bundle, Entry encounterE
...
+   *          The ImagingStudy to map to FHIR and add to the bundle
+   * @return The added Entry
+   */
+  private static Entry imagingStudy(Entry personEntry, Bundle bundle, Entry encounterEntry,
+      ImagingStudy imagingStudy) {
+    ca.uhn.fhir.model.dstu2.resource.ImagingStudy imagingStudyResource =
+        new ca.uhn.fhir.model.dstu2.resource.ImagingStudy();
+
+    OidDt studyUid = new OidDt();
+    studyUid.setValue("urn:oid:" + Utilities.randomDicomUid(0, 0));
```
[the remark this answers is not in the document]
cjduffett: Yes, I think that's reasonable. I'll experiment with adding that to the constructors of each `ImagingStudy`, `Series`, and `Instance` object.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
cjduffett: Good catch on the `@SerializedName`, it is indeed unnecessary and I'll remove it. That was a leftover from my experimentation with the deserialization of ImagingStudy states.

cjduffett:
@dehall Thank you for the thorough review! I'll address your comments RE `randomDicomUid`, `@SerializedName`, and javadocs.

***

Holistically, an ImagingStudy is a complement to the existing Procedure, Observation, and DiagnosticReport states. While there are equivalent SNOMED codes for some imaging-type procedures (e.g. "MRI" = `113091000`), the ImagingStudy covers imaging procedures in more detail and allows more granular representation of what body sites were imaged and what images were taken, especially if multiple body sites or different modalities are required for a single study.

If you like, there is an optional place to capture `procedureCode` in a [FHIR ImagingStudy](https://www.hl7.org/fhir/imagingstudy.html), so perhaps we could also add the SNOMED `procedureCode` to the ImagingStudy state, or immediately precede ImagingStudy states with a Procedure state in the same Encounter?

It's also worth noting that the FHIR philosophy regarding Procedures is that:

> The Procedure resource should not be used to capture an event if a more specific resource already exists.

***

In the near future I plan to have Observation states directly following an ImagingStudy reflect a physician's observations about the results of that study. A [FHIR DiagnosticReport](https://www.hl7.org/fhir/diagnosticreport.html) actually has an explicit reference to an `imagingStudy`, so I image the combination of states:

```
ImagingStudy -> Observation(s) -> DiagnosticReport
```
Will work nicely and accurately reflect the way ImagingStudies and their results are meant to be captured.

cjduffett:
@dehall regarding `randomDicomUids`, I couldn't put the generation of the random IDs in the Series or Instance constructors because those objects are created at deserialization time and have no way of knowing what index they have in their respective lists. Instead, I consolidated all dicom UID assignment in `HealthRecord.imagingStudy()` where the UIDs are assigned immediately after the ImagingStudy is created and before it's added to the patient's record.

The FHIR and CCDA exporters now reference the same `dicomUid` for each ImagingStudy, Series, and Instance instead of generating new randomDicomUIDs every time.

I also added an extra 8-digit `salt` to each UID since the unix timestamp is not enough to guarantee uniqueness of UIDs between 2 ImagingStudies that are generated within the same millisecond.

cjduffett:
@jawalonoski In an effort to improve coverage of ImagingStudy export with minimal changes I added some simple X-rays to the injuries module. Given the less-than-likely probability of 10 patients having a recent injury that required an X-ray, this may have been a futile effort though.

Ideally I'd like to test each export method of the FHIR exporters, for example something like:

```java
@Test
private void test_imaging_study_export() {
  // mock a patientEntry + populate the properties that get used
  BundleEntryComponent patientEntry = new BundleEntryComponent();
  
  // mock a encounterEntry + populate the properties that get used
  BundleEntryComponent encounterEntry = new BundleEntryComponent();

  // mock a Bundle + populate the properties that get used
  Bundle bundle = new Bundle();

  // mock an ImagingStudy + populate the properties that get used
  ImagingStudy study = new ImagingStudy();

  BundleEntryComponent exportedImagingStudy = FhirStu3.imagingStudy(
    patientEntry, bundle, encounterEntry, study
  );

  assertEquals(exportedImagingStudy.property, someValue);
  ...
}
```

However, because all of the individual export methods of the FHIR exporters are private methods, it seems the only way to test the exporters currently is to create an entire mock `Person` and run them through the public `FhirStu3.convertToFhir()` method.

I think that's out-of-scope for my PR, but a worthwhile endeavor in the near future.

jawalonoski:
Regarding `ImagingStudy`,`Procedure`, etc.

I would prefer if modules have both a `Procedure` and `ImagingStudy` in a chain, even though it is not required.

Why?

Because the [US Core Implementation Guide](http://www.hl7.org/fhir/us/core/) does not profile `ImagingStudy`, but it does profile `Procedure`.

This means that most large vendors in the US will support `Procedure` via the Argonauts program, but not necessarily `ImagingStudy`. If modules produce both, then the fact that a patient had an X-Ray or MRI or Ultrasound will be found by clients no matter the system importing/exposing the data.

jawalonoski:
An alternative to putting `Procedure` and `ImagingStudy` in a sequence within a module would be to have `ImagingStudy` generate a Procedure automatically behind the scenes. I'd also be okay with that approach, but I'm open to discussion.

cjduffett:
It seems error-prone to expect module authors to always remember to add a Procedure state directly preceding ImagingStudy states. I propose adding a `procedure_code` to the `ImagingStudy` state that identifies it's equivalent SNOMED procedure code. I think it would be best for `HealthRecord.imagingStudy()` to add a procedure to the patient's record behind the scenes when an ImagingStudy state is processed.

jawalonoski:
OK, I'm good with that approach.

cjduffett:
@jawalonoski @dehall this is ready for final review. I will update the Wiki after this is merged.

jawalonoski:
This looks good. Thank you for this substantial contribution @cjduffett. Please update the wiki appropriately.

PR #288, opened 2018-04-09
Fixes using Module Builder.

### LINKED ISSUE #188: Opioid Addiction Encounter Issue
The opioid addiction module has the following sequence of states:

1. Enter_Directed_Use_Condition3
2. Enter_Directed_Use_Encounter3 (wellness = true)
3. Directed_Use_Prescription3
4. End_Directed_Use_Encounter

This has the effect of ending the wellness encounter prematurely, since other modules that follow (esp. Wellness) need to add additional events. Maybe this shouldn't be a wellness?

### LINKED ISSUE #232: Default Terminal State Encountered in Dermatitis Module
I think there's a logic error in Living With Dermatitis that sometimes ends up in the default terminal state. I was able to eliminate that state with the following change:

```
        {
          "condition": {
            "condition_type": "And",
            "conditions": [
              {
                "condition_type": "PriorState",
                "name": "Childhood_Atopic_Dermatitis"
              },
              {
                "condition_type": "PriorState",
                "name": "Potentially_Outgrow_AD"
              }
            ]
          },
          "transition": "Potential_AD_Flare_Up"
        },

```

to just:

 ```
       {
          "condition": {
            "condition_type": "PriorState",
            "name": "Childhood_Atopic_Dermatitis"
          },
          "transition": "Potential_AD_Flare_Up"
        },

```

I'm not positive about the correctness of the change.

### LINKED ISSUE #235: Knee Injury Care Plan Reason
`PriorStates [Knee_Injury_CarePlan, ED_Visit_For_Knee_Injury, Knee_Injury, Adult_Incidence_Rates, Wait_For_Injury, Initial, ]`

Here the care plan references a "knee_injury" condition onset that has not yet occurred. I can't figure out an easy fix for this one, since the onset logic cannot just be moved before the care plan.

Maybe just remove the reason for now?

### PR DESCRIPTION
**Pull request by jawalonoski**
Fix #188 using Module Builder. In the opioid module: For the 3rd directed-use entry point, this changes Encounter type to inpatient so the current wellness encounter is not prematurely ended.

Fix #232 using Module Builder. Simplified dermatitis conditional transition logic.

Fix #235 using Module Builder: CarePlans referenced conditions that did not exist. Reordered knee injuries and spinal injuries to be defined before care plans were started.

### CODE REVIEW

File: src/main/resources/modules/dermatitis.json [hunk 1]
```diff
@@ -630,8 +579,8 @@
             "condition_type": "Active Medication",
             "codes": [
               {
-                "system": "RxNorm",
-                "code": "835900",
+                "system": "SNOMED-CT",
```
dehall: This looks like a reversion of #277 ... I guess we didn't merge those module changes back into the module builder. I can almost guarantee this isn't the only time that's going to happen so maybe we want a less manual process to keep them in sync.

jawalonoski: Weird. I didn't even notice. We can probably hook something into a successful build on master to push to the module builder, in the same way we do the graphviz images to the gh-pages.

jawalonoski:
@dehall Ready for review again.

PR #293, opened 2018-04-27
New Command Line Options

jawalonoski:
This pull request add three new command line options.

- `-g` **Gender** This option allows you to generate a population that is all male or all female, using `-g M` or `-g F` respectively.

- `-a` **Age** This option allows you to generate a population in a specific age range using `-a min-max`. For example, to generate people between the ages of 20 and 30, use `-a 20-30`.

## Removed from PR
- `-c` **Codes** This option allows you to generate a population that has a given distribution of codes. The `-c` argument can be used more than once to specify additional codes. For example, to specify a population that is half diabetic, you can use `-c 44054006 0.5`. The code `44054006` is the SNOMED code used for diabetes, and `0.5` is equivalent to 50%.

Warning on the `-c` code option -- this setting will likely cause Synthea to **significantly** over generate patients in order to result in the correct number of patients with the given code. It is possible that Synthea will never complete, if the built-in probabilities in the system make the required combination of criteria unreachable.


dehall:
I will look in more detail as to root cause later but as-is this isn't working as expected for me. The gender option works fine. The age option seems to work, it generates the right number of patients all within the given age range, but then it seems to get caught in a loop somewhere and never terminates. The codes option seems to loop forever as well but it doesn't generate all the expected patients. I'm testing with a very small population size so it shouldn't take very long: `.\run_synthea.bat -p 4 -c 44054006 0.5`

Another more generic thought on command line args: could we make every command line argument also a property in synthea.properties? Then we could even extend it such that every property could also be specified on the command line, ex `run_synthea --exporter.fhir.export false` and give simple aliases for important ones. Plus it shouldn't be too hard to implement. Don't worry about doing that now but something to think about going forward.

jawalonoski:
It has something to do with how I implemented the `meetsCriteria` method and the `continue` statement if they do not.

I'm removing the `code` option and simplifying this PR to only include `-g` for gender  and `-a` for ages....

jawalonoski:
Blah, I left some stupid cruft in there. Maybe that will teach me not to commit anything late in the day.

Issue #297, opened 2018-05-15
CSV File Changes


jawalonoski:
- Add zip codes to the patients file.
- Add start/stop to the encounters.

jawalonoski:
The wiki has also been updated to reflect these changes.

jawalonoski:
Looks like maybe a join error producing the data. We'll look at it, thanks.

PR #300, opened 2018-05-16
Changes for Issue #297

jawalonoski:
This pull request addresses the two issues documented in #297.

This corrects Encounter start and stop times so that they are more realistic (e.g. they are no longer instantaneous, they take extra time depending on the procedures conducted, etc.). Outpatient visits start at 15 minutes, emergency visits start at 60 minutes, and inpatient encounters use at least a full 24 hours.

On CSV export, I'll have to adjust the wiki to reflect changes to the `encounters.csv` file (date is replaced by start and stop iso8601 strings) and the `patients.csv` file (the address is broken up into street, city, state, zip).

This PR includes associated miscellaneous changes: for example, opioid overdoses are now correctly identified as emergency encounters instead of ambulatory.


dehall:
For some reason this seems to work fine every time in eclipse, but on the command line one of the FHIR exporter tests fails.

These are the validation errors from one run on the command line:
```
[error] Encounter @ /*:Encounter[namespace-uri()='http://hl7.org/fhir'][1]/*:period[namespace-uri()='http://hl7.org/fhir'][1]; Test=not(exists(f:start)) or not(exists(f:end)) or (f:start/@value <= f:end/@value); Message=per-1: If present, start SHALL have a lower value than end
    [error] Claim @ /*:Claim[namespace-uri()='http://hl7.org/fhir'][1]/*:billablePeriod[namespace-uri()='http://hl7.org/fhir'][1]; Test=not(exists(f:start)) or not(exists(f:end)) or (f:start/@value <= f:end/@value); Message=per-1: If present, start SHALL have a lower value than end
    [error] Encounter @ /*:Encounter[namespace-uri()='http://hl7.org/fhir'][1]/*:period[namespace-uri()='http://hl7.org/fhir'][1]; Test=not(exists(f:start)) or not(exists(f:end)) or (f:start/@value <= f:end/@value); Message=per-1: If present, start SHALL have a lower value than end
    [error] Claim @ /*:Claim[namespace-uri()='http://hl7.org/fhir'][1]/*:billablePeriod[namespace-uri()='http://hl7.org/fhir'][1]; Test=not(exists(f:start)) or not(exists(f:end)) or (f:start/@value <= f:end/@value); Message=per-1: If present, start SHALL have a lower value than end
    [error] Encounter @ /*:Encounter[namespace-uri()='http://hl7.org/fhir'][1]/*:period[namespace-uri()='http://hl7.org/fhir'][1]; Test=not(exists(f:start)) or not(exists(f:end)) or (f:start/@value <= f:end/@value); Message=per-1: If present, start SHALL have a lower value than end
    [error] Claim @ /*:Claim[namespace-uri()='http://hl7.org/fhir'][1]/*:billablePeriod[namespace-uri()='http://hl7.org/fhir'][1]; Test=not(exists(f:start)) or not(exists(f:end)) or (f:start/@value <= f:end/@value); Message=per-1: If present, start SHALL have a lower value than end expected
```

jawalonoski:
Yeah, I think I have a fix...

dehall:
I was typing up the details below as this new change came in. Not quick enough I guess.
I think the new change is fine to prevent the validation issues but it may be worth fixing the below if possible

---
Upon inspection this exposes a couple issues which should be easy to fix:

1) the encounter in the DeathModule is never ended
2) the Chronic_Sinusitis_Followup encounter in the sinusitis module has a path that doesn't end before a week delay
3) the Encounter state doesn't automatically end the previous encounter if there was one. At a glance, only the Appendicitis module actually expects this, but there may be more

jawalonoski:
I'm going to fix the sinusitis thing. Standby.

PR #307, opened 2018-05-22
Fix #281 - Transaction Bundles.

### LINKED ISSUE #281: Can snythea create a Transcation type of Bundle?
Currently, the Bundle type in Synthea is"Collection" but was wondering if there is a way to generate a Bundle type that is "transaction" in which each resource within the bundle is distinct? 
Thank you

jawalonoski:
Allows users to create FHIR `transaction` Bundles instead of `collection` Bundles.

To enable, modify synthea.properties:

`exporter.fhir.transaction_bundle = false`

Change `false` to `true` and rerun with any FHIR exporter enabled.


dehall:
Code looks fine. I'm trying to understand the request that led to this change -- does this satisfy "each resource within the bundle is distinct" ?

jawalonoski:
The request is making the bundles ready to go to `POST` to a FHIR endpoint out of the box... versus the current format requires a client to make changes to every document.

PR #313, opened 2018-05-25
Metabolic Codes

jawalonoski:
This PR adds additional metabolic syndrome codes.

image

I ran `./run_synthea -s 0 -p 10000` and got these results (filtering codes in the metabolic syndrome set and filtering out dead people):

DESCRIPTION | n | new
-- | -- | --
Body mass index 30+ - obesity (finding) | 3932 | :new:
Prediabetes | 2371
Body mass index 40+ - severely obese (finding) | 946 | :new:
Diabetes | 472
Hypertriglyceridemia (disorder) | 436 | :new:
Metabolic syndrome X (disorder) | 428 | :new:
Hyperglycemia (disorder) | 344 | :new:
Neuropathy due to type 2 diabetes mellitus (disorder) | 166
Diabetic renal disease (disorder) | 143
Diabetic retinopathy associated with type II diabetes mellitus (disorder) | 134
Nonproliferative diabetic retinopathy due to type 2 diabetes mellitus (disorder) | 85
Proliferative diabetic retinopathy due to type II diabetes mellitus (disorder) | 22
Proteinuria due to type 2 diabetes mellitus (disorder) | 7
History of lower limb amputation (situation) | 1

A search of the Global Burden of Disease did not turn up any disability weights associated with the new metabolic codes. I did find one for hypothyroidism though, so I preemptively added it.

### DIFF
--- a/src/main/resources/gbd_disability_weights.json
+++ b/src/main/resources/gbd_disability_weights.json
@@ -538,5 +538,11 @@
     "system": "SNOMED-CT",
     "code": "197927001",
     "disability_weight": 0.051
-    }
+    },
+
+  "Hypothyroidism" : {
+    "system": "SNOMED-CT",
+    "code": "40930008",
+    "disability_weight": 0.019
+  }
 }
--- a/src/main/resources/modules/metabolic_syndrome_care.json
+++ b/src/main/resources/modules/metabolic_syndrome_care.json
@@ -75,7 +75,7 @@
           "transition": "Record_HA1C"
         },
         {
-          "transition": "Wellness_Encounter"
+          "transition": "Obese_Check"
         }
       ]
     },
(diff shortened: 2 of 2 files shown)


src/main/resources/modules/metabolic_syndrome_care.json
```diff
@@ -43,1883 +42,2236 @@
...
-            "conditions" : [
-              {
-                "condition_type" : "Attribute",
-                "attribute" : "blindness",
-                "operator" : "is not nil"
+    "Diagnose_Retinopathy": {
+      "type": "ConditionOnset",
+      "codes": [
+        {
+          "system": "SNOMED-CT"
```
dehall: These might be better off as "Active Condition"s but it's not a big deal.

src/main/resources/gbd_disability_weights.json
```diff
@@ -538,5 +538,11 @@
     "system": "SNOMED-CT",
     "code": "197927001",
     "disability_weight": 0.051
-    }
+    },
+
+  "Hypothyroidism" : {
```
dehall: Seeing this here makes me think these disability weights (and other things like costs) should not be split up in this separate file, but defined in the module itself. Right now the module is just the states, but we could easily extend it so that a module could list out the codes it's going to use, and then define costs or disability weights as appropriate. That way the module would be self-contained.

arscan:
Deployed to module builder:

https://synthetichealth.github.io/module-builder/#metabolic_syndrome_care

PR #316, opened 2018-05-30
Reformat modules through module builder to standardize formatting.

arscan:
In module builder, you can now do:

```
npm run reformat-modules ../synthea/src/main/resources/modules
```

That will reformat the synthea modules in-place.  Right now it is just simply `string -> JSON -> string indented 2 spaces`.  But we may want to do more in the future.

The changes appear correct because when I repackage them into the module builder bundle, no changes show up (meaning only formatting was altered).  But you may want to run a batch to double check.

It seems like it would make sense to have a script within synthea that auto-formats modules, if people are writing them on their own, and have a test that fails if they aren't formatted properly (and ask them to run the format script).  But that would be a tradeoff between code duplication vs. dependency on installing module builder locally (or loading/exporting them in the module builder).


src/main/resources/modules/female_reproduction.json
```diff
@@ -147,7 +137,6 @@
         "indicate the number out of 100 women who experience planned or unplanned ",
         "pregnancy within the first year of use. These percentages are scaled for ",
         "a monthly time step: ",
-
```
arscan: I'm conflicted on formatting within remarks.  Should we retain these paragraph markers (perhaps with a line with `"",`)?  We'll just lose this information if we merge this in now.

dehall: I'm ok with losing these blank lines. As long as it doesn't reformat the content within quotes, ex keeps the table formatting shown here, it's fine.

src/main/resources/modules/metabolic_syndrome_disease.json
```diff
@@ -1,799 +1,1033 @@
...
+      ],
+      "complex_transition": [
+        {
+          "condition": {
+            "condition_type": "Race",
+            "race": "White"
           },
-          "distributions" : [
-            { "distribution" : 0.083, "transition" : "Eventual_Diabetes", "remarks" : ["0.065 * 1.278"] },
-            { "distribution" : 0.45, "transition" : "Eventual_Prediabetes", "remarks" : ["0.38 * 1.2"] },
```
dehall: A note just for future consideration: are there ways to keep this kind of thing on one line? I feel like the one-liner is a lot easier to follow than the exploded version here.

arscan: There are ways of doing this, but it would be a little messy.  I can think of a creative way of using `toJSON` within objects using `JSON.stringify` to replace the distribution contents with a keyword, which we then replace with our custom format after it has been turned into a string.  Maybe something worth adding a low priority issue for.

PR #322, opened 2018-06-01
Hypothyroidism

jawalonoski:
This pull request adds hypothyroidism separately from the veteran modules.

It also fixes a few checkstyle issues and removes an unused deprecated property.

After this PR is merged, the last remaining change for the CHW analysis is the price adjustments.

### DIFF
--- a/src/main/resources/modules/primary_atrophic_hypothyroidism.json
+++ b/src/main/resources/modules/primary_atrophic_hypothyroidism.json
@@ -104,8 +104,8 @@
       ],
       "direct_transition": "Label Panel Results",
       "range": {
-        "low": 5,
-        "high": 2
+        "low": 2,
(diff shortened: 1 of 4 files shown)


src/main/resources/modules/primary_atrophic_hypothyroidism.json
```diff
@@ -0,0 +1,201 @@
...
+      "codes": [
+        {
+          "system": "LOINC",
+          "code": "3016-3",
+          "display": "Thyrotropin [Units/​volume] in Serum or Plasma"
+        }
+      ],
+      "direct_transition": "Label Panel Results",
+      "range": {
+        "low": 5,
```
dehall: Low & high here appear to be flipped

PR #324, opened 2018-06-01
Cost variation

dehall:
Updates the cost logic in the following ways:
1) Cost data now includes a minimum, mode, and maximum, so that individual claims will be priced using a triangular distribution:
2) Costs are adjusted based on a regional (state-based) factor. As a first pass, this factor is based on the 2011 CMS urban and rural GAFs, aggregated and weighted based on the urban/rural population breakdown of that state. Will add that info to the wiki page once this is merged. A future implementation will likely want a more robust adjustment factor, but that would not be a significant change.
3) Simplified the data model by removing the ClaimItem class and adding costs directly to the Entry.
4) In anticipation of future changes, the cost function now takes in the Entry, the Person, the Provider, and the Payer. Currently the provider and payer are not used.
5) Adds "dispenses" (aka refills + the original prescription) and "total cost" (== cost * dispenses) to the CSV export. This exporter uses the medication prescription details, if available, to figure out the number of refills per medication. If no prescription details are available it assumes a refill per month. Sample:   
[medications.csv.txt](https://github.com/synthetichealth/synthea/files/2077725/medications.csv.txt)


jawalonoski:
Eh. I actually have mixed feelings about eliminating the ClaimItems and adding the costs directly to the Entry. For one thing, not all Entries have a cost. I'm also not sure about what information we may need in the future. Oh well, that should be fine for now.

src/main/java/org/mitre/synthea/export/CSVExporter.java
```diff
@@ -487,7 +487,7 @@ private void medication(String personID, String encounterID,
     s.append(coding.code).append(',');
     s.append(clean(coding.display)).append(',');
 
-    s.append(String.format("%.2f", Costs.calculateCost(medication, true))).append(',');
+    s.append(String.format("%.2f", medication.cost())).append(',');
```
jawalonoski: Can you add the number of refills and total cost like we talked about?

dehall: I can, though the numbers will be completely invented, and it will likely use a one-size fits all function that may be completely inappropriate in some cases. 

The MedicationOrder state has an optional `prescription` object that includes # of refills, duration, etc. https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-States#prescription But only a few medications have prescription details, and nothing has `refills`. We could calculate something using a `duration` provided, if there is one. But that's completely independent of when medications start and stop so could lead to unusual results. Or I could pick a # of refills assuming 1 per month?

dehall: Updated with "dispenses" because I wanted it to be clear that the number does include the original prescription. Using the term "refills" makes it less clear as to whether the # represents N or N-1.

src/main/java/org/mitre/synthea/export/CSVExporter.java
```diff
@@ -487,7 +494,45 @@ private void medication(String personID, String encounterID,
...
+      dispenses = medDuration / durationMs;
+    } else {
+      // assume 1 refill / month
+      long durationMs = Utilities.convertTime("months", 1);
+      dispenses = medDuration / durationMs;
+    }
+    
+    if (dispenses < 1) {
+      // integer division could leave us with 0, 
+      // if the active time is less than the
```
jawalonoski: unfinished sentence...

src/main/java/org/mitre/synthea/world/concepts/Costs.java
```diff
@@ -84,22 +121,77 @@ public static boolean hasCost(Entry entry) {
...
+      state = Location.getAbbreviation(state);
+      if (LOCATION_ADJUSTMENT_FACTORS.containsKey(state)) {
+        locationAdjustment = (double) LOCATION_ADJUSTMENT_FACTORS.get(state);
+      }
+    }
+    
+    return baseCost * locationAdjustment;
+  }
+  
+  private static class CostData {
```
jawalonoski: Add Javadocs.

jawalonoski:
Just fix that unfinished comment and I'll merge it.

PR #325, opened 2018-06-01
Custom reports from arbitrary SQL

dehall:
Adds the ability to automatically run custom reports after a population is generated, using arbitrary SQL. Queries in the file src/main/resources/custom_queries.sql (configurable) will be run one at a time, and exported to CSV in output/reports. Lines starting with -- will be ignored, so this file includes a few sample queries to get started with.

This is not a "finished product" as it has some limitations but it is sufficiently usable that someone with an understanding of the DataStore data model could produce useful reports. After this is merged I plan to add a wiki page with an overview of the data model.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/CustomSqlReport.java
+++ b/src/main/java/org/mitre/synthea/export/CustomSqlReport.java
@@ -1,6 +1,7 @@
 package org.mitre.synthea.export;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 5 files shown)


jawalonoski:
This is a neat pull request. My issue with it is that we've reached the point where we have multiple types of reports -- custom, prevalence, and cost_access_outcomes. There will be more to come. Another qualm is the database itself... it doesn't scale very well, even though it is awesome. We need to come up with a strategy to consolidate and simplify reporting and querying.

PR #327, opened 2018-06-07
WIP: fix city/town name xref issues

dehall:
work in progress to cleanup demographics/zipcode lookup mismatches.
- [x] handle state-specific labels in code (Town, borough, village)
- [ ] remove state-specific labels from data files?
- [ ] manually curate remaining mismatches

list of all mismatches:
[missing_zips.csv.txt](https://github.com/synthetichealth/synthea/files/2078619/missing_zips.csv.txt)

(found using sql query: `select name, stname from demographics d where not EXISTS (select 1 from zipcodes z where d.NAME = Z.NAME AND D.STNAME = Z.USPS)`  )

### DIFF
--- a/src/main/java/org/mitre/synthea/world/geography/Location.java
+++ b/src/main/java/org/mitre/synthea/world/geography/Location.java
@@ -3,6 +3,7 @@ package org.mitre.synthea.world.geography;
 import com.google.common.collect.Table;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jawalonoski:
I have a few issues where this is headed.

First, not all named places in the US have a postal code. So, there will be some places with missing zips. Right now, they get `00000`. This behavior is fine for now, and I don't think we should throw a runtime exception. The zip code isn't that important.

Second, I do not want to hard-code specific suffixes (like "Town" or "village" or whatever) for specific US-states. This adds more "US specific" ties in the code itself.

dehall:
The issue isn't really about zip codes, it's the geographic coordinates which are included in zipcodes.csv. 00000 would be fine but we need the lat/lons to do provider lookups. I'll rework this to just change the data files

PR #335, opened 2018-06-12
Urgent care

cmbear37:


src/main/java/org/mitre/synthea/modules/EncounterModule.java
```diff
@@ -44,19 +57,60 @@ public boolean process(Person person, long time) {
...
+        if (person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL) == null){
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, 0);
+        }
+        if (person.symptomTotal() != (int)person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL)) {
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, person.symptomTotal());
+          person.addressLargestSymptom();
+          Encounter encounter = person.record.encounterStart(time, EncounterType.EMERGENCY.toString());
+          encounter.name = "Encounter Module Symptom Driven";
+          encounter.provider = person.getEmergencyProvider(time);
+          encounter.reason = ENCOUNTER_EMERGENCY;
```
jawalonoski: The reason for the encounter isn't that it is an emergency encounter.

If you can't turn the symptom into a code (hint: doesn't look like you can), then you should delete this line.

src/main/java/org/mitre/synthea/modules/EncounterModule.java
```diff
@@ -44,19 +57,60 @@ public boolean process(Person person, long time) {
...
+        if (person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL) == null){
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, 0);
+        }
+        if (person.symptomTotal() != (int)person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL)) {
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, person.symptomTotal());
+          person.addressLargestSymptom();
+          Encounter encounter = person.record.encounterStart(time, EncounterType.URGENTCARE.toString());
+          encounter.name = "Encounter Module Symptom Driven";
+          encounter.provider = person.getUrgentCareProvider(time);
+          encounter.reason = ENCOUNTER_URGENTCARE;
```
jawalonoski: Same here. The reason is for an urgent care encounter isn't that the patient has an urgent care encounter. Delete this line.

src/main/java/org/mitre/synthea/modules/EncounterModule.java
```diff
@@ -44,19 +57,60 @@ public boolean process(Person person, long time) {
...
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, 0);
+        }
+        
+        if (person.symptomTotal() != (int)person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL)) {
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, person.symptomTotal());
+          person.addressLargestSymptom();
+          Encounter encounter = person.record.encounterStart(time, EncounterType.WELLNESS.toString());
+          encounter.name = "Encounter Module Symptom Driven";
+          encounter.provider = person.getUrgentCareProvider(time);
+          encounter.reason = ENCOUNTER_CHECKUP;
```
jawalonoski: Again. Delete this line. Reason can't be the encounter itself.

src/main/java/org/mitre/synthea/modules/EncounterModule.java
```diff
@@ -44,19 +57,60 @@ public boolean process(Person person, long time) {
...
+        if (person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL) == null){
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, 0);
+        }
+        
+        if (person.symptomTotal() != (int)person.attributes.get(LAST_VISIT_SYMPTOM_TOTAL)) {
+          person.attributes.put(LAST_VISIT_SYMPTOM_TOTAL, person.symptomTotal());
+          person.addressLargestSymptom();
+          Encounter encounter = person.record.encounterStart(time, EncounterType.WELLNESS.toString());
+          encounter.name = "Encounter Module Symptom Driven";
+          encounter.provider = person.getUrgentCareProvider(time);
```
jawalonoski: Urgent Care Provider? Why not just Ambulatory / Wellness provider?

src/main/java/org/mitre/synthea/modules/EncounterModule.java
```diff
@@ -118,6 +172,18 @@ public static void emergencyEncounter(Person person, long time) {
...
 
+  public static void urgentCareEncounter(Person person, long time) {
+    // find closest service provider with urgent care service
+    Provider provider = person.getUrgentCareProvider(time);
+    provider.incrementEncounters("urgent_care", Utilities.getYear(time));
+
+    Encounter encounter = person.record.encounterStart(time, "urgent_care");
+    encounter.codes.add(ENCOUNTER_URGENTCARE);
+    // assume people will be in urgent care for one hour
+    person.record.encounterEnd(time + TimeUnit.HOURS.toMillis(1), "emergency");
```
jawalonoski: `emergency` should be `urgent_care`

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1273,13 +1272,15 @@ public boolean process(Person person, long time) {
     private String cause;
     private Range<Integer> range;
     private Exact<Integer> exact;
+    public boolean addressed;
```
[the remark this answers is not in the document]
dehall: As I think about it this makes sense. Maybe a module would want to branch on medication/adherence and set a symptom to a low value but not 0, and mark it as addressed. I'm ok with this.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1273,13 +1272,15 @@ public boolean process(Person person, long time) {
...
 
     @Override
     protected void initialize(Module module, String name, JsonObject definition) {
       super.initialize(module, name, definition);
       if (cause == null) {
         cause = module.name;
       }
+      addressed = false;
```
dehall: This isn't necessary because the default for booleans is false.

src/main/java/org/mitre/synthea/modules/EncounterModule.java
```diff
@@ -44,19 +58,54 @@ public boolean process(Person person, long time) {
...
-      // add a symptom driven encounter if symptoms are severe
-      person.resetSymptoms();
-      Encounter encounter = person.record.encounterStart(time, EncounterType.WELLNESS.toString());
-      encounter.name = "Encounter Module Symptom Driven";
-      encounter.codes.add(ENCOUNTER_CHECKUP);
-      encounter.codes.add(getWellnessVisitCode(person, time));
-      person.attributes.put(ACTIVE_WELLNESS_ENCOUNTER, true);
-      startedEncounter = true;
-    }
+    } else if (person.symptomTotal() > EMERGENCY_SYMPTOM_THRESHOLD) {
```
dehall: Lot of duplication here and it feels like there may be opportunity to extract out common logic, but this isn't a big deal.

src/main/java/org/mitre/synthea/world/agents/Person.java
```diff
@@ -83,6 +84,7 @@ public Person(long seed) {
     attributes = new ConcurrentHashMap<String, Object>();
     vitalSigns = new ConcurrentHashMap<VitalSign, Double>();
     symptoms = new ConcurrentHashMap<String, Map<String, Integer>>();
+    symptomStatuses = new ConcurrentHashMap<String, Map<String, Boolean>>();
```
dehall: I think it would be ideal to have a real Symptom class that was associated to the Conditions, but I'm ok with this as a first step.

PR #346, opened 2018-06-20
added functional status assessment states to total joint replacement …

laambrosino:
…module and created a submodule for the assessments

### DIFF
--- a/src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
+++ b/src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
@@ -13,31 +13,6 @@
           "system": "LOINC",
           "code": "72091-2",
           "display": "Knee injury and Osteoarthritis Outcome Score [KOOS]"
-        },
-        {
-          "system": "LOINC",
-          "code": "72101-9",
-          "display": "Symptoms score [KOOS]"
-        },
-        {
-          "system": "LOINC",
-          "code": "72102-7",
-          "display": "Pain score [KOOS]"
-        },
-        {
-          "system": "LOINC",
-          "code": "72100-1",
-          "display": "Activities of daily living score [KOOS]"
-        },
-        {
-          "system": "LOINC",
-          "code": "72099-5",
-          "display": "Sport-recreation score [KOOS]"
-        },
-        {
-          "system": "LOINC",
-          "code": "72098-7",
-          "display": "Quality of life score [KOOS]"
         }
       ],
       "direct_transition": "KOOS_Attribute_Set",
@@ -56,16 +31,6 @@
           "system": "LOINC",
           "code": "61576-5",
           "display": "PROMIS short form - global - version 1.1"
-        },
-        {
-          "system": "LOINC",
-          "code": "71972-4",
-          "display": "PROMIS-10 Global Physical Health (GPH) score"
-        },
-        {
-          "system": "LOINC",
-          "code": "71970-8",
-          "display": "PROMIS-10 Global Mental Health (GMH) score"
         }
       ],
       "direct_transition": "PROMIS10_Attribute_Set",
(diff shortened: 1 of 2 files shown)


src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
```diff
@@ -0,0 +1,1372 @@
...
+    "This is a submodule that includes the possible assessments a patient will go through prior to and following either a total knee replacement or a total hip replacement.\n"
+  ],
+  "states": {
+    "KOOS_Total_Assessment": {
+      "type": "MultiObservation",
+      "category": "survey",
+      "number_of_observations": 5,
+      "codes": [
+        {
+          "system": "LOINC",
```
dehall: The list of codes here should only contain the code(s) for the group as a whole, not the individual components (since those codes are listed on the individual observations themselves). In this case please just list the one for the top-level KOOS, not the individual scores. Same thing for the PROMIS, VR12, etc.

laambrosino: Got it. I didn't realize it would group them without putting those in, but I understand it now so I'll fix that.

src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
```diff
@@ -0,0 +1,1372 @@
...
+    },
+    "HOOS_Attribute_Set": {
+      "type": "SetAttribute",
+      "attribute": "assessment_done",
+      "direct_transition": "FSA_Performed",
+      "value": "HOOS"
+    },
+    "FSA_Attribute_Knee": {
+      "type": "SetAttribute",
+      "attribute": "functional_status_assessments",
```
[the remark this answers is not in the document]
laambrosino: Got it. I'll fix that. I think I started to do it because I had an idea for it but then it wasn't necessary.

src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
```diff
@@ -0,0 +1,1141 @@
...
+      ],
+      "direct_transition": "HOOS_Attribute_Set",
+      "remarks": [
+        "Five different subscales are scored separately: HOOS Pain (9 items); HOOS Symptoms (7 items); HOOS ADL (17 items); HOOS Sport/Rec (5 items); HOOS QOL (4 items). The scores are transformed to a 0–100 scale, with zero representing extreme knee problems and 100 representing no knee problems.",
+        "http://www.koos.nu"
+      ]
+    },
+    "Joint_Replacement_Assessment_Terminal": {
+      "type": "Terminal",
+      "conditional_transition": [
```
dehall: Please remove the transitions from the Terminal state -- terminals can't have transitions. (I'm actually surprised the UI allowed it, and it doesn't show on the graphical view so it's easy to miss)

src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
```diff
@@ -0,0 +1,1141 @@
...
+      "type": "SetAttribute",
+      "attribute": "assessment_done",
+      "direct_transition": "FSA_Performed",
+      "value": "HOOS"
+    },
+    "FSA_Performed": {
+      "type": "Simple",
+      "direct_transition": "Joint_Replacement_Assessment_Terminal"
+    },
+    "PROMIS10_GPH": {
```
dehall: Is there any difference between the knee and hip version of the PROMIS10 and VR12 mental health tracks? If not it seems like you could save a lot of duplication by having both hip and knee transition to the same state.

src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
laambrosino: Didn't even realize it was there - got it.

src/main/resources/modules/total_joint_replacement/functional_status_assessments.json
laambrosino: They are the same. Thanks, I'll fix it.

PR #357, opened 2018-06-28
Cystic fibrosis

KeeyanGhoreshi:
The finished cystic fibrosis module.

### DIFF
--- a/src/main/resources/modules/cystic_fibrosis.json
+++ b/src/main/resources/modules/cystic_fibrosis.json
@@ -1,0 +1,1665 @@
+{
+  "name": "Cystic Fibrosis",
+  "remarks": [
+    "Cystic Fibrosis, or mucoviscidosis, is an autosomal recessive genetic disorder which is named after that scarring (fibrosis) and cyst formation in the pancreas.  It is characterized by thick, viscous secretions such as thickening of mucus which causes many respiratory problems.  Statistics and information were gathered from the pages and information available at:\nhttps://www.cff.org\n\nSome data and statistic came from:\nhttps://www.sciencedirect.com/science/article/pii/S1569199315002106?viewFullText=true#f0015\n\n\n\n\n\n"
+  ],
+  "states": {
+    "Initial": {
+      "type": "Initial",
+      "direct_transition": "Potential_Onset"
+    },
+    "Terminal": {
+      "type": "Terminal"
+    },
+    "Potential_Onset": {
+      "type": "Simple",
+      "complex_transition": [
+        {
+          "condition": {
+            "condition_type": "Race",
+            "race": "white"
+          },
+          "distributions": [
+            {
+              "transition": "Cystic_Fibrosis",
+              "distribution": 0.0003333333
+            },
+            {
+              "transition": "Terminal",
+              "distribution": 0.9996666667
+            }
+          ]
+        },
+        {
+          "condition": {
+            "condition_type": "Race",
+            "race": "black"
+          },
+          "distributions": [
+            {
+              "transition": "Cystic_Fibrosis",
+              "distribution": 0.000066225
+            },
+            {
+              "transition": "Terminal",
+              "distribution": 0.999933775
+            }
+          ]
+        },
+        {
+          "distributions": [
+            {
+              "transition": "Cystic_Fibrosis",
+              "distribution": 0.000028571
+            },
+            {
+              "transition": "Terminal",
+              "distribution": 0.999971429
+            }
+          ],
+          "condition": {
+            "condition_type": "Race",
+            "race": "asian"
+          }
+        }
+      ],
+      "remarks": [
+        "https://ghr.nlm.nih.gov/condition/cystic-fibrosis#statistics",
+        "",
+        "The frequency dropdown shows statistics about fibrosis prevalence by race."
+      ]
+    },
+    "Cystic_Fibrosis": {
+      "type": "ConditionOnset",
+      "target_encounter": "Fibrosis_Screening",
+      "codes": [
+        {
+          "system": "SNOMED-CT",
+          "code": 190905008,
+          "display": "Cystic Fibrosis"
+        }
+      ],
+      "complex_transition": [
+        {
+          "condition": {
+            "condition_type": "Gender",
+            "gender": "M"
+          },
+          "distributions": [
+            {
+              "transition": "Male_Infertility",
+              "distribution": 0.98
+            }
+          ]
+        },
+        {
+          "condition": {
+            "condition_type": "Gender",
+            "gender": "F"
+          },
+          "distributions": [
+            {
+              "transition": "Female_Infertility",
+              "distribution": 0.5
+            }
+          ]
+        },
+        {
+          "distributions": [],
+          "transition": "chronic_cough"
+        }
+      ]
+    },
+    "Age <1": {
+      "type": "Delay",
+      "exact": {
+        "quantity": 1,
+        "unit": "weeks"
+      },
+      "distributed_transition": [
+        {
+          "transition": "symptoms_dont_improve",
+          "distribution": 0.82
+        },
+        {
+          "transition": "Meconium Ileus",
+          "distribution": 0.18
+        }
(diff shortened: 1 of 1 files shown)


src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1538 @@
+{
+  "name": "Cystic Fibrosis",
+  "remarks": [
+    "Cystic Fibrosis, or mucoviscidosis, is an autosomal recessive genetic disorder which is named after that scarring (fibrosis) and cyst formation in the pancreas.  It is characterized by thick, viscous secretions such as thickening of mucus which causes many respiratory problems.  Statistics and information were gathered from the pages and information available at:\nhttps://www.cff.org\n\nSome data and statistic came from:\nhttps://www.sciencedirect.com/science/article/pii/S1569199315002106?viewFullText=true#f0015\n\n\n\n\n\n"
```
dehall: Great to have remarks with sources at the top level, but it would be even more useful if the individual states where there are distributed transitions had specific links to what the cited info is and where it comes from

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1538 @@
...
+        }
+      ]
+    },
+    "Lung_Transplant_Denied": {
+      "type": "SetAttribute",
+      "attribute": "lung_transplant",
+      "direct_transition": "Life_With_Cystic_Fibrosis",
+      "value": "denied"
+    },
+    "Lung_Transplant": {
```
dehall: This Procedure state has to occur within an Encounter

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1538 @@
...
+      "allow": {
+        "condition_type": "Not",
+        "condition": {
+          "condition_type": "PriorState",
+          "name": "Develop_Diabetes"
+        }
+      },
+      "direct_transition": "Develop_Diabetes"
+    },
+    "no_double_infection": {
```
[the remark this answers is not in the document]
KeeyanGhoreshi: Would it be sufficient to have an intermediate state that transitions to Develop_Infection if some condition is met and away if not?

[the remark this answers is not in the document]
dehall: Yes that should be fine - just don't want to get stuck in the Guard forever

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1649 @@
...
+        "https://www.cff.org/About-Us/Media-Center/Press-Releases/The-Cystic-Fibrosis-Foundation-Applauds-FDA-Approval-of-Kalydeco/"
+      ]
+    },
+    "Prescribe_Mucus_Thinner": {
+      "type": "MedicationOrder",
+      "codes": [
+        {
+          "system": "RxNorm",
+          "code": 74704,
+          "display": "Mucus Thinner"
```
jawalonoski: Check this display name... "Mucus Thinner"

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1649 @@
...
+            {
+              "transition": "startDiseaseChain",
+              "distribution": 0.931
+            },
+            {
+              "transition": "patient_lives",
+              "distribution": 0.02
+            }
+          ],
+          "condition": {
```
jawalonoski: This condition will never be reached.

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1649 @@
...
+      "codes": [
+        {
+          "system": "SNOMED-CT",
+          "code": 448813005,
+          "display": "Sepsis caused by Pseudomonas (disorder)"
+        }
+      ],
+      "direct_transition": "low_blood_pressure"
+    },
+    "Emergency_Room": {
```
jawalonoski: Every quarterly visit is an emergency room visit?

jawalonoski: Or at least, the two are mutually exclusive?

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1649 @@
...
+      "cause": "",
+      "exact": {
+        "quantity": 1
+      },
+      "direct_transition": "Getting_Diagnosed"
+    },
+    "Quarterly_Checkup": {
+      "type": "Encounter",
+      "encounter_class": "ambulatory",
+      "codes": [],
```
jawalonoski: Encounter code is required.

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1649 @@
...
+      "reason": "Cystic_Fibrosis",
+      "remarks": [
+        "https://www.cff.org/Life-With-CF/Daily-Life/Fitness-and-Nutrition/Nutrition/Taking-Care-of-Your-Digestive-System/Enzymes/",
+        "",
+        "\"Between 85 and 90% of CF patients have pancreatic insufficiency."
+      ]
+    },
+    "Life_With_Cystic_Fibrosis": {
+      "type": "Delay",
+      "exact": {
```
jawalonoski: You might want to change this delay from 13 weeks fixed, to be a range of like 12-14 weeks, just because scheduling encounters is not always exact.

src/main/resources/modules/cystic_fibrosis.json
```diff
@@ -0,0 +1,1649 @@
...
+            "condition_type": "PriorState",
+            "name": "Develop_Infection_2"
+          }
+        },
+        {
+          "transition": "Develop_Infection_2"
+        }
+      ]
+    },
+    "Lung_Transplant_Wait": {
```
jawalonoski: Right now this lung transplant encounter lasts 4-5 hours. There probably should be an inpatient hospital stay here that lasts at least several days.

src/main/resources/modules/cystic_fibrosis.json
KeeyanGhoreshi: I'm not sure what you mean here, the Quarterly Visit is ambulatory.

jawalonoski: Does a patient have an emergency room visit XOR a quarterly visit? Can they have both in the same quarter?

KeeyanGhoreshi: They can't have both in the same quarter, they have to either get their checkup or go to the emergency room.

jawalonoski: OK, can you change that? If you have a standing quarterly visit, you're going to go to that, even if you happen to randomly and periodically end up the emergency room.

KeeyanGhoreshi: Can do

dehall: In the latest version if the patient goes to the emergency room and doesn't die they will have their quarterly checkup immediately after. This may be a good candidate to merge into 2 modules as there are parallel tracks here. At minimum there should be some delay between the end of the emergency room encounter and the start of the quarterly checkup.

KeeyanGhoreshi: I added a delay state for now to fix this.  I don't know what you mean by "merge into 2 modules" though.

dehall: Bad phrasing on my part. I meant split into 2 modules. It's possible for one module to represent multiple parallel tracks, but the logic can get complicated. Splitting it apart into multiple modules could make each one simpler, at the cost of having more modules to keep track of.

Yet another option would be to reduce the time delay in the loop so that it runs more often, for example every month, checking if the quarterly checkup hasn't happened in some time using a `PriorState` logic, then falling back to some percent chance of ER visit, or finally just looping back to the delay.

I'm ok with the current impl but I'll defer to @jawalonoski

PR #358, opened 2018-06-29
Clinician

cmbear37:
- Added Primary Care Offices for patients to visit (and assigned when appropriate)
- Generates the correct amounts of clinicians per provider, or default is one general practice clinician
- Assigns general practice clinician of provider to each encounter
- Displays clinician in Text Export


jawalonoski:
Please check the `checkstyle` output... You can ignore the `Missing a Javadoc comment.` warnings, but others should be dealt with:

```
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/export/Exporter.java:221: Line is longer than 100 characters (found 102). [LineLength]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/agents/Clinician.java:9: Wrong lexicographical order for 'com.google.common.collect.HashBasedTable' import. Should be before 'java.util.concurrent.atomic.AtomicInteger'. [CustomImportOrder]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/agents/Clinician.java:10: Wrong lexicographical order for 'com.google.common.collect.Table' import. Should be before 'java.util.concurrent.atomic.AtomicInteger'. [CustomImportOrder]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/agents/Clinician.java:11: Wrong lexicographical order for 'com.google.gson.internal.LinkedTreeMap' import. Should be before 'java.util.concurrent.atomic.AtomicInteger'. [CustomImportOrder]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/agents/Provider.java:284:7: Line contains a tab character. [FileTabCharacter]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/agents/Provider.java:288:7: Line contains a tab character. [FileTabCharacter]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java:100: Line is longer than 100 characters (found 107). [LineLength]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/geography/Location.java:242:6: Unused @param tag for 'person'. [JavadocMethod]
[ant:checkstyle] [WARN] /home/travis/build/synthetichealth/synthea/src/main/java/org/mitre/synthea/world/geography/Location.java:247: Overload methods should not be split. Previous overloaded method located at line '168'. [OverloadMethodsDeclarationOrder]
```

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -35,7 +41,7 @@
 
   // ArrayList of all providers imported
   private static ArrayList<Provider> providerList = new ArrayList<Provider>();
-  private static QuadTree providerMap = new QuadTree(500, 500); // node capacity, depth
+  private static QuadTree providerMap = new QuadTree(1400, 1400); // node capacity, depth
```
dehall: I know the initial number and this new number are likely both pulled from a hat, but could you test out how much this impacts memory usage when full? We don't need a ton of detail but a general sense would be good. We can talk about how to do that.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -49,15 +55,26 @@
   public String type;
   public String ownership;
   public int quality;
+  public static String numClinicians;
+  public ArrayList<Clinician> clinicians;
   private DirectPosition2D coordinates;
-  private ArrayList<String> servicesProvided;
+  public ArrayList<String> servicesProvided;
+  private Map<String,Integer> specialtyInfo;
```
dehall: Are the specialtyInfo, clinicians, and numClinicians fields used anywhere? If they are just written to and never used they should be removed.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -235,13 +272,63 @@ public static void loadProviders(String state, String abbreviation, String filen
...
+        //only enter this if THERE IS AN INFO FILE
+        //if there is no info file or if the provider is not found in the info file
+        // make them one general clinician that's always chosen
+        if (csvSpecialties == null || !found) {
+          parsed.specialtyInfo.put("GENERAL PRACTICE", 1);
+          parsed.clinicianMap.put("GENERAL PRACTICE", 
+              generateClinicianList(parsed, 1, "GENERAL PRACTICE"));
+        } else {
+          for (Map<String,String> rowSpecialty : csvSpecialties) {
+            if (rowSpecialty.get("address").equals(address) && rowSpecialty.get("city").equals(city)
```
dehall: This is essentially going to do a "join" by address-city-state by looping over the entire file. It will work but it's slow. This is where combining the two files into one would help a lot.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -235,13 +272,63 @@ public static void loadProviders(String state, String abbreviation, String filen
...
+                    && rowSpecialty.get("state").equals(abbreviation)) {
+              for (Map.Entry<String, String> e : rowSpecialty.entrySet()) {
+                // if the field is not any of the provider info
+                if (!e.getKey().equals("address") && !e.getKey().equals("city")
```
dehall: Someone could want to add a new unrelated column and not realize that they have to add new checks here. Not a hard and fast rule, but it's often safer to look for things you do want instead of excluding specific things you don't want. If you extract the list of specialties to a new class, you could add a getSpecialtiesList() or something and then loop over that here to pick the columns you want.

src/main/java/org/mitre/synthea/world/agents/Clinician.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.sis.index.tree.QuadTreeData;
+import org.mitre.synthea.engine.EventList;
+import org.mitre.synthea.world.concepts.HealthRecord;
+import org.mitre.synthea.world.concepts.VitalSign;
+
+
+
+
+
+public class Clinician implements Serializable, QuadTreeData {
```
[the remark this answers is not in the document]
jawalonoski: Because java doesn't support multiple inheritance, we need to decide whether it makes more sense for `Clinician` to `extend Person` or `extend Provider` or neither.

Extending `Person` provides all sorts of demographics and the ability to generate medical records for the clinicians themselves. While extending `Clinician` allows us to track how many procedures a doctor performed, for example. A third option would be to simply contain `Person` and `Provider` objects and use them appropriately. We need to keep in mind that Clinicians will be able to work at multiple `Provider` organizations. We may want to separate the `Interface` from the classes, so we can reuse them.

src/main/java/org/mitre/synthea/world/agents/Provider.java
jawalonoski: The original number was quite pulled from a hat, although it looks nicely rounded. It started out smaller until it ran out of space loading the providers. I slowly increased the size until it safely was able to load the biggest states: e.g. California, Texas.

.checkstyle
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dehall: Please remove this file - it's an Eclipse-specific configuration file so it shouldn't be in the repo. I'd recommend not using `git add *` or `git add .` because you can wind up adding things you don't really want.

src/main/java/org/mitre/synthea/helpers/SimpleCSV.java
```diff
@@ -37,9 +38,24 @@
 
     MappingIterator<LinkedHashMap<String, String>> it = mapper.readerFor(LinkedHashMap.class)
         .with(schema).readValues(csvData);
+    
 
     return it.readAll();
   }
+  
+  public static Iterator<LinkedHashMap<String, String>> parseLineByLine(String csvData) throws IOException {
```
dehall: Please add a Javadoc comment for this new method. It should mostly be the same as `parse` method above but distinguish why someone would want to use this one instead. (ie, reduced memory usage).

Also please review the formatting on this method, it doesn't look like Checkstyle will approve of this

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -50,14 +60,20 @@
   public String ownership;
   public int quality;
   private DirectPosition2D coordinates;
-  private ArrayList<String> servicesProvided;
+  public ArrayList<String> servicesProvided;
+  public Map<String, ArrayList<Clinician>> clinicianMap;
   // row: year, column: type, value: count
   private Table<Integer, String, AtomicInteger> utilization;
+  public long seed;
```
dehall: Please remove this if it's not used anywhere.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -355,7 +356,89 @@ public Person generatePerson(int index, long personSeed) {
...
+    try {
+      
+      // NOTE: Currently used default demographics for clinicians
+      
+      //Random randomForDemographics = new Random(clinicianSeed);
+      //Demographics city = location.randomCity(randomForDemographics);
+      Map<String, Object> out = new HashMap<>();
+
+      
+      //String race = city.pickRace(randomForDemographics);
```
dehall: I think I understand why all these lines are commented out -- you want to use the same demographics for the city that the provider is in, and there's no easy way to do that at present, right? -- but could you add more comments explaining that so someone seeing this fresh would understand what's going on?

config/checkstyle/checkstyle.xml
```diff
@@ -1,242 +0,0 @@
-<?xml version="1.0"?>
-<!DOCTYPE module PUBLIC
```
dehall: Keep/bring back this file - the one to delete is .checkstyle .  Other than that I think all the comments have been addressed.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -176,72 +194,121 @@ private static Provider findService(Person person,
...
+                parsed.generateClinicianList(1, ClinicianSpecialty.GENERAL_PRACTICE));
+          }
+        }
+        
+        
+        //TODO - determine how many clinicians based off the population
+
+        parsed.clinicianMap.put("GENERAL PRACTICE", 
+            parsed.generateClinicianList(1, "GENERAL PRACTICE"));
+
```
dehall: Not sure this is necessary here - looks like both the `if` and the `else` blocks above set the General Practice value to at least 1.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -176,72 +194,121 @@ private static Provider findService(Person person,
...
+        if (row.get("hasSpecialties") == null || row.get("hasSpecialties").equals("FALSE")) {
+          parsed.clinicianMap.put(ClinicianSpecialty.GENERAL_PRACTICE, 
+              parsed.generateClinicianList(1, ClinicianSpecialty.GENERAL_PRACTICE));
+        } else {
+          for (String specialty : ClinicianSpecialty.getSpecialties()) { 
+            if (row.get(specialty) != "") {
+              if (!row.get(specialty).equals("0")) {
+                parsed.clinicianMap.put(specialty, 
+                    parsed.generateClinicianList(Integer.parseInt(row.get(specialty)), specialty));
```
dehall: While you are looking at the note below can you also clean this section up? Calling `row.get(specialty)` 3 times here means it's time to extract to a variable. Also always compare 2 strings using .equals not ==. Plus the 2 nested `if`s could be merged together. 
Something like:
```
String specialtyCount = row.get(specialty);
if (specialtyCount != null && !specialtyCount.trim().equals("")) {
 clinicianMap.put(....)
}
```

cmbear37:
Hey Jay, 

So the defaults are there because I've run into some troubles of trying to find the provider city in the demographics file. Around 3-4 weeks ago, I tried for a while to generate random demographics like we do for patients, but I can't seem to "find" the city in the file, so Synthea fails. The problem is that in Location, it calls demographics.containsKey("LEOMINSTER") but in the demographics.csv file it's "Leominster". I tried to convert the city to sentence case, do case-insensitive searching, convert the demographics file to all caps (by both iterators and for loops), and convert the demographics HashMap to a TreeMap, but I can't get any of them to work. I tried all of those again today, along with regex/pattern-matching and directly trying to change the demographics.csv, but to no avail. I'm going to leave the code as is today; the solution seemed like it would be straightforward, but it's really escaping me. Let me know if there's anything else I can do or fix.

jawalonoski:
Note: Need to add unit test to load providers from a big state like Texas or California...

PR #366, opened 2018-07-03
Text exporter encounters

laambrosino:
The other branch was messy in general so I deleted it and created this one...  I think there are still conflicts so I can try rebasing like you suggested


src/main/resources/synthea.properties
```diff
@@ -14,9 +14,12 @@ exporter.fhir.transaction_bundle = false
 exporter.hospital.fhir.export = true
 exporter.hospital.fhir_dstu2.export = false
 exporter.csv.export = false
+
 exporter.text.export = false
+exporter.text.export2 = false
```
dehall: Could you use a more descriptive name than export2? It's not immediately clear what this means. Maybe `exporter.text.per_encounter_export` or something like that.

src/main/java/org/mitre/synthea/export/TextExporter.java
```diff
@@ -242,8 +401,33 @@ private static void basicInfo(List<String> textRecord, Person person, long endTi
...
+   * @param person
+   *          The person to export
+   * @param encounter
+   *          The encounter all of the information refers to
+   */
+  private static void encounterReport(List<String> textRecord, Person person, Encounter encounter) {
+    String encounterTime = dateFromTimestamp(encounter.start);
 
-    if (encounter.reason == null && encounter.provider == null) {
+    if (encounter.reason == null){
```
dehall: Looks like this `&& encounter.provider == null)` got lost in the rebase. This should still be in there.

src/main/java/org/mitre/synthea/export/TextExporter.java
```diff
@@ -263,18 +532,44 @@ private static void encounter(List<String> textRecord, Encounter encounter) {
...
+   */
+  private static void conditionpast(List<String> textRecord, Entry condition, Encounter encounter) {
+    String start = dateFromTimestamp(condition.start);
+    if ((condition.stop == 0L || condition.stop > encounter.stop) && (condition.start < encounter.start)) {
+      //checks that the condition hasn't ended by the time of the encounter
+      //and began prior to the encounter
+      String description = condition.codes.get(0).display;
+        textRecord.add("  " + start + " : " + description);
+    } else {
+      return;
```
dehall: Don't need to explicitly return in a case like this - you can just omit the `else`

src/main/java/org/mitre/synthea/export/TextExporter.java
```diff
@@ -188,7 +187,167 @@ public static void export(Person person, long time) throws IOException {
...
+   *                        Activity: Inhaled steroid therapy
+   *                        Activity: Home nebulizer therapy
+   *                        Activity: Breathing control
+   *   
+   * --------------------------------------------------------------------------------
+   */
+  
+  public static void exportEncounter(Person person, long time) throws IOException {
+  
+  /**
```
dehall: This javadoc comment should be above the method signature

src/main/resources/synthea.properties
```diff
@@ -14,9 +14,12 @@ exporter.fhir.transaction_bundle = false
 exporter.hospital.fhir.export = true
 exporter.hospital.fhir_dstu2.export = false
 exporter.csv.export = false
-exporter.text.export = false
+
+exporter.text.export = true
```
jawalonoski: Text exporters should default to `false`

src/main/java/org/mitre/synthea/export/Exporter.java
```diff
@@ -287,4 +295,13 @@ public static String filename(Person person, String extension) {
           + extension;
     }
   }
+
+  public static String filename2(Person person, String encounterNumber, String extension) {
```
jawalonoski: `filename2` is not a good name (try not to ever use numbers in a function name) -- try something more descriptive like `filename_per_encounter(...)` or something.

src/main/java/org/mitre/synthea/export/TextExporter.java
```diff
@@ -263,18 +518,42 @@ private static void encounter(List<String> textRecord, Encounter encounter) {
    * @param condition
    *          The condition to add to the export
    */
-  private static void condition(List<String> textRecord, Entry condition) {
+  private static void condition(List<String> textRecord, Entry condition, Boolean end) {
```
dehall: One last request before we merge this - for these methods where you've added a new `end` parameter please add a description of the parameter to the javadoc above the method.

PR #371, opened 2018-07-05
Lab Panels embedded directly within DiagnosticReport states

dehall:
Updates the DiagnosticReport and MultiObservation states to directly embed Observations, instead of requiring separate states and then just referencing a number of observations to group.

The implementation assumes that the internal embedded observation will be the same as a full Observation state, just minus the "type" and transition. For simplicity we can just call `.process()` on the embedded Observation states and the logic is exactly the same as before. On one hand this makes the change easier, but it may be a little more confusing looking at it fresh. I'm open to reworking the internals if people want me to.


arscan:
Question: Are there any fields in the observation that would make sense to roll up to the `DiagnosticReport` level?  For example, I'm seeing `category` is set to `laboratory` in all the observations for the metabolic panels in the wellness encounter module: https://synthetichealth.github.io/module-builder/#wellness_encounters

Should the category be the same for all Observations within the DiagnosticReport?  Even better, do they always have to be a `laboratory`?  Maybe it is enough to just default to laboratory for now.

Also, noticed that the Observations in there have an `encounter` set in Wellness Enounter, even though that isn't in the spec (and not supported by the state editor, though it does show up in the graph because that was a direct port of the old ruby graphviz generator).

jawalonoski:
From FHIR:

> The DiagnosticReport resource is suitable for the following kinds of diagnostic reports:
> * Laboratory (Clinical Chemistry, Hematology, Microbiology, etc.)
> * Pathology / Histopathology / related disciplines
> * Imaging Investigations (x-ray, CT, MRI etc.)
> * Other diagnostics - Cardiology, Gastroenterology etc.

Also, regarding the nature of mixed content:

> ... the DiagnosticReport resource typically includes additional clinical context and some mix of atomic results, images, imaging reports, textual and coded interpretation, and formatted representations.

So, in general, I think it is possible to have `Observation` resources with categories of `laboratory`, `imaging`, `exam` and maybe others (in addition to other resource types). See http://hl7.org/fhir/2018May/valueset-observation-category.html#expansion

What I'm saying is we _could_ roll up the categories to the DiagnosticReport, but we may regret it later.

PR #377, opened 2018-07-09
Guard date granularity

laambrosino:
I added in two more options for the Date condition: month and date (full date). The tests pass and the program runs, but I'm not sure if there are any other files that need to be edited for this change.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Logic.java
+++ b/src/main/java/org/mitre/synthea/engine/Logic.java
@@ -78,10 +78,11 @@ public abstract class Logic {
   /**
    * The Date condition type tests the current year, month, or date being simulated.
    * For example, this may be used to drive different logic depending on the suggested
-   * medications or procedures of different time periods, or model different frequency of conditions.
+   * medications or procedures of different time periods, or model different frequency
+   * of conditions.
    */
   public static class Date extends Logic {
-    private String type;
+    private String datetype;
     private int year;
     private int month;
     private String date; //must be in format yyyy-MM-dd HH:mm:ss.SSS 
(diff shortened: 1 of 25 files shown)


src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -73,18 +76,51 @@ public boolean test(Person person, long time) {
...
-   * The Date condition type tests the current year being simulated. For example, this may be used
-   * to drive different logic depending on the suggested medications or procedures of different time
-   * periods, or model different frequency of conditions.
+   * The Date condition type tests the current year, month, or date being simulated.
+   * For example, this may be used to drive different logic depending on the suggested
+   * medications or procedures of different time periods, or model different frequency
+   * of conditions.
    */
   public static class Date extends Logic {
+    private String datetype;
```
dehall: Adding a new field to say which thing you're using works just fine, but as you've seen it means that you have to update every existing module to be compliant.   
Another way I would suggest is to instead change the `year` and `month` fields to `Integer` instead of `int`, and then test against whichever one is actually provided, i.e. not null. That gives you the benefit of backward compatibility.

src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -73,18 +76,51 @@ public boolean test(Person person, long time) {
...
+            currentyear = Utilities.getYear(time);
+            return Utilities.compare(currentyear, year, operator);
+          case "Month":
+            currentmonth = Utilities.getMonth(time);
+            return Utilities.compare(currentmonth, month, operator);
+          case "Date":
+            DateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+            java.util.Date testdate = new java.util.Date();
+            java.util.Date currentdate = new java.util.Date();
+            current = Utilities.getDate(time);
```
dehall: This is kind of a roundabout way to compare two dates, and the more conversions the more likely something can go wrong. You have `time` as a long already, so if you could convert the provided string date into a `long` you could compare those directly, and not have to implement `Utilities.getDate` and `Utilities.compare(Date,Date,String)`.

laambrosino: I would need to add `Utilities.compare(long, long, String)` correct?

dehall: You could, but a `long` can be implicitly widened to a `double`, so if you have 2 `long`s you can use the existing `Utilities.compare(Double, Double, String)` as-is. 
```java
long time = 1;
long time2 = 2;
String operator = ">";
boolean x = Utilities.compare(time, time2, operator);
```

Academic note: there are of course `long`s that cannot be represented as `double`s, but those are all greater than 2^53 ( https://stackoverflow.com/a/1848762 ) which represents a timestamp so far in the future we don't have to worry.

src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -73,18 +76,43 @@ public boolean test(Person person, long time) {
...
-      return Utilities.compare(current, year, operator);
+      if (year != null) {
+        currentyear = Utilities.getYear(time);
+        return Utilities.compare(currentyear, year, operator);
+      } else if (month != null) {
+        currentmonth = Utilities.getMonth(time);
+        return Utilities.compare(currentmonth, month, operator);
+      } else if (date != null) {
+        DateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+        java.util.Date testdate = new java.util.Date();
```
dehall: Since the original value of `testdate` is never used, you could remove that part of the line and leave it as just a declaration. Ie just  `java.util.Date testdate;`

src/main/java/org/mitre/synthea/helpers/Utilities.java
```diff
@@ -58,6 +59,12 @@ public static int getYear(long time) {
     return calendar.get(Calendar.YEAR);
   }
 
+  public static int getMonth(long time) {
+    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+    calendar.setTimeInMillis(time);
+    return calendar.get(Calendar.MONTH);
```
jawalonoski: This line of code returns month on a zero-based index... January is 0, and December is 11.

jawalonoski: Please add a unit test that checks for month equality (e.g. `==`). I suspect that you'll need to add 1 to the return value, since in the GMF people will probably enter `12` when they mean December.

PR #378, opened 2018-07-11
Migrate lung cancer probability calculations to GMF

dehall:
Moves the logic that calculates probability of lung cancer from the java-based LifecycleModule to a new submodule. This is intended to be as close to a 1-1 translation of the java logic as possible, but only called once since the value is only read once.

New submodule:
lung_cancer_probabilities

(note the extra transition change in the lung_cancer module just fixes an unreachable state)

### DIFF
--- a/src/main/java/org/mitre/synthea/modules/LifecycleModule.java
+++ b/src/main/java/org/mitre/synthea/modules/LifecycleModule.java
@@ -95,7 +95,6 @@ public final class LifecycleModule extends Module {
       grow(person, time);
     }
     startSmoking(person, time);
-    chanceOfLungCancer(person, time);
     startAlcoholism(person, time);
     quitSmoking(person, time);
     quitAlcoholism(person, time);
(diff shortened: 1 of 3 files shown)


jawalonoski:
Did you compare before and after rates of lung cancer? Wondering if it functions as intended.

dehall:
Short answer: rates before and after are roughly the same. I ran a population of 1000 65-70 yr olds, in master 18 died of LC, with the update 19 died of LC. 0 with LC alive at the end of simulation in either case.

Longer answer: The rates are the same but there are a couple issues. The #s in this module are cited, but the "quitting smoking" numbers are completely invented, and the expected lifespan for people with lung cancer seems very low. As of now, 99% of smokers will quit, so almost nobody goes through the "smoker == true" track. Of those who do get lung cancer, they die very quickly so the #s in the living population don't add up.
I can address those issues here, but I think they've existed for a while.

jawalonoski:
The survival rates for lung cancer are extremely low. The 5 year survival rate is low. Stage 3 and later has median survival time of a year or less.

That you observed that "_99% of smokers will quit_" doesn't pass common sense.

We'll have to address this at a later time.

PR #380, opened 2018-07-11
Terminology

KeeyanGhoreshi:
Value Sets and Terminology implementation code, uses VSAC SVS


src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -353,6 +356,7 @@ public boolean test(Person person, long time) {
   private abstract static class ActiveLogic extends Logic {
     protected List<Code> codes;
     protected String referencedByAttribute;
+    protected String value_set;
```
dehall: Field names in Java code should be camelCased, ie valueSet,, and the field names on the JSON should be snake_cased, ie value_set. It's a little bit "magical" but the engine converts one to the other when the JSON is parsed.

src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -367,9 +371,35 @@ public boolean test(Person person, long time) {
...
+        List<String> valueSetCodes = Terminology.sess.getAllCodes(this.value_set)
+                .stream()
+                .map(code -> {return code.code;})
+                .collect(Collectors.toList());
+
+
+        // Takes intersection of codes in value set and codes present in the record and updates in place.
+        // RetainAll returns true if it changes something, you could use that fact to define a boolean
+        // if you wanted to check if ALL the codes in the value set are present.
+        presentCodes.retainAll(valueSetCodes);
```
dehall: Does this impact the `person.record.present` map in any way? I.e., if codes are removed in this operation, do they also get removed in the `present` map?

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -6,6 +6,7 @@
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import javafx.util.Pair;
```
dehall: IIRC JavaFX is getting pulled out of the base JDK, so if we can avoid referencing the javafx package that would be ideal.

src/main/java/org/mitre/synthea/engine/Logic.java
KeeyanGhoreshi: Yes I checked and it does, I'll change it to make an actual copy so it doesn't affect the original.

src/test/java/org/mitre/synthea/world/concepts/TerminologyTest.java
```diff
@@ -0,0 +1,288 @@
...
+            e.printStackTrace();
+        }
+
+
+    }
+
+    @Test
+    public void makeRequest1() {
+        FormBody formBody = new FormBody.Builder().add("json", "{1:2}").build();
+        Request request = Terminology.makeRequest("http://validate.jsontest.com/", formBody);
```
jawalonoski: I don't want real HTTP requests in the unit tests. Use a fake address with Mocks. For example, https://github.com/square/okhttp/tree/master/mockwebserver

src/test/java/org/mitre/synthea/world/concepts/TerminologyTest.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    @Test
+    public void makeRequestBody() {
+        Map<String, String> alpha = new HashMap<>();
+        alpha.put("yes", "no");
+        alpha.put("thirst", "quenched");
+        RequestBody requestBody = Terminology.makeRequestBody(alpha);
+        Request request = new Request.Builder().url("http://httpbin.org/post").post(requestBody).build();
```
jawalonoski: No real requests.

src/test/java/org/mitre/synthea/world/concepts/TerminologyTest.java
```diff
@@ -0,0 +1,288 @@
...
+//    public void makeGenericPost() {
+//    }
+//
+//    @Test
+//    public void makeGenericGet() {
+//    }
+
+    @Test
+    public void makeRequest() {
+        String url = "http://echo.jsontest.com/title/ipsum/content/blah";
```
jawalonoski: No real requests.

src/test/java/org/mitre/synthea/world/concepts/TerminologyTest.java
```diff
@@ -0,0 +1,288 @@
...
+            assertTrue(code.equals("427089005") || code.equals("E08") || code.equals("250.00"));
+        }
+        boolean allEqual = resultsList.isEmpty() || resultsList.stream().allMatch(resultsList.get(0)::equals);
+        assertFalse(allEqual);
+
+        resultsList.clear();
+
+        if(session.getAuthToken()!=null){
+            for (int i = 0; i < 8; i++) {
+                Code retVal = session.getRandomCode("2.16.840.1.113883.3.464.1003.198.12.1071","CPT",
```
jawalonoski: Do not use CPT.

src/main/resources/synthea.properties
```diff
@@ -41,7 +41,7 @@ generate.timestep = 604800000
 # time is in ms
 # 1000 * 60 * 60 * 24 * 7 = 604800000
 
-generate.database_type = none
+generate.database_type = in-memory
```
jawalonoski: This should be `none` by default.

src/test/java/org/mitre/synthea/world/concepts/TerminologyTest.java
```diff
@@ -239,50 +217,43 @@ public void getServiceTicket() {
...
-            assertTrue(body.contains("validate"));
-            assertTrue(body.contains("true"));
-            assertFalse(body.contains("error"));
-        } catch (IOException e) {
-            e.printStackTrace();
-        }
+
+        // makeRequest doesn't actually make the request, it makes the request object
+        // which can be used to make the request.  Potentially needs to be renamed.
+        Request request = Terminology.makeRequest("http://real.website.com/", formBody);
```
jawalonoski: `makeRequest` should probably be `buildRequest`

src/main/java/org/mitre/synthea/world/concepts/Terminology.java
```diff
@@ -0,0 +1,575 @@
...
+  private static final String VSAC_VALUE_SET_RETRIEVAL_URL = "vsac.nlm.nih.gov";
+  private static final FhirContext ctx = FhirContext.forDstu3();
+  private static final Map<String, ValueSet> valueSets = loadValueSets();
+  private static final Map<String,String> codeLookup = loadLookupTable();
+
+  private static final Logger logger = LoggerFactory.getLogger(Terminology.class);
+  private static boolean connectionFailed = false;
+
+
+  public static Session sess;
```
dehall: This is more personal preference but a more descriptive name here would be nice. This is **a** Session but Session is public so is there a specific reason I would want to use this one vs creating another one?
Also this is public static but not final so there's the possibility that it could change unexpectedly. Maybe it should be final?

src/main/java/org/mitre/synthea/world/concepts/Terminology.java
```diff
@@ -0,0 +1,575 @@
...
+     * it can't find anything.
+     * @param url the uri, oid, or url of the value set
+     * @return the value set that was requested, either from file or VSAC
+     */
+    ValueSet getValueSet(String url) {
+      // Call to get valueSet from a Map
+      ValueSet vs = valueSets.get(url);
+      if (vs == null & authToken != null) {
+        String ticket = getServiceTicket(authToken,client);
+        Response response = Terminology.getValueSet(client,url,ticket);
```
dehall: I'm a little confused on the relationship between the Session class and the Terminology class -- are there really two distinct concepts here? Is it possible to move the methods from one into the other and simplify things?

KeeyanGhoreshi: The session was originally to make an object that had access to the auth token and client since giving the Terminology class access to the auth token might have caused problems if you wanted two different clients.  Also, some classes might not need or want the connection to VSAC, so the Session class used to hold only methods directly related to getting info from VSAC.

Looking at the code now, there's not much of a reason to keep them separated.  I'll make all the methods part of the Terminology class and give it some member variables for the token and client.

src/main/java/org/mitre/synthea/world/concepts/Terminology.java
KeeyanGhoreshi: This will be removed when I address the issue below.

src/main/java/org/mitre/synthea/world/concepts/Terminology.java
```diff
@@ -0,0 +1,575 @@
...
+    URL valueSetsFolder = ClassLoader.getSystemClassLoader().getResource("code_system_lookup.json");
+    try {
+      JsonObject codeSystemLookup = loadJsonFile(Paths.get(valueSetsFolder.toURI()));
+      for(String member : codeSystemLookup.keySet()) {
+        retVal.put(member,codeSystemLookup.get(member).getAsString());
+      }
+    } catch (URISyntaxException e) {
+      e.printStackTrace();
+    }
+    return null;
```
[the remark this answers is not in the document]
KeeyanGhoreshi: I put this function on the backburner because I wasn't sure if swapping OID for URL was really something that was necessary or if doing it with a lookup table was necessarily the best way.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -826,12 +826,11 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
   private static BundleEntryComponent condition(BundleEntryComponent personEntry, Bundle bundle,
       BundleEntryComponent encounterEntry, HealthRecord.Entry condition) {
     Condition conditionResource = new Condition();
-
     conditionResource.setSubject(new Reference(personEntry.getFullUrl()));
     conditionResource.setContext(new Reference(encounterEntry.getFullUrl()));
 
     Code code = condition.codes.get(0);
-    conditionResource.setCode(mapCodeToCodeableConcept(code, SNOMED_URI));
+    conditionResource.setCode(mapCodeToCodeableConcept(code, code.system));
```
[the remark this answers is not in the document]
KeeyanGhoreshi: The codeLookup from the issue below could be used to map the system to a URI.  Would this just be loading the file in the exporter and using it in `mapCodeToCodeableConcept`?

src/test/java/org/mitre/synthea/world/concepts/TerminologyTest.java
```diff
@@ -0,0 +1,255 @@
+package org.mitre.synthea.world.concepts;
```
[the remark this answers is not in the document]
KeeyanGhoreshi: my IDE doesn't show any checkstyle problems with test files so it got overlooked.  I'll change the settings.

src/main/java/org/mitre/synthea/export/FhirStu3.java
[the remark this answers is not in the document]
KeeyanGhoreshi: Also do you think it would be better to just use the lookup table to translate the systems when they're received by `Terminology`, or should I transfer the lookup function to the exporter classes, and make the switch to URI only right before it gets exported?

src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -372,9 +377,28 @@ public boolean test(Person person, long time) {
   public static class ActiveCondition extends ActiveLogic {
     @Override
     public boolean test(Person person, long time) {
+      if (this.valueSet != null) {
+
+        Set<String> presentCodes = new HashSet<>(person.record.present.keySet());
```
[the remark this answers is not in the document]
jawalonoski: That is highly unlikely due the patterns used in the different code systems we use.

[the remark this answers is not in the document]
dehall: It's even impossible between ICD-9, 10, LOINC, and the other systems. But between RxNorm and SNOMED I can't say that it's impossible. And the reason I'm bringing it up here is that a value set could include codes from other systems and that increases the possibility of collisions.

Plus if a collision does occur it's going to cause unexpected and inconsistent behavior in a module which won't necessarily be found quickly or easily. I'd rather try to prevent those kinds of issues before they happen.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -130,6 +131,9 @@
   @SuppressWarnings("rawtypes")
   private static final Map languageLookup = loadLanguageLookup();
 
+  @SuppressWarnings("rawtypes")
+  private static final Map codeSystemLookup = loadLookupTable();
```
dehall: This is ok but it's going to need to be used by both the FHIR 2 and 3 exporters, and potentially more, so it would be good to put it into a common place; we don't want to load the same file into memory twice. I'd propose either ExportHelper or Utilities

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -983,10 +1002,10 @@ private static Type mapValueToFHIRType(Object value, String unit) {
 
     } else if (value instanceof Condition) {
       Code conditionCode = ((HealthRecord.Entry) value).codes.get(0);
-      return mapCodeToCodeableConcept(conditionCode, SNOMED_URI);
+      return mapCodeToCodeableConcept(conditionCode, conditionCode.system);
```
dehall: Changing all these to take in `code.system` will work; I think that plus the changes to the method below will result in identical output to how it works today. But conceptually it feels wrong to me - the intent of providing a separate `system` parameter to the method was that it would support using a different system than what was already in `code.system`. With this change, everywhere it's called it gets passed in the value from `code.system` so the extra parameter isn't needed. 
Can you confirm that's the case and if so, remove the extra parameter from the method and its doco so that it just uses the `from.system` implicitly without having to pass it in? There may someday be a case in the future that we want to provide a different system, but if it's not used we can just clean it up.

jawalonoski: Check this carefully...

nblake989:
@KeeyanGhoreshi any update on this work to add other value sets? I’ve been investigating additional terminology for my use cases for generating more realistic EOB and Claim resources that look like a Payer’s data. I know of the VSAC work, but there is also iMAGIC from SNOMED to crosswalk to ICD10 uses the contextual patient age and gender data that is necessary to accurately point to ICD codes from SNOMED. Curious how I can help get this work back up and running and rebased from the current repo which is about a year and 1000 commits ahead of this PR. THANKS!

jawalonoski:
@nblake989 There has not been any recent work on this branch.

I've been reluctant to merge it in for a few reasons. I'm not going to enumerate all the reasons, but a primary cause of reluctance has been that I've wanted to avoid creating a dependency on VSAC or other external services. There are other non-technical concerns that may or may not be valid.

That being said, it is a useful feature that several users have requested, which is why I haven't closed it or rejected it, and it has remained in limbo.

jawalonoski:
Thank you for the pull request, but it is being closed due to conflicts and inactivity. This is superseded by #709

PR #381, opened 2018-07-12
Codes as observation values (logic aspect)

laambrosino:


src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -170,7 +170,8 @@ public boolean test(Person person, long time) {
     private String operator;
     private List<Code> codes;
     private String referencedByAttribute;
-    private Double value;
+    private Object value;
```
dehall: Just want to point out this change was more necessary than you realized because an Observation value could already be a String, not necessarily a Double.

src/main/java/org/mitre/synthea/engine/Logic.java
```diff
@@ -191,8 +193,16 @@ public boolean test(Person person, long time) {
           return false;
         }
       }
-
-      return Utilities.compare(observation.value, this.value, operator);
+      if (valueCode != null) {
+        value = valueCode;
+      } 
+      if (operator == "is nil") {
```
dehall: In Java, String equality testing should be done with `str1.equals(str2)` , the exception being when you want to check for null. The short version is that == checks whether 2 variables point to the same object in memory, `.equals` checks whether 2 variables point to objects which have the same contents. This is a good overview:  https://stackoverflow.com/a/513839 They actually say use `Objects.equals` which is ok too but not as common.

So these should be `operator.equals("is nil")` and `operator.equals("is not nil")`

PR #384, opened 2018-07-13
Replace VitalSign enums with Strings

dehall:
Enums are a useful tool when one has a comprehensive set of concepts. We have nowhere near a comprehensive set of vital signs, so an enum is likely not appropriate. This changes vital signs to be a String instead, allowing for arbitrary vital signs to be set and used in any module.
To minimize the changes needed across the code, I left in the VitalSign class with a set of predefined constants. These are the vital signs that the core system knows about. If and as these are migrated to GMF we can delete those.

Key point: This should have no outward-facing impact, purely an implementation detail. The module builder already allows free text in `vital_sign` fields.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Generator.java
+++ b/src/main/java/org/mitre/synthea/engine/Generator.java
@@ -374,7 +374,7 @@ public class Generator {
       System.out.format("SYMPTOMS: %d\n", person.symptomTotal());
       System.out.println(person.record.textSummary());
       System.out.println("VITAL SIGNS");
-      for (VitalSign vitalSign : person.vitalSigns.keySet()) {
+      for (String vitalSign : person.vitalSigns.keySet()) {
         System.out.format("  * %25s = %6.2f\n", vitalSign,
             person.getVitalSign(vitalSign).doubleValue());
       }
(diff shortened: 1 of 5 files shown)


jawalonoski:
This PR came out of nowhere. Is this meant to be paired with #386?

> Enums are a useful tool when one has a comprehensive set of concepts. We have nowhere near a comprehensive set of vital signs, so an enum is likely not appropriate.

Perhaps we only support a particular set of vital signs, in which case the enum is comprehensive. Typically, these were only set in the Java code, so an enum is appropriate in that case.

> This changes vital signs to be a String instead, allowing for arbitrary vital signs to be set and used in any module.

Again, presumably this prepares the code for #386?

> Key point: This should have no outward-facing impact, purely an implementation detail. The module builder already allows free text in vital_sign fields.

Module builder capability (or dare I say bug?) should not part of our key justification.

dehall:
Not related to or necessary for 386. It's more in support of having the modules define their own vital signs instead of just relying on a black box in the engine that module authors can't see into. We do allow modules to set vital signs, though that isn't actually used anywhere yet. There's a disconnect there though in that if we allow modules to set vital signs, those values are just going to get overwritten by the engine later.

jawalonoski:
This is pretty out of date. I'm going to close this for now. We can revisit this concept again in the future.

Issue #189, opened 2017-04-17
Expressions


jeffeastman:
Adding an enhancement request that's been discussed on various email threads. I have a working implementation that I use to initialize and age vital signs. In a nutshell, Expressions have the following subtypes that are more fully defined in the attached document:

- UnaryOp (common unary operators)
- BinaryOp (common binary operators)
- Constant (constant values, Ranges & Expressions)
- Variable (vital signs, patient attributes or Ranges)
- Function (multiple argument Expressions with implementation programmatically provided)
- Choice (IF-THEN_ELSE)

[Expressions for Synthea v2.docx](https://github.com/synthetichealth/synthea/files/925613/Expressions.for.Synthea.v2.docx)

cmoesel:
I haven't been a part of these conversations, but incorporating expressions is an interesting idea.  One thing I'd suggest is that you consider alignment with the [HL7 Clinical Quality Language (CQL)](http://www.hl7.org/implement/standards/product_brief.cfm?product_id=400).

This is a health IT expression language designed by MITRE (and sponsored by CMS & ONC) that is primarily used in eCQM and CDS, but is also quite fit for other applications.  In short, it actually has two components: a human-readable authoring language (CQL) and an abstract syntax tree style logical model (ELM).  Looking at your examples, they seem very close to the JSON serialization of ELM -- so I think alignment might not require too much effort and would provide some nice consistency.

Anyway, just throwing that out there as an interested spectator.  Let me know if you'd be interested in seeing some ELM JSON examples.  Also, the CQL repository has some [XML examples](https://github.com/cqframework/clinical_quality_language/tree/master/Examples) available, from which you can easily imagine the corresponding JSON representations.

jeffeastman:
+1 I have also noted this similarity and would be keen to move to CQL/ELM representation. I need to do more reading on these topics so any suggestions or examples would be appreciated.

jeffeastman:
I've made some terminology changes to better align with the ELM nomenclature. Here's a revised document that describes them with the vital_signs module updated. I have this working in my implementation. The json is still a ways from what I perceive it should be from the xml definitions and examples I've found, but I have not found any good json examples to reference. Since you MITRE folks designed ELM, perhaps you could help me get this to a better representation?

[Expressions for Synthea v3.docx](https://github.com/synthetichealth/synthea/files/943129/Expressions.for.Synthea.v3.docx)

cmoesel:
[...]

PR #386, opened 2018-07-13
WIP: Expressions using CQL Engine

dehall:
For discussion:
Implements a simple "expressions" concept (#189) using a CQL engine. SetAttribute, Observations, and VitalSigns now include an "expression" field which gets translated and passed to a CQL engine for evaluation.  The details of the implementation are all abstracted away in a new ExpressionProcessor class. The translation layer allows for selected attributes, identified with a ruby-esque "#{attr}", on the patient to be referenced and added to the evaluation context.

For example, the expression "#{age} / 2 + 7" for a patient with the attribute "age" defined as 26 will result in the following CQL being evaluated:

```
library Synthea version '1'

context Patient


define age: 26
define result: age / 2 + 7
```

My goal here was to enable the kind of basic calculations that are used in the current Java modules, without forcing module writers to care about the details and boilerplate of CQL.

Some open questions:
1. Do we want to allow more complex logic, perhaps allowing people to write raw/multiline CQL to be passed directly to the engine?
2. How should we support referencing vital signs? I do like keeping vital signs separate from attributes, but at the same time if we can define arbitrary vital signs there isn't much distinguishing them and it may be easier to just merge vital signs back into attributes.
3. How should we support referencing the current time? (Ex. to reproduce the current smoking logic which uses a linear formula based on the current year) Do we want to add any convenience items or leverage the existing CQL date operators?
4. Is this missing anything else?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,11 @@ dependencies {
   compile group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
   compile group: 'org.apache.commons', name: 'commons-text', version: '1.2'
 
+  compile group: 'org.opencds.cqf', name: 'cql-engine', version: '1.2.20'
+  compile group: 'info.cqframework', name: 'cql', version: '1.2.20'
+  compile group: 'info.cqframework', name: 'model', version: '1.2.20'
+  compile group: 'info.cqframework', name: 'cql-to-elm', version: '1.2.20'
+
   // Java 9 no longer includes these APIs by default
   compile group: 'javax.xml.bind', name: 'jaxb-api', version: '2.3.0'
   compile 'org.glassfish.jaxb:jaxb-runtime:2.3.0'
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -1056,6 +1056,7 @@ public abstract class State implements Cloneable {
     private String unit;
     private Range<Double> range;
     private Exact<Double> exact;
+    private String expression;
 
     @Override
     public VitalSign clone() {
@@ -1064,6 +1065,7 @@ public abstract class State implements Cloneable {
       clone.exact = exact;
       clone.vitalSign = vitalSign;
       clone.unit = unit;
+      clone.expression = expression;
       return clone;
     }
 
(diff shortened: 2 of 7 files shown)


nanowizard:
Reviving this thread as mathematical expressions for vital sign / attribute assignment will be a necessity for physiology submodule support that I'm working on. I expect models will frequently provide parameters which do not line up specifically with Synthea's set of vital signs, but which are mathematically related.

For example, the initial model I'm working with is a circulatory haemodynamics model which provides left ventricle volume (V_lv). In order to match that up with our Left Ventricular Ejection Fraction (LVEF) vital sign, we'll need an expression of the form "LVEF = max(V_lv) - min(V_lv)".

I suspect we could easily run into situations where more complex mathematical expressions involving square roots, exponentials, trig functions, logarithms, etc. are required. Looking through the repositories it's unclear to me if CQL supports these functions. Some alternative libraries which may provide mathematical expression functionality more specifically are [mXparser](http://mathparser.org/) or [EvalEx](https://github.com/uklimaschewski/EvalEx).

If CQL supports the functions mentioned above and can support sufficiently powerful expressive capability then I'm all for using that. But if not then I'd recommend going with something more math expression specific.

Thoughts?

cmoesel:
CQL supports many mathematical operations.  Check out the following to parts of the reference section in the CQL spec:
https://cql.hl7.org/09-b-cqlreference.html#arithmetic-operators-4
https://cql.hl7.org/09-b-cqlreference.html#aggregate-functions

Do those sets of capabilities seem robust enough for what you'd need?

cmoesel:
One other thing I'll add is regarding the approach of generating a CQL file w/ the pertinent data hard-coded.  The description uses this example:

```
library Synthea version '1'

context Patient

define age: 26
define result: age / 2 + 7
```

This approach requires CQL-to-ELM to recompile every CQL snippet for every Synthea patient.  CQL-to-ELM compilation is fairly fast, but I wouldn't say it's cheap.  It would be more efficient if you could write the CQL snippet with `parameter`s for the variable data.  Then you only need to run CQL-to-ELM once and you pass in the `parameter` values when you execute it for each patient.  Using this approach, your example might look like this:

```
library Synthea version '1'

parameter PatientAge Integer

context Patient

define result: PatientAge / 2 + 7
```

The other approach, which is more complicated but also more robust, would be to use CQL retrieves to get the current Synthea patient's data for calculation -- but then you'd need to implement a data source to work w/ the Synthea data.  If you did that, though, you could use something like CQL's `AgeInYears` operator to automatically get the patient's age.  But... for now, the more reasonable approach is probably `parameter`s.

nanowizard:
Thank you @cmoesel! This is all great info and it seems like CQL should give us everything we need. I'll start working that into my physiology branch and let you know if I run into any hiccups.

PR #388, opened 2018-07-19
added smoking status observation to wellness encounter

laambrosino:


src/main/resources/modules/wellness_encounters.json
```diff
@@ -455,9 +454,115 @@
...
+            "value": true
+          }
+        },
+        {
+          "transition": "Record_Former_Smoker",
+          "condition": {
+            "condition_type": "And",
+            "conditions": [
+              {
+                "condition_type": "Attribute",
```
dehall: Transition options are processed in order and the first one that evaluates to true gets selected, sort of like an `if` in code, so it's not necessary to explicitly specify "smoker == false" in this condition, or "smoker == false and quit_smoking_age is nil" in the third transition below.

dehall:
Re-ran the 1000 patients with the latest changes and confirmed now there are instances of all 3 codes, in particular former smokers. (There are way too many because the quit smoking logic gives people a 1% chance of quitting every week, which equates to around 40% per year, but that's a separate issue)

I'll merge this once the corresponding module-builder PR is merged.

PR #389, opened 2018-07-23
Making private State instVars protected so they can be subclassed

jeffeastman:
I'd like to be able to subclass the Synthea States but private variables and lack of getters/setters makes this impossible. In this commit I've chosen the minimalist change and made the instVars protected. I'm open to getters/setters if preferred.


jeffeastman:
Closing this request while I get a bit further into my investigation. Its looking like the most efficient change is to add getters/setters.

PR #392, opened 2018-07-25
Change Chlorphenerime to RxnCUIs SCDs on the market.

shabiel:
239981 -> 477045
389128 -> 1363309


jawalonoski:
@shabiel, thanks!

There was a transient build error, which resolved cleanly, so the pull request looks good.

I’ll merge this in. There will be a delay between the merge and having these updates show up in the graphical module builder tool, because that update is currently manual.

PR #393, opened 2018-07-26
Guard date granularity

laambrosino:
The way this logic worked before wasn't able to be implemented into the module builder, so this is an edited version that allows the module builder to work with it (it's also a better and more accurate way to do it anyway)

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Components.java
+++ b/src/main/java/org/mitre/synthea/engine/Components.java
@@ -66,4 +66,14 @@ public abstract class Components {
     public String unit;
   }
   
+  public static class DateInput {
+    public int year;
+    public int month;
+    public int day;
+    public int hour;
+    public int minute;
+    public int second;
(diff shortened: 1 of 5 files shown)


jawalonoski:
Is the idea to always have complete dates? Or do you want to support logic for just a month/year or specific year? If the latter, please add some unit tests where only the month/year and year are specified. 

For example, check that the year is equal or before 1970. Does that mean the JSON attributes for month, day, etc are missing, null, empty, zeros? How to distinguish zeros from nulls?

arscan:
Perhaps checking out how she set up the module builder interface will answer your questions.  I deployed a version that has her proposed module builder update here:

http://employeeshare.mitre.org/r/rscanlon/about-me/module-builder/

I did notice that we allow the user to enter in things like 2011/a/5, which should at least cause a warning in the 'analysis' portion, if not just outright reject it.  I haven't had time to look closer though.

PR #402, opened 2018-08-02
Fix rare birth weight exception.

jawalonoski:
Fixes a rare birth weight exception that gets thrown when the random "roll of the dice" is less than a tiny minimum value.

### DIFF
--- a/src/main/java/org/mitre/synthea/world/concepts/BirthStatistics.java
+++ b/src/main/java/org/mitre/synthea/world/concepts/BirthStatistics.java
@@ -152,7 +152,7 @@ public class BirthStatistics {
       
       if (rhispanic == hispanic
           && rsex.equals(babySex)
-          && x < roll) {
+          && ((x < roll) || (data == null))) {
         data = row;
       }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/mitre/synthea/world/concepts/BirthStatistics.java
```diff
@@ -152,7 +152,7 @@ public static void setBirthStatistics(Person mother, long time) {
       
       if (rhispanic == hispanic
           && rsex.equals(babySex)
-          && x < roll) {
+          && ((x < roll) || (data == null))) {
```
dehall: I think actually you just want to flip the expression --> `if roll < x`. The roll is guaranteed to be 0<= roll < max so if x == the max it's not possible for x < roll, and there's overcounting in the cases where 0 <= roll < min. 

To make it more concrete because this was hard to wrap my head around, here are the first 3 lines of the file:

hispanic_mother | baby_sex | lmp_gestational_age | weight | sum
-- | -- | -- | -- | --
TRUE | F | 17 | 31 | 31
TRUE | F | 18 | 31 | 62
TRUE | F | 19 | 50 | 112

So `roll` will be in the range [0,112). If it's in the range [0,31) we want to choose the first row, if in [31, 62) we want the second row, or if it's in [62, 112) we want the third row. For row 1: x == 31, so we want `if roll < 31` --> `if roll < x`. For row 2: x == 62, so we want `if roll < 62` --> again `if roll < x`

PR #404, opened 2018-08-17
Veterans

jawalonoski:
This branch models veterans, VA facilities, and veteran-specific care.

This code was used to produce the synthetic veteran data available at http://portal.lighthouse.vaftl.us

There are merge conflicts with the `master` branch and careful review needs to take place before merging.

- [x] Birthweight fix needs to be cherry-picked from master
- [x] Resolve Conflicts
- [x] RxNorm codes fixed by Sam in master may need to be corrected here prior to merge


src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -714,14 +714,19 @@ private static double reverseCalculateCreatinine(Person person, double crcl, lon
...
+  
   private static final Code NATURAL_CAUSES = new Code("SNOMED-CT", "9855000",
       "Natural death with unknown cause");
 
   private static void death(Person person, long time) {
-    double roll = person.rand();
-    double likelihoodOfDeath = likelihoodOfDeath(person.ageInYears(time));
-    if (roll < likelihoodOfDeath) {
-      person.recordDeath(time, NATURAL_CAUSES, "death");
+    if (ENABLE_DEATH_BY_NATURAL_CAUSES) {
```
dehall: Calling this out to get more attention to it -- I added this because for a veteran population, the target ages were on average higher than the general population, and so to generate a large set of living patients there were a lot of dead patients which added no value and just slowed down the process. Even in a normal population, most use cases only care about living patients, so adding more dead ones which only have a sudden and completely random "natural death by unknown causes" doesn't add value, and again it slows things down and takes up disk space.

My preference is to leave the config setting disabled by default, and if people specifically want dead patients they can enable it.

jawalonoski: Agreed. The setting is disabled by default in the updates I've made.

PR #410, opened 2018-08-23
Coordinate fix

### LINKED ISSUE #395: Latitude and Longitude are reversed
The latitude and longitude of patient coordinates appear to be reversed. It's consistent though as patients are correctly assigned to the geographically closest provider, not some random corner of the state.

From a sample FHIR record for a patient in MA:

```
        "address": [
          {
            "extension": [
              {
                "url": "http://hl7.org/fhir/StructureDefinition/geolocation",
                "extension": [
                  {
                    "url": "latitude",
                    "valueDecimal": -71.026717
                  },
                  {
                    "url": "longitude",
                    "valueDecimal": 42.021617
                  }
                ]
              }
            ],
```

KeeyanGhoreshi:
fixes #395 by swapping `getX()` and `getY()` and uses the functions in the `Person` class that do the same thing as the code in the exporter.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/FhirDstu2.java
+++ b/src/main/java/org/mitre/synthea/export/FhirDstu2.java
@@ -388,7 +388,7 @@ public class FhirDstu2 {
       addrResource.setCountry(COUNTRY_CODE);
     }
 
-    DirectPosition2D coord = (DirectPosition2D) person.attributes.get(Person.COORDINATE);
+    DirectPosition2D coord = person.getLatLon();
     if (coord != null) {
       ExtensionDt geolocationExtension = new ExtensionDt();
       geolocationExtension.setUrl("http://hl7.org/fhir/StructureDefinition/geolocation");
@@ -396,8 +396,8 @@ public class FhirDstu2 {
       ExtensionDt longitudeExtension = new ExtensionDt();
       latitudeExtension.setUrl("latitude");
       longitudeExtension.setUrl("longitude");
-      latitudeExtension.setValue(new DecimalDt(coord.getY()));
-      longitudeExtension.setValue(new DecimalDt(coord.getX()));
+      latitudeExtension.setValue(new DecimalDt(coord.getX()));
+      longitudeExtension.setValue(new DecimalDt(coord.getY()));
       geolocationExtension.addUndeclaredExtension(latitudeExtension);
       geolocationExtension.addUndeclaredExtension(longitudeExtension);
       addrResource.addUndeclaredExtension(geolocationExtension);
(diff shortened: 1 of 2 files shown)


dehall:
Even though we normally think "latitude, longitude" and "x,y" ... latitude isn't the x coordinate. Either way it's going to be confusing, but I'd rather be confusing and consistent with standards. Can you look into what it would take to swap latitude and longitude at the source, such that x = longitude and y = latitude? There may be multiple places - 1 for Person and 1 for Provider. I think all the data files we have use "lat" and "long" correctly so it's wherever those turn into x and y.

KeeyanGhoreshi:
The latitude and longitude get put into a persons attributes in `Location`, which gets the latitude and longitude from `Place`, which seems to load from file by name instead of using X or Y, so that's correct.  From there it puts lat and lon into a `DirectPosition2D`, which accesses the attributes with `getX` or `getY`.  Switching them right there to make lat accessible with `getY` doesn't work, because the way providers are found is using distance which is calculated by an external library (`QuadTree` and by extension `LatLonPointRadius`) which uses the X value as latitude.  Swapping them makes the provider search turn up null since we're in Antartica instead of Massachusetts.  

The best way to do this that I see is just making the methods in the Person class `getLatitude` and `getLongitude` and avoiding all the X and Y confusion in the first place.

dehall:
Renaming the methods would be nice but they're the implementation of the `QuadTreeData` interface so I don't think that's an option. From that interface I see these javadoc comments:

```java
  /**
   * Returns the Java 2D x-coordinate for the longitude.
   *
   * @return the Java 2D x-coordinate
   */
  public double getX();

  /**
   * Returns the Java 2D y-coordinate for the latitude.
   *
   * @return the Java 2D y-coordinate
   */
  public double getY();
```

So we should adhere to that description have have x return longitude and y return latitude. Since choosing providers works today, we'd have to change both the person location and provider location. So both `new DirectPosition2D(lat, lon);` become `new DirectPosition2D(lon, lat);`

src/main/java/org/mitre/synthea/export/FhirDstu2.java
```diff
@@ -388,16 +388,16 @@ private static Entry basicInfo(Person person, Bundle bundle, long stopTime) {
...
       ExtensionDt geolocationExtension = new ExtensionDt();
       geolocationExtension.setUrl("http://hl7.org/fhir/StructureDefinition/geolocation");
       ExtensionDt latitudeExtension = new ExtensionDt();
       ExtensionDt longitudeExtension = new ExtensionDt();
       latitudeExtension.setUrl("latitude");
       longitudeExtension.setUrl("longitude");
-      latitudeExtension.setValue(new DecimalDt(coord.getY()));
-      longitudeExtension.setValue(new DecimalDt(coord.getX()));
+      latitudeExtension.setValue(new DecimalDt(coord.getX()));
+      longitudeExtension.setValue(new DecimalDt(coord.getY()));
```
jawalonoski: Since you reversed X and Y don't you need to change these back? Longitude is now the X coordinate.

KeeyanGhoreshi: Yes, I forgot about the DSTU2 exporter.

PR #411, opened 2018-08-24
Explanation of benefit

KeeyanGhoreshi:
Adding explanation of benefit to the exporter.  Starting out with getting the structure of the EOB correct and setting down some groundwork before going back and removing some of the more hardcoded bits and pieces that aren't already contained in another, already available resource.


KeeyanGhoreshi:
The EOB supports a bit of the insurance adjudication but is makes a lot of assumptions. For example, the deductible is always 0 and the patient always pays 20% of the bill, the allowed charge and the submitted charge are always the same, and some sections, like `Carrier Line Reduced Payment Physician Assistant Code` are absent since Synthea doesn't produce that information.  

The same is true for the care team, which only has the primary practitioner from the claim on the team and doesn't have a qualification.  Some extensions involving medicare specifically are also missing, like the extension for `https://bluebutton.cms.gov/resources/variables/prpayamt/` which covers payment by a primary payer other than medicare for a medicare covered claim.  

Some other bits are hard coded for simplicity, like the identifier and the diagnosis, which is always the principal diagnosis.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +819,284 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    // Set References
+    eob.setPatient(claim.getPatient());
+    eob.setOrganization(claim.getOrganization());
+    eob.setProvider(claim.getProvider());
+    eob.setReferral(new Reference("#1"));
+
+    // Hardcoded identifier
+    eob.addIdentifier()
+        .setSystem("https://bluebutton.cms.gov/resources/variables/clm_id")
+        .setValue("23853631084");
```
dehall: What's the significance of these codes? Should they be hardcoded to a fixed value or generated on a per-eob basis?

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +819,284 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+            .setCode("999")
+            // bluebutton placeholder system
+            .setSystem("https://bluebutton.cms.gov/resources/variables/prvdr_spclty")))
+    );
+
+    Provider provider = person.getProvider(encounter.type, encounterResource
+        .getPeriod()
+        .getEnd()
+        .getTime());
+    eob.setProvider(new Reference().setReference(provider.uuid));
```
dehall: This should probably use the `findProviderUrl()` lookup

src/main/java/org/mitre/synthea/export/FhirStu3.java
KeeyanGhoreshi: The `clm_id` is just a unique identifier I'm pretty sure, I've changed the hard coded value with a `UUID.randomUUID().toString()`.  The second part, the `claim group`, I don't know what it is.  It's present in all the example EOB but the url leads to a 404 and I can't find any reference to a `claim group` on the bluebutton website.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +819,284 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    }
+    eob.setDiagnosis(eobDiag);
+    List<ExplanationOfBenefit.ProcedureComponent> eobProc = new ArrayList<>();
+
+    // get the insurance info at the time that the encounter happened
+    List insuranceList = (List) person.attributes.get("insurance");
+    int age = person.ageInYears(encounterResource.getPeriod().getEnd().getTime());
+    String insurance = (String) insuranceList.get(age);
+    eob.setInsurance(new ExplanationOfBenefit.InsuranceComponent().setCoverage(
+        new Reference().setReference(insurance)
```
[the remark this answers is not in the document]
KeeyanGhoreshi: would it be worth just going to the source and replacing the strings with insurance resources?  the options could be Medicare, a couple of generic private options, and of course None.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    totalCost.setSystem("urn:iso:std:iso:4217");
+    totalCost.setCode("USD");
+    totalCost.setValue(encounter.claim.total());
+    eob.setTotalCost(totalCost);
+
+    // Set References
+    eob.setPatient(claim.getPatient());
+    eob.setOrganization(claim.getOrganization());
+    eob.setProvider(claim.getProvider());
+    eob.setReferral(new Reference("#1"));
```
jawalonoski: While contained references (e.g. `#1`) are allowed, they are gross and I discourage their use. That being said, this is what CMS did, so I'm OK mimicking their output for now.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+
+
+    eob.addIdentifier()
+        .setSystem("https://bluebutton.cms.gov/resources/variables/clm_id")
+        .setValue(UUID.randomUUID().toString());
```
jawalonoski: I believe this is supposed to be the claim id (e.g. `claim.getId()`) and not a random UUID... that being said, this is an abomination, since we should just do something like `eob.setClaim(claim.getId());` rather than use this CMS extension. That being said, you should do both.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    // Hardcoded group id
+    eob.addIdentifier()
+        .setSystem("https://bluebutton.cms.gov/resources/identifier/claim-group")
+        .setValue("93480754268");
```
jawalonoski: I can't find a definition for this Identifer -- can you?

I would set the value to be all nines (e.g. `99999999999`)

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    eob.setDiagnosis(eobDiag);
+    List<ExplanationOfBenefit.ProcedureComponent> eobProc = new ArrayList<>();
+
+    // get the insurance info at the time that the encounter happened
+    String insurance = encounter.claim.insurance.getInsuranceName();
+    eob.setInsurance(new ExplanationOfBenefit.InsuranceComponent()
+        .setCoverageTarget(new Coverage()
+            .addPayor(new Reference()
+                .setReference(insurance))));
+
```
jawalonoski: Remove extra newline

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+      itemComponent.setDiagnosisLinkId(item.getDiagnosisLinkId());
+      itemComponent.setInformationLinkId(item.getInformationLinkId());
+      itemComponent.setNet(item.getNet());
+      itemComponent.setQuantity(item.getQuantity());
+      itemComponent.setUnitPrice(item.getUnitPrice());
+      itemComponent.setEncounter(item.getEncounter());
+
+      // can probably set code based on encounter type with switch statement
+      itemComponent.setService(new CodeableConcept().addCoding(new Coding()
+          .setSystem("https://bluebutton.cms.gov/resources/codesystem/hcpcs")
```
jawalonoski: In this for-loop, these ItemComponents align with Claim items. 

Therefore, if the item refers to a Procedure, then this should be the Procedure SNOMED code. If this item was a Diagnostic Report, then this item should be the proper LOINC code.

HCPCS is a superset of codes which includes CPT, which we do not support. This should not be hard-coded.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+              .setCode("principal")
+              .setDisplay("The single medical diagnosis that is most relevant "
+                  + "to the patient's chief complaint or need for treatment.")
+              .setSystem("https://bluebutton.cms.gov/resources/codesystem/diagnosis-type")));
+      diagnosisComponent.setSequence(claimDiagnosis.getSequence());
+      diagnosisComponent.setPackageCode(claimDiagnosis.getPackageCode());
+      eobDiag.add(diagnosisComponent);
+    }
+    eob.setDiagnosis(eobDiag);
+    List<ExplanationOfBenefit.ProcedureComponent> eobProc = new ArrayList<>();
```
jawalonoski: The variable `eobProc` is unused.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+          code = "21";
+          display = "Inpatient Hospital";
+      }
+      location.addCoding()
+          .setCode(code)
+          //.setSystem("http://hl7.org/fhir/ValueSet/service-place") > if we wanted hl7
+          .setSystem("https://bluebutton.cms.gov/resources/variables/line_place_of_srvc_cd")
+          .setDisplay(display);
+      itemComponent.setLocation(location);
+
```
jawalonoski: Remove extra newline.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+
+      itemComponent.setCategory(new CodeableConcept().addCoding(new Coding()
+          .setSystem("https://bluebutton.cms.gov/resources/variables/line_cms_type_srvc_cd")
+          .setCode("1")
+          .setDisplay("Medical care")));
+
```
jawalonoski: Remove 2 extra newlines.

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+
+
+      // Adjudication
+      if (item.hasNet()) {
+
+        // Assume that the patient has already paid deductible and
```
jawalonoski: Technically, the patient might not have insurance... in which case there is no deductible or coinsurance. Mark this section with a TODO comment.

```java
// TODO: Change deductibles, coinsurance, and payment amounts based on insurance.
```

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    Money payment = new Money();
+    payment.setValue(totalPayment)
+        .setCode("USD")
+        .setSystem("urn:iso:std:iso:4217");
+    eob.setPayment(new ExplanationOfBenefit.PaymentComponent()
+        .setAmount(payment));
+
+    // Hardcoded
+    List<Reference> recipientList = new ArrayList<>();
+    recipientList.add(new Reference("#2"));
```
jawalonoski: Having `#2` and `#3` here will not validate because there are no providers with those IDs.

I suggest you follow the Blue Button approach until we add individual clinicians with fake NPIs:

```json
"requester": {
                            "agent": {
                                "identifier": {
                                    "system": "http://hl7.org/fhir/sid/us-npi",
                                    "value": "999999999999"
                                }
                            }
                        },
                        "recipient": [
                            {
                                "identifier": {
                                    "system": "http://hl7.org/fhir/sid/us-npi",
                                    "value": "999999999999"
                                }
                            }
                        ]
```

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+            .setCode("999")
+            // bluebutton placeholder system
+            .setSystem("https://bluebutton.cms.gov/resources/variables/prvdr_spclty")))
+    );
+
+    Provider provider = person.getProvider(encounter.type, encounterResource
+        .getPeriod()
+        .getEnd()
+        .getTime());
+
```
jawalonoski: Remove extra newline.

jawalonoski:
Also... technically, we should use different profiles for different claims. For example, there are different Blue Button profiles for Inpatient vs. Outpatient claims. See https://bluebutton.cms.gov/developers/#blue-button-implementation-guide

src/main/java/org/mitre/synthea/export/FhirStu3.java
KeeyanGhoreshi: I cannot find the definition either.  I searched the bluebutton site and there's no sign of what it means.  I'm considering removing it but its in all their examples.

jawalonoski: Don't remove it. Please change it to `99999999999`

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -810,6 +820,313 @@ private static BundleEntryComponent encounterClaim(BundleEntryComponent personEn
...
+    Money totalCost = new Money();
+    totalCost.setSystem("urn:iso:std:iso:4217");
+    totalCost.setCode("USD");
+    totalCost.setValue(encounter.claim.total());
+    eob.setTotalCost(totalCost);
+
+    // Set References
+    eob.setPatient(claim.getPatient());
+    eob.setOrganization(claim.getOrganization());
+    eob.setProvider(claim.getProvider());
```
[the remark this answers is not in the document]
KeeyanGhoreshi: Does Synthea even produce practitioner information?  It looks like the provider is just the organization (and `findProviderUrl` returns an organization as well), so setting the provider seems a bit redundant.  I was thinking of taking info from the Synthea `Provider` class like the address and name and putting it in a `practitioner` resource, but I might just leave it out altogether.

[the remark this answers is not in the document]
dehall: Individual practitioners aren't in yet -- see PR #358

build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
   // This dependency is found on compile classpath of this component and consumers.
   compile 'com.google.code.gson:gson:2.8.0'
   compile 'ca.uhn.hapi.fhir:hapi-fhir-base:3.3.0'
+  compile 'ca.uhn.hapi.fhir:hapi-fhir-structures-r4:3.3.0'
```
jawalonoski: No R4 yet. Please remove.

KeeyanGhoreshi: Removing the r4 structures causes unit test to fail and I'm not sure why. The validator throws and error complaining that it doesn't have some r4 class, and the only information I could find online was this [github issue](https://github.com/jamesagnew/hapi-fhir/issues/957) which doesn't really help.

KeeyanGhoreshi:
Yes I can use the JSON version

jawalonoski:
This pull request adds `ExplanationOfBenefit` resources to FHIR STU3 only that comply with the CMS Blue Button 2.0 Implementation Guide.

PR #417, opened 2018-09-13
Feature/fhir r4

glow-mdsol:
A PR for making generation of R4 FHIR resources possible, mostly a copy pasta and getting it to the state where the tests run.
* Added R4 Exporter and Tests
* Added configuration for r4 exports in Exporter.java
* Added a discrete test suite for STU3
* Some code reformatting of test cases, both syntactical (using Before and After, for example) and code format (Per intellij)

### DIFF
--- a/src/test/java/org/mitre/synthea/export/FHIRDSTU2ExporterTest.java
+++ b/src/test/java/org/mitre/synthea/export/FHIRDSTU2ExporterTest.java
@@ -15,6 +15,8 @@ import java.util.List;
 
 import org.hl7.fhir.dstu3.model.DiagnosticReport;
 import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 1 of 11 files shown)


jawalonoski:
Thank for the pull request @glow-mdsol!

Could you please reformat your code with indentations of size `2` instead of `4`?

Indentations of size 2 is our current standard - you might notice that the build is generating a ton of checkstyle warnings about the indentations. Besides getting rid of those warnings, it will be easier to review your pull request to see what actually changed other than whitespace.

Thanks!

glow-mdsol:
Apologies, yes I saw them in the travis-log but hadn't twigged to that as part of the build before launch.  I've now run that and pushed the updated version

src/test/java/org/mitre/synthea/export/FHIRDSTU2ExporterTest.java
```diff
@@ -45,26 +59,24 @@ public void testFHIRDSTU2Export() throws Exception {
     validator.setValidateAgainstStandardSchematron(true);
 
     List<String> validationErrors = new ArrayList<String>();
-    
+
     int numberOfPeople = 10;
     Generator generator = new Generator(numberOfPeople);
     for (int i = 0; i < numberOfPeople; i++) {
       int x = validationErrors.size();
-      TestHelper.exportOff();
```
jawalonoski: You need to turn export off before EVERY patient, because exporting the data is destructive, and if we don't turn it off it here, exporting will automatically get run in the `generator.generatePerson(i)` call. So when you go to run it again a few lines down with `FhirDstu2.convertToFHIR(...)`, the results will be different the second time around.

jawalonoski: I'll take care of that though.

jawalonoski:
Thank you for your contribution!

PR #425, opened 2018-09-27
Fix for the CSVExporter 

SimonUzL:
The format of the costs in the encounter relation depends on the running system. Costs are formatted by a ',' in Germany (for example) instead of a '.' which leads to a "broken" CSV file. My suggestion is to set the locate to US.


jawalonoski:
Thank you for the contribution!

PR #431, opened 2018-10-02
Fix Claim reference

idubrov:
The issue this PR fixes is reference to a claim inside the same bundle given as claim id without resource type (i.e, "123" instead of "Claim/123"), which I think is an invalid reference.

The way I fixed it is by using Bundle.entry.fullUrl (similar to how other references are done).


jawalonoski:
Thank you for the fix!

PR #432, opened 2018-10-04
CSVExporter and CDWExporter Format Issue Fix

SimonUzL:
It was not possible to run

./gradlew build check test

without errors in Germany (for example), because numeric values were formatted with a ',' instead of a '.', which leads to wrong formatted csv files. I recently added an equal pull request which was merged, but i have overseen some formatting issues.


jawalonoski:

Issue #291, opened 2018-04-17
Generate fitness, health and wellness data.


anthonyroach:
I'm interested in generating synthetic data representing a person's self reported and device collected fitness, health and wellness data such as calories burned, exercise duration, heart rate, and intermittent fasting duration.  My interest is driven by the need for health and wellness service providers (like my employer) to have demo and test data for their services and products. 

I think a good first stab at this would be implementing a "Fitness" or "Wellness" module that generates Observations as described here:

http://www.healthintersections.com.au/?p=2487

The purpose of this issue is really to gauge interest in this sort of thing and see whether a module like this would be accepted by the maintainers, and also to gather feedback on the idea. I plan to implement this myself.

dehall:
This sounds like a great idea and we would definitely welcome a pull request.

A few thoughts:
- Currently Synthea requires all Observations to be part of an Encounter, even though in this case there wouldn't be a real encounter. A quick workaround would be to add a "dummy" encounter, but then Synthea would associate that to a provider, which probably isn't what we want. I don't see this as a blocker but we can discuss further as necessary. 
- How much data would you expect to generate? Fitbits can produce tons of heart rate data points and make nice graphs from that -- we haven't done anything close to that volume and level of granularity but it would be interesting to hear your findings.
- Producing realistic but varied vital signs, such as heart rate, is likely to be a challenge in the current iteration of the Generic Module Framework. We'd like to add some capability to use expressions (#189) in the GMF but haven't settled on a final approach yet.

anthonyroach:
Thanks for the feedback, and glad to hear that you like the idea.

- I hadn't realized that an Encounter is required. Certainly creating a "dummy" encounter would work in the short term. Is removing the requirement on having an Encounter something you guys are open to as a separate issue?
- I wasn't planning to generate massive amounts of data like a Fitbit would produce, at least not initially. I was thinking more like sampled values or average values (e.g. resting heart rate each day, and heart rate during exercise). For FHIR it looks like the correct way to model high volume or continuous data from a device is to have a single Observation for some reasonable time period and then have the value be a [SampleData](https://www.hl7.org/fhir/datatypes.html#SampledData) which contains an array of numbers (encoded as a string, interestingly enough). I would imagine supporting SampleData would be more involved than what I'd like to bite off right now.
[...]

PR #448, opened 2018-12-01
Generate vital-signs with a trending generator

tiloc:
This is a work-in-progress illustration of a trending vital-signs generator as discussed in Issue #291. It is implemented for blood pressure, but could easily be extended to weight and other vital signs.

**Key aspects:**
- vital signs are no longer set to a specific double value. Instead, a generator (ValueGenerator) is assigned to the vital sign. The generator can generate a double value at a given time stamp. Different implementations of such generators can use different strategies how they determine the value as a function of time.
- for backwards-compatibility and easier ingestion of this PR, there is an implementation of the generator (ConstantValueGenerator) which will return a constant value for any given time. Likewise, there is a generator (RandomValueGenerator) which will return a random value from a range for any given time. This is being used for all vital signs which have not been really migrated to this new model or don't require any sophistication.

For **blood pressure**, a new sophisticated generator has been created. It can do the following:
1. stitch together an overall curve from a sequence of short-term trends. As implemented right now, it choses a new target value every 2-5 days and then trends towards it.
2. it can limit the rate of change
3. it can reproducibly travel a few days into the past and into the future. It uses a ring buffer to cap memory usage.
4. since it is a function of time and can smoothly determine a value for any given timestamp, it can be used for daily, or even more frequent (ICU/inpatient continuous monitoring) observations without any workarounds.

This screenshot shows a BP graph from daily self-measurements which were obtained without any workarounds.

screen shot 2018-12-01 at 23 23 12


tiloc:
@jawalonoski I have amended the PR with a few cleanups and think it is ready for an initial review. Let me know what you think.

jawalonoski:
I manually merged in this pull request, fixing the conflict in `LifecycleModule.java` and reseting the defaults in `synthea.properties`.

Thank you @tiloc for the contribution!!

🎉

Issue #439, opened 2018-10-24
Dead patients overspill the desired patient count


vadi2:
If you enable `lifecycle.death_by_natural_causes`, you will get `-p #` + dead patients instead of `-p #` total as expected.

jawalonoski:
It is supposed to do that. Death by natural causes just adds a random probability of death to patients that increases as they age. Otherwise, they only die if they hit a `Death` state in one of the modules.

The standard behavior for `-p N` is to generate `N` living patients. If `M` people die along the way, the total patients produced will be `N + M`

vadi2:
I'm still not sure on the reason not to respect `-p N` though - if I want people 1000 people and some of them are dead of natural causes, that's kind of difficult to achieve, isn't it?

jawalonoski:
Today, when you use `-p N` you are really saying "I want N living people."

The reason being... Whoville has a population of 10000. A population count for a city does not include dead people. So, when you generate Whoville, you need 10000 living people.

It is possible to add another configuration parameter that disregards whether or not people are alive. Is that what you are asking for?

vadi2:
Yes! Thanks, that explanation makes sense, and yes, another param would be great.

jawalonoski:
You can now use the `-o false` command line argument to count dead patients as part of the population.

PR #450, opened 2018-12-05
Dead population overspill

dreac0nic:
Adds the flag mentioned in #439, disabling the regeneration of persons if they are generated dead.


jawalonoski:
Thank you for the PR!

Issue #415, opened 2018-09-11
Run a module in isolation


jawalonoski:
Several users, including members of the [OSEHRA Synthetic Patient Data Working Group](https://www.osehra.org/groups/synthetic-patient-data-project-group) have expressed the desire to run a single module in isolation. This would be useful in testing a new module by itself. For example,

```
./run_synthea -m path/to/module/new_disease_module.json
```

A potential issue with this is the occasional interdependencies between modules, often in the form of attributes, and maybe also submodules.

evenaglia:
A PR has been submitted that should address this feature request: https://github.com/synthetichealth/synthea/pull/453

PR #453, opened 2018-12-06
Issue 415: Add cmd line opt -m to specify which modules to load

evenaglia:
New command line option "-m" allows one or more filename wildcards to specify which modules should be loaded.

Changes incidental to this commit include:
* Module loading is no longer done during the static class initialization of Module.java and is now performed lazily.
* Any reference to Module.getModules() to fetch all modules will cause all modules to be loaded, so it is preferable to use Module.getModules(Predicate<String>) is preferable. This predicate is stored as a member variable in Generator.
* TransitionMetrics methods recordStats(Person, long) and printStats(int) have been removed as they implicitly load all metrics and were unnecessary in the contexts in which they were called.


evenaglia:
It should be noted that per the discussion in https://github.com/synthetichealth/synthea/issues/415 it was suggested that support be added to specify a complete path to a module.json to be injected. This was not included in this change to reduce the complexity. module.json files need to be in the classpath:modules folder to be scanned and loaded.

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -311,7 +344,7 @@ public Person generatePerson(int index, long personSeed) {
         }
         
         if (this.metrics != null) {
-          metrics.recordStats(person, time);
+          metrics.recordStats(person, time, modules);
```
jawalonoski:
```diff
- metrics.recordStats(person, time);
+ metrics.recordStats(person, time, modules);
```
This doesn't exactly work, because the modules are being removed in the underlying collection in L318:

```java
iter.remove(); // this module has completed/terminated.
```

If you do this, only the modules left at the end (which could possibly be none except a few core modules) will be recorded -- you'll be missing a lot of data.

jawalonoski: I'll fix this after merging.

jawalonoski:
Thank you for the pull request.

Issue #454, opened 2018-12-07
Flapping test causes intermittent failures


stpapp:
The FHIRSTU3ExporterTest.testFHIRSTU3Export() test intermittently fails, see attached file with the output of the test log.  I managed to isolate a Person.seed value which can reliably recreate the failure:  1544144011558

The test failure seems to be tied to this commit:  https://github.com/synthetichealth/synthea/commit/b19651b996d0b8379f0515aa96daa4ff35d505f6  as I was able to remove that code and observed no issues when running the test with the seed listed above, re-enabling that change re-introduces the test failures.  

Note:
I've altered the code in the test and supporting classes:
Generator.generatePerson():236 was hardcoded to pass the the seed listed above:
    return generatePerson(index, 1544144011558);
FHIRSTU3ExporterTest.testFHIRSTU3Export() var numberOfPeople was changed to "1"

[testConsoleOutput.txt](https://github.com/synthetichealth/synthea/files/2655509/testConsoleOutput.txt)

jawalonoski:
It looks like this is related to the lack of a `Practitioner` being associated with an `Encounter`. Every instance I have found seems to be associated with an emergency encounter generated by the cardiovascular module: `CardiovascularDiseaseModule.java`

PR #455, opened 2018-12-07
WIP: Fix flapping test #454

stpapp:
This PR attempts to fix an intermittent test failure in FHIRSTU3ExporterTest.testFHIRSTU3Export() which seemed to be caused by a malformation of the Fhir extension "onBehalfOf" where the resource type was a "MedicationRequest".  See the attached log file in #454 
Full disclosure:  I'm not very familiar with the Fhir format, and wasn't able to grok exactly what format this particular tag should take, but noticed flipping the assignment from "onBehalfOf" to "onBehalfOfTarget" seemed to resolve the failures I observed so figured I'd throw a PR in for your review and feedback.
Additionally this PR refactors some of the validation/verification code into a common method and adds a test specifically for this intermittent failure by calling out the specific seed I observed the failure from.

Changes:
- attempts to fix the flapping test which fails intermittently
- refactors some common code out for error validation/verification
- adds a new test which tests specifically for the flapping test case by using person.seed: 3579459278049789113L

### DIFF
--- a/src/main/java/org/mitre/synthea/export/FhirStu3.java
+++ b/src/main/java/org/mitre/synthea/export/FhirStu3.java
@@ -1768,7 +1768,7 @@ public class FhirStu3 {
         (org.hl7.fhir.dstu3.model.Encounter) encounterEntry.getResource();
     MedicationRequestRequesterComponent requester = new MedicationRequestRequesterComponent();
     requester.setAgent(encounter.getParticipantFirstRep().getIndividual());
-    requester.setOnBehalfOf(encounter.getServiceProvider());
+    requester.setOnBehalfOfTarget(encounter.getServiceProviderTarget());
     medicationResource.setRequester(requester);
 
     if (medication.stop != 0L) {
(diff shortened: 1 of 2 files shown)


jawalonoski:
Thank you very much for the PR!

However, I do not think this has the correct behavior. 

This change:
```diff
- requester.setOnBehalfOf(encounter.getServiceProvider());
+ requester.setOnBehalfOfTarget(encounter.getServiceProviderTarget());
``` 

Alters the output from:
```json
        "requester": {
          "agent": {
            "reference": "urn:uuid:62cd8de0-8b6f-4107-bcc2-a573263cba13"
          },
          "onBehalfOf": {
            "reference": "urn:uuid:b082884e-bd84-4743-a881-da69e84a7561"
          }
        },
```
to:
```json
        "requester": {
          "agent": {
            "reference": "urn:uuid:4e1cf31f-d259-4965-aab2-3148706b9c5c"
          }
        },
```

So, unless I'm missing something this just removes the `onBehalfOf` attribute, instead of fixing it.

PR #460, opened 2018-12-20
Add append mode and folder-per-run mode to CSVExporter

dehall:
Adds 2 new options to the CSVExporter:

- append mode, where running synthea more than once will cause CSV data to be appended instead of clearing out each time. For instance, this could be used to help produce a single set of CSVs spanning patients across multiple states, since only one state can be run at a time.
- folder-per-run mode, where each run of synthea will have CSVs exported to a subfolder of `output/csv`. Currently the CSVs get reset each time, so if you want to keep a run you have to specifically copy it off somewhere else. This option makes CSVs more like other export formats in that files will persist until manually deleted.

Both options default to false, so the default behavior will be unchanged.


jawalonoski:
@dehall Please add these new configuration parameters to the wiki: https://github.com/synthetichealth/synthea/wiki/Common-Configuration

Issue #473, opened 2019-02-07
failed to build tests when working with JDK 11


Sobhym:
Related issue: #349. 

I couldn't run the tests when working with JDK 11. Here is the output
```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':test'.
> failed to read class file \path\synthea\build\classes\java\test\apptest.class

```
I found that the issue was reported to gradle, refer to,
https://github.com/gradle/gradle/issues/7059

It should have been fixed for gradle 5.0-milestone-1, please upgrade gradle.

Thanks

jawalonoski:
Let me know if this fixes your issue or not. Thanks.

Sobhym:
The issue is fixed. Thanks.

PR #474, opened 2019-02-08
Upgrade to Gradle 5.2.1

jawalonoski:
Issue reporting JDK11 build issue. See #473


jawalonoski:
Ugh... it looks like upgrading Gradle reduces the memory available to the build. See https://github.com/gradle/gradle/issues/7746. Need to increase memory manually for Travis to pass to complete tests. Sigh.

PR #477, opened 2019-02-15
Fix for #443: renamed ID headers to Id in CSV export

### LINKED ISSUE #443: CSV: Excel CSV import returns an SYLK file format error
Apparently some versions of MS Excel will throw errors opening our CSV files if the very first characters in the file are "`ID`" -- as it incorrectly assumes they are SYLK files and fails to properly load them.

Reference: https://answers.microsoft.com/en-us/msoffice/forum/msoffice_excel-mso_win10-mso_o365b/sylk-error-when-opening-csv-files-in-excel/3764f46e-52de-49f2-a77a-21ab9cf84f75

For the time being, I'm setting the "status" of this issue as `Won't Fix` but I'm willing to discuss and interested in feedback from the community.

ramdesh:
Fixes #443 by renaming `ID` headers in all exported CSV files to `Id`

### DIFF
--- a/src/main/java/org/mitre/synthea/export/CSVExporter.java
+++ b/src/main/java/org/mitre/synthea/export/CSVExporter.java
@@ -166,7 +166,7 @@ public class CSVExporter {
    * @throws IOException if any IO error occurs
    */
   private void writeCSVHeaders() throws IOException {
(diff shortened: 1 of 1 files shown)


jawalonoski:
Thank you for the pull request @ramdesh

Issue #472, opened 2019-02-06
Issues with Prevalence Report


dehall:
Observed a few issues with generating a report based on a modified prevalence template.
 - [ ] with age group column set to `*`, occurrences and population were orders of magnitude higher than the actual population. Expected was similar #s as shown in the "all conditions" list at the bottom
 - [ ] occurrences for one row had occurrences > population for a prevalence > 1
 - [ ] the "all conditions" population is 1 less than the actual generated population. not always the case so maybe a bug with generation instead?

See attached for a run generated with 100 patients: 

[sample_prev_report.csv.txt](https://github.com/synthetichealth/synthea/files/2838508/sample_prev_report.csv.txt)

dehall:
Update: 
The first issue is because unlike the other columns there's no specific support for setting the age group column to `*` and it wound up counting rows in the attributes table. I'll update the code to support that.

Not sure what exactly was going on in the second case, but I'll rewrite the queries in a way that should prevent it. Currently there are 2 separate queries for population and occurrences; for performance and consistency it makes more sense to just have one.

I think the issue with the "all conditions" population is that patients may have a date of death that's in the future so the query looking for `date of death is not null` doesn't count them. The trouble with a future date of death here is that we lose the context of the run and don't know where the cutoff between present and future is. My proposed fix is to clear out the future date and cause of death from the patient object if they are still alive at the end of the simulation.

dehall:
closing this as the prevalence report has now been removed

PR #480, opened 2019-02-20
Prevalence report tweaks

dehall:
- Fixes the queries used in the prevalence report to ensure consistency and accuracy -- see issues mentioned in #472. Previously there were 3 different queries, now there's only 1.
- Adds a "code" column in addition to the "display" column. These are used in the query with an OR, so it looks for rows where either the code or the display matches. This means either column could be left empty, but I filled in the codes in the template anyway.
- In the DeathModule, remove the cause and time of death if the person isn't actually dead yet at the end of simulation (ie, if a Death state indicated some time left to live, which has not yet elapsed).  This is done because if the date of death is set, we don't have a good way to later distinguish those records with a date of death that hasn't occurred yet. I note that this could interfere with a "save and reload patients" capability, so if that gets built this should be revisited.

Sample report generated with seed 0 and 100 population:
[prev_data1550669925466.csv.txt](https://github.com/synthetichealth/synthea/files/2884787/prev_data1550669925466.csv.txt)

Addresses #472

### DIFF
--- a/src/main/java/org/mitre/synthea/modules/DeathModule.java
+++ b/src/main/java/org/mitre/synthea/modules/DeathModule.java
@@ -33,6 +33,9 @@ public class DeathModule {
       // otherwise it's tough to tell after the fact whether this patient did die or will die
       person.attributes.remove(Person.CAUSE_OF_DEATH);
       person.record.death = null;
+      if (person.hasMultipleRecords) {
+        person.records.values().forEach(hr -> hr.death = null);
+      }
       // TODO: if stopping/restarting/reloading becomes a thing, this may have to be reworked
     } else if (person.attributes.containsKey(Person.CAUSE_OF_DEATH)) {
       // create an encounter, diagnostic report, and observation
--- a/src/main/java/org/mitre/synthea/modules/DeathModule.java
+++ b/src/main/java/org/mitre/synthea/modules/DeathModule.java
@@ -28,16 +28,7 @@ public class DeathModule {
    * @param time - the time of the death exam and certification.
    */
   public static void process(Person person, long time) {
-    if (person.alive(time)) {
-      // remove the death from the record.
-      // otherwise it's tough to tell after the fact whether this patient did die or will die
-      person.attributes.remove(Person.CAUSE_OF_DEATH);
-      person.record.death = null;
-      if (person.hasMultipleRecords) {
-        person.records.values().forEach(hr -> hr.death = null);
-      }
-      // TODO: if stopping/restarting/reloading becomes a thing, this may have to be reworked
-    } else if (person.attributes.containsKey(Person.CAUSE_OF_DEATH)) {
+    if (!person.alive(time) && person.attributes.containsKey(Person.CAUSE_OF_DEATH)) {
       // create an encounter, diagnostic report, and observation
 
       Code causeOfDeath = (Code) person.attributes.get(Person.CAUSE_OF_DEATH);
--- a/src/main/java/org/mitre/synthea/modules/DeathModule.java
+++ b/src/main/java/org/mitre/synthea/modules/DeathModule.java
@@ -28,7 +28,13 @@ public class DeathModule {
    * @param time - the time of the death exam and certification.
    */
   public static void process(Person person, long time) {
-    if (!person.alive(time) && person.attributes.containsKey(Person.CAUSE_OF_DEATH)) {
+    if (person.alive(time)) {
+      // remove the death from the record.
+      // otherwise it's tough to tell after the fact whether this patient did die or will die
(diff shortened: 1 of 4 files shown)


jawalonoski:
Hmmm.... I'm not sure about those DeathModule changes. In the Exporter, we have code to filter out future death information. Is there a way to reuse that instead? Or if we keep the DeathModule change, then the Exporter code potentially becomes redundant.

https://github.com/synthetichealth/synthea/blob/6dbf88c6ab83425fb5964c99f04a8d6a1a921ecf/src/main/java/org/mitre/synthea/export/Exporter.java#L320-L330

Also, it is now possible that patient has multiple records with the new `exporter.split_records` property, so if you want to erase future death, you might want to account for all the records.

https://github.com/synthetichealth/synthea/blob/6dbf88c6ab83425fb5964c99f04a8d6a1a921ecf/src/main/java/org/mitre/synthea/world/agents/Person.java#L84-L87

dehall:
@jawalonoski Finally got a chance to address this, so it's now ready for a re-review.

 - I removed the part of the Exporter filter that looks for future dated "Cause of Death" Observations and Reports, since those won't be created with future dates anymore, but I left in the part that preserves cause of death entries in general so that those will always be left in the record even if more than *n* years old. (ex., if a person died 20 years ago but the filter keeps 10 years, it will still keep the cause of death info. It's a bit of an edge case but it ensures that all records for deceased patients will still indicate cause of death)
 - Added a line to clear out date of death on split records.
 - Rebased since there was a lot new and wanted to make sure there was nothing conflicting

src/main/java/org/mitre/synthea/modules/DeathModule.java
```diff
@@ -28,7 +28,16 @@
...
+    if (person.alive(time)) {
+      // remove the death from the record.
+      // otherwise it's tough to tell after the fact whether this patient did die or will die
+      person.attributes.remove(Person.CAUSE_OF_DEATH);
+      person.record.death = null;
+      if (person.hasMultipleRecords) {
+        person.records.values().forEach(hr -> hr.death = null);
+      }
+      // TODO: if stopping/restarting/reloading becomes a thing, this may have to be reworked
+    } else if (person.attributes.containsKey(Person.CAUSE_OF_DEATH)) {
```
jawalonoski: I don't understand why you are changing this code.

If the person is alive right now, you are removing their cause of death? If you do that, in the future, when they actually die, we won't know why. (in the save and restore scenario that @hadleynet is working on)

And it doesn't even look like you are examining the cause of death attribute in the prevalence report anyway...

One thought is that in the prevalence report, instead of SQL like `WHERE p.DATE_OF_DEATH is null` you could do something like `WHERE p.DATE_OF_DEATH is null or p.DATE_OF_DEATH > ?` and you set the date to when the simulation stopped (which could be passed in with the Generator)

dehall: Yeah that SQL is the reason why this section was modified. The trouble with doing `where DATE_OF_DEATH > ?`  is that the DB doesn't contain any info on when the simulation stopped so there's no way to tell after the fact which patients with a death date were actually still alive at the end of the simulation. Not that the DB is a majorly used feature but being able to tell which patients are alive and which are dead is an important distinction.

jawalonoski: I know, but the exporters do know, and you could pass that information into the PrevalenceReport exporter.

dehall: I'm thinking of the file-based DB option where the patient records are persistent but the end date of the simulation isn't. That said I don't know if anyone actually uses that feature.
I had forgotten that all the other exporters reference the simulation stop time to check alive/dead status so I can do the same thing here.

Maybe this would be a separate PR but what if I added a new table in the DB that tracked the details of the simulation? It would include things like the seed, the simulation stop time, maybe some run settings, and an ID to link back to the patients and providers from that run.

hadleynet: FWIW, in the serialization-support branch, I added a "lastUpdated" field to Person to track the end of each simulation run so I can restart where it stopped after a serialize/deserialize.

dehall: Reverted this section and updated the report SQL with `or p.DATE_OF_DEATH > ?`

jawalonoski:
@dehall is this still needed?

dehall:
@jawalonoski It's still relevant to autotuning. I eventually want to explore alternatives to the in-memory database that the PrevalenceReport uses because it gets too big on populations of any nontrivial size, but for now this is still useful.

dehall:
Closing this -- going forward the database will be deprecated, so no point in updating any of the reports that depend on it

PR #482, opened 2019-02-22
removed extra comma delimiter in provider csv exporter

fdefalco:
An extra comma delimiter at the end of each record was being exported to the provider csv file.


jawalonoski:

Issue #483, opened 2019-02-28
Multiple providers with same coordinates, only one ever chosen


dehall:
In generating a large number of patients for Austin, Texas, it was observed that only 2 organizations were ever written to the organizations.csv output file. Investigation revealed that all 1,249 primary care facilities listed in Austin TX in the provider input files have the exact same coordinates of 30.239513, -97.69127, and all 15 hospitals in Austin TX have the same coordinates of  30.30718, -97.756. 

I see 2 issues here:
1) The coordinates for these locations do point to somewhere in Austin Texas, but it's not the correct location for all these facilities. Ideally the coordinates could be re-geocoded and fixed.
2) Certain facilities might actually exist at the same coordinates (eg, different rooms or floors in the same building) so there should be some way for all facilities at the same coordinates to be selected. Right now it seems like only one will consistently be chosen.

Note that this was specifically observed in generating patients in Austin, Texas, but this is likely to be an issue for many other locations as well.

jawalonoski:
If a volunteer wants to correctly geolocate the addresses in the provider files (see `src/main/resources/providers/*`) that would be appreciated.

Alternatively, I think that a reasonable short-term solution would be to use the patient's seeded random number generator to select one of the equidistant providers at random.

Long-term, there are some plans to change the provider selection algorithm away from "geographically closest location" to be more complicated an include factors like quality, insurance status, insurance network, et cetera.

PR #484, opened 2019-03-05
Randomly choose a provider if there is more than one at a given location

dehall:
Initial workaround for #483 until all providers coordinates are updated and verified. Updates the logic that selects a "closest" provider by keeping a list of providers at the current closest distance, and then at the end it chooses one randomly.

### DIFF
--- a/src/main/java/org/mitre/synthea/world/agents/Provider.java
+++ b/src/main/java/org/mitre/synthea/world/agents/Provider.java
@@ -179,24 +179,33 @@ public class Provider implements QuadTreeData {
     DirectPosition2D coord = person.getLatLon();
     List<QuadTreeData> results = providerMap.queryByPointRadius(coord, searchDistance);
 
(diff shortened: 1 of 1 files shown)


jawalonoski:
Looks good. Did you test it to see if it behaves as expected?

dehall:
Anecdotally, it does result in more and different providers being chosen. I can write a unit test if you'd like but it's tough to test randomness.
That said I was surprised that there were still fewer providers chosen than I expected, though I don't have a good mathematical sense of how many providers should be expected. For example, given the 1200+ primary care locations in Austin, TX, and _n_ patients, how many actual providers _p_ should be selected? Obviously 1 <= _p_ <= 1200 and given the caching synthea does _p_ <= _n_ , but I've forgotten everything I once knew about combinatorics to make a more intelligent prediction.

jawalonoski:
Good enough.

PR #492, opened 2019-03-21
Incorporate anemia in other disease modules

petercyli:
Updated anemia submodule for incorporation into other disease modules.  Modules affected include:
- anemia___unknown_etiology.json
- congestive_heart_failure.json
- colorectal_cancer.json
- copd.json
- cystic_fibrosis.json
- dialysis.json
- hypothyroidism.json
- lung_cancer.json
- lupus.json
- metabolic_syndrome_care.json
- pregnancy.json


petercyli:
In addition, the primary_atrophic_hypothyroidism.json is removed since it is a duplicate of the hypothyroidism.json.
By the way, it was confusing to have the metabolic_syndrome_care.json as the filename for the "metabolic syndrome standard of care" module when the Module Builder always generate the filename based on the disease model name.  It would have been better to just keep the name as created by the Module Builder.

jawalonoski:
Thank you for the pull request! 👍 

Some of the module filenames were created before the module builder existed... e.g. metabolic syndrome. Any filename mismatches are due to legacy reasons.

src/main/resources/modules/colorectal_cancer.json
```diff
@@ -948,7 +948,83 @@
...
+              "distribution": 0.66
+            },
+            {
+              "transition": "Oncologist_Encounter",
+              "distribution": 0.34
+            }
+          ],
+          "condition": {
+            "condition_type": "Attribute",
+            "attribute": "colorectal cancer stage",
```
jawalonoski: This attribute is `colorectal cancer stage` but needs to be `colorectal_cancer_stage`

PR #500, opened 2019-04-03
Respect SNOMED-CT coded medication orders within FHIR exports

johngrimes:
This pull request adds the ability to specify "SNOMED-CT" within the "system" field of a code within a MedicationOrder state definition.

With this change, the FHIR DSTU2, STU3 and R4 generators will now correctly output the SNOMED CT URI ("http://snomed.info/sct"), rather than ignoring the value in the module and defaulting to the RxNorm URI.


jawalonoski:
Are there any modules that use SNOMED for medications? By default, there shouldn't be. Are you running your own modules with SNOMED medications?

johngrimes:
No, there are no default modules that use SNOMED CT for medications, as it is not currently supported. Specifying a "SNOMED-CT" and a SNOMED CT cdode within a module currently results in a RxNorm URI being emitted to the FHIR output, for medications.

We have a requirement to use SNOMED CT medication codes for our own modules, so it would be great if we could do this simply by specifying a system in the code within the state definition, in the same way that you currently can for encounter reasons, conditions, etc.

jawalonoski:
Thanks for the pull request!

PR #507, opened 2019-04-24
Use OpenJDK in Travis

jawalonoski:
Try out OpenJDK 8, 11, and 12 in Travis...


jawalonoski:
This PR was a proof of concept, but I'm not going to merge it because it triples the Travis build time. If users have future issues with JDK 11 or JDK 12 we can relook at this.

PR #508, opened 2019-04-26
Fix FHIR Encounter class codes

eedrummer:
When exporting Encounter FHIR resources, Synthea is currently outputting
a lowercased version of the EncounterType enum for the Encounter.class
element. However, this value should be from the ActEncounterCode VS:
http://www.hl7.org/implement/standards/fhir/v3/ActEncounterCode/vs.html

This change adds the appropriate value into the EncounterType enum and
updates the STU3 and R4 exporters to use it.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/FhirR4.java
+++ b/src/main/java/org/mitre/synthea/export/FhirR4.java
@@ -619,7 +619,7 @@ public class FhirR4 {
       encounterResource.addType(mapCodeToCodeableConcept(code, SNOMED_URI));
     }
 
-    encounterResource.setClass_(new Coding().setCode(encounter.type));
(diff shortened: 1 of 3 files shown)


jawalonoski:

If you're going to all the trouble to use the `http://terminology.hl7.org/CodeSystem/v3-ActCode` code system in R4, do you want to specify that in the `Coding.system`?

Same question with `http://hl7.org/fhir/v3/ActCode` in STU3?

eedrummer:
That's a good point. I just pushed a commit that adds the system for STU3 and R4. It appears that both FHIR versions refer to the same value set.

jawalonoski:
Thanks for the PR!

PR #512, opened 2019-04-30
Support the creation of MedicationAdministration resources for MedicationOrder states

okeefm:
Companion PR to https://github.com/synthetichealth/module-builder/pull/245. Adds support in the State engine, HealthRecord, and FHIR Exporters for an `administration` attribute on `MedicationOrder` states.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -780,13 +780,15 @@ public abstract class State implements Cloneable {
    * prescribed. MedicationOrder states may only be processed during an Encounter, and so must occur
    * after the target Encounter state and before the EncounterEnd. See the Encounter section above
    * for more details. The MedicationOrder state supports identifying a previous ConditionOnset or
-   * the name of an attribute as the reason for the prescription.
+   * the name of an attribute as the reason for the prescription. Adding a 'administration' field
+   * allows for the MedicationOrder to also export a MedicationAdministration into the exported FHIR record.
    */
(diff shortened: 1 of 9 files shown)


jawalonoski:
Thanks for the pull request!

It looks like you have unit tests, but do you have any that actually test the export of the MedicationAdministration resource and validate it?

okeefm:
I don't currently, because I wasn't sure how to do that in the test framework that already exists. It looks like the exporter tests generate random patients and test their export, so I'm not sure what the best way is to test the export of a specific part of the `HealthRecord`.

jawalonoski:
Well, the very best way would be to have a module that actually produces a `MedicationAdministration`, because then it would happen automatically. Second best would be to try and run the exporter on the patient in your existing unit tests and validate the exported resources.

okeefm:
I do have a (sub-)module that contains states with `administration` attributes, which I included in a couple of top-level modules. It's not finalized (e.g. I haven't validated all of the states with a clinician), but it's at least quasi-empirically based. If I add that in, that will take care of the testing automagically?

mgramigna:
@jawalonoski I've been working with @okeefm on the remaining testing for this PR.

I just pushed some changes that add the "administration" flag to the `MedicationOrder`s for `lung_cancer` and `general_anathesia`. This way, the FHIR exporter tests will generate/validate the corresponding `MedicationAdministration` resources at least some of the time.

I talked with @dehall about where it might make the most clinical sense to actually have an administration for a certain medication, and we settled on these two modules for now. Let me know if you think this is sufficient.

PR #516, opened 2019-05-08
Enabled elasticsearch export feature. This feature is triggered by --…

santhoshjohn78:
…exporter.elastic.export true option which is passed in the command line. Check synthea.properties for all the available options to export to the elasticsearch instance of your choice.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,6 +62,9 @@ dependencies {
   compile "org.slf4j:slf4j-api:1.6.1"
   compile "org.slf4j:slf4j-nop:1.6.1"
 
+// https://mvnrepository.com/artifact/org.elasticsearch.client/elasticsearch-rest-client
+  compile 'org.elasticsearch.client:elasticsearch-rest-high-level-client:5.6.16'
+
   // Use JUnit test framework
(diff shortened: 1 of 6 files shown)


jawalonoski:
Thank you for this pull request, which seems to add the ability to automatically push FHIR DSTU2 into Elastic Search after the FHIR DSTU2 exporter is finished.

We don't typically accept pull requests of this type -- we consider the post-processing, extract, transformation, and loading (ETL) of Synthea data into external systems to be outside the scope of the project. While the use of Elastic Search is important to you, it may not be to others, who might prefer something like OMOP, VistA, Document DBs (e.g. Mongo DB), or something else entirely.

While we aim to support multiple formats (e.g. FHIR, C-CDA, CSV), we do not intend to directly support external systems integrations.

jawalonoski:
@santhoshjohn78 thank you for the pull request.

But, I'm going to close this pull request as I've explained above.

I've added a notice to the wiki to avoid pull requests like this in the future. https://github.com/synthetichealth/synthea/wiki/Contributing#external-systems-integrations

PR #517, opened 2019-05-15
Weight loss

eedrummer:
This is a new java module that simulates weight loss. Probabilities around weight loss are managed in `synthea.properties`. They didn't seem like a good fit for the biometrics file, but I'm happy to move them there if people thing that is where they should belong.

The way that the module works currently is that if someone is over 5 years old and their BMI is over 30, there is a 49.3% chance that they will try to lose weight. If they decide to lose weight, there is a 60.5% chance that they will adhere to weight loss measures and start to lose weight. For those that adhere to weight loss, they will lose between 7% and 10% of their weight, with pediatric cases never being allowed to drop below the 60th percentile weight for their age in months. For those who successfully lose weight, there is a 20% chance that they will keep it off. For those that do, they keep it off for life. For those who don't, their weight will revert to the original over a 5 year period. For cases where weight loss starts before age 20 but reversion will conclude after age 20, the individual will head towards the 240 month weight for their percentile.


jawalonoski:
Thank you for such an interesting pull request @eedrummer!

Couple of thoughts:
- I do think it makes more sense for those parameters to go in the `biometrics.yml` file, since the existing height/weight parameters are in there (near the top, under lifecycle). Please move those over.
- Did you look at the checkstyle complaints? I'm surprised it didn't go into hysterics over your wildcard import.
- Have you tried plotting the weight gain/loss over time of any patients? I think that be very interesting to look at as a visual sanity check.

eedrummer:
I've fixed the checkstyle issues and moved the config over to BiometricsConfig.

As an example, you can see the weight changes for four different patients:
weight-loss-charts

jawalonoski:
Thanks for the pull request @eedrummer!!

PR #523, opened 2019-06-05
Lookuptabletransitionrebase

RobertScalfani:
Addition of Lookup Table Transition


RobertScalfani:
Fixed Coverage and Tests

PR #531, opened 2019-06-21
Feature/add s3 write

BrianKolowitz:
Hi, we wanted to do something like ./run_synthea -p 100000000 and skip local file storage so we added an s3 writer. the code is now extensible and can add other writers such as an HTTPWriter which could post direct to a FHIR server. The PR was getting pretty big  so we stopped just short of doing that.

PR #532, opened 2019-06-21
Feature/provider seed

BrianKolowitz:
I added a feature to generate a distinct set of providers across different patient runs. I'd like the same provider population to be used independent of which process or seed is used to generate the patients.

To test this i ran this command 3 times ```./run_synthea -p 100 -o false``` and named the files
provider.1.csv
provider.2.csv
provider.3.csv

I then wrote this python script to check for intersections of providers across the sets

```python
import os
import glob
import csv

csv_path = os.path.abspath('synthea/output/csv')
files = glob.glob(f'{csv_path}/providers.*.csv')

provider_sets = []
for index, file in enumerate(files):
    provider_sets.append(set())
    with open(file) as f:
        dict_reader = csv.DictReader(f)
        for row in dict_reader:
            provider_sets[index].add(row['NAME'])

print(len(provider_sets))
print(provider_sets[0].intersection(*provider_sets[1:]))
```

which output
```
3
set()
```

i then ran this command 3 times ```./run_synthea -p 100 -o false -cs 1``` and named the files
provider.4.csv
provider.5.csv
provider.6.csv

which output
```
3
{'Steven797 Turcotte120', 'Modesto621 Feil794', 'Concepción765 Domínquez36', 'Blaine377 Hickle134', 'Shayla126 Kirlin939', 'Josefa6 Mayer370', "Kerry175 O'Connell601", 'Cameron381 Paucek755', 'Williams176 Hartmann983', 'Marci572 Huels583', 'Mammie808 McKenzie376', 'Marinda592 Romaguera67', 'Sherwood961 Gerlach374', 'Rodrigo242 Jakubowski832', 'Juana825 Raya927', 'Minna389 Hills818', 'Vern731 Yundt842', 'Stuart913 Dietrich576', 'Genevie117 Johns824', 'Jadwiga271 Tillman293', 'Adam631 Sporer811', 'Lakesha141 Durgan499', 'Clementina782 Cronin387', 'Bonita405 Nikolaus26', 'Bettina657 Feil794', 'Jeni110 Heller342', 'Kenneth671 Pagac496', 'Laura391 Escamilla823', 'Gabriela205 Santiago500', 'August363 MacGyver246', 'Tiffani287 Bartell116', 'Jordan900 Morissette863', 'Esteban536 Kub800', 'Huey641 Brakus656', 'Prince887 Schneider199', 'Parker433 Predovic534', 'Reginald96 Feest103', 'Vance413 Gulgowski816', 'Dannie881 Turcotte120', 'Erica194 Gorczany269', 'Donn979 Boyer713', 'Antonia30 Fajardo595', 'Kiara978 Thiel172', 'Jackelyn13 Russel238', 'Detra426 Jerde200', 'Ollie731 Reichel38'}
```


jawalonoski:
Thank you for another useful pull request! 👍 

I quickly scanned this -- no obvious issues. I'll make a more careful review and merge after #531 and possibly a few others.

BrianKolowitz:
Thanks. I'm open to suggestions on how to make the persistent UUID better.

src/main/resources/synthea.properties
```diff
@@ -23,7 +23,8 @@ exporter.hospital.fhir_dstu2.export = false
 exporter.practitioner.fhir.export = true
 exporter.practitioner.fhir_stu3.export = false
 exporter.practitioner.fhir_dstu2.export = false
-exporter.csv.export = false
+exporter.practitioner.preserve_uuid = true
+exporter.csv.export = true
```
jawalonoski: Please do not change the default setting of `exporter.csv.export`. It should be `false` by default.

src/main/java/org/mitre/synthea/world/agents/Clinician.java
```diff
@@ -37,17 +37,22 @@
...
   public Map<String, Object> attributes;
   private ArrayList<String> servicesProvided;
   private int encounters;
   public long populationSeed;
   
-  public Clinician(long seed) {
-    this.uuid = UUID.randomUUID().toString();
-    this.seed = seed; // keep track of seed so it can be exported later
-    random = new Random(seed);
+  public Clinician(long clinicianSeed, Random clinicianRand, Boolean preserveClinicianUUID, long identifier) {
```
jawalonoski: Hmm... it seems like `Boolean preserveClinicianUUID` is an unnecessary argument... doesn't it?

I'm thinking we should just use the ` this.uuid =  new UUID(clinicianSeed, identifier).toString();` assignment in all cases, because a `clinicianSeed` will always be passed in (by default it matches the runtime seed).

If the user provides a seed at runtime with `-cs N` then don't we always want a reproducible UUID based on that seed, regardless of whether or not the `exporter.practitioner.preserve_uuid` property was set?

Am I missing something?

src/main/resources/synthea.properties
BrianKolowitz: fixed

src/main/java/org/mitre/synthea/world/agents/Clinician.java
BrianKolowitz: Good point. It's not necessary if we're okay with the way UUIDs are generated. I removed it from the code.

jawalonoski:
Thanks for the pull request!

PR #538, opened 2019-07-01
New hypertension module with controlled high blood pressure via medic…

petercyli:
…ation treatment

Made changes based on Jay's suggestions:
•	The attribute “bp_treated?” is being READ in CardiovascularDiseaseModule, but NOTHING writes there. Perhaps references to this should be changed to READ from your new “blood_pressure_controlled” attribute.
•	There should be no SecureRandom in TrendingValueGenerator. SecureRandom is normally best practice (for cryptographic purposes), but in Synthea this is really bad, because it makes it hard to reproduce problems (we want the random generators to be the same every time). This wasn’t your change, but you should delete the private prng variable and change the call in “nextValue()” to be “nextValue = person.random.nextGaussian() * standardDeviation + mean;”
•	After reading the systolic blood pressure in calculateCardioRisk(), you should immediately read the diastolic blood pressure, even if you don’t use it. Just call the function “person.getVitalSign(VitalSign.DIASTOLIC_BLOOD_PRESSURE, time);”
o	This will ensure the systolic and diastolic trending value generators are created at the same time, which is exactly what you want to happen.

By setting the "blood_pressure_controlled" flag to true immediatedly after the medication order is issued,  and then have the new blood pressure reading occurring a month or more later at the follow up encounter resulted in corrected blood pressure readings.  The one month or more delay removes the effect of averaging prior readings stored in the ring buffers using by trend generator. 

The hypertension section of the Metabolic Syndrome Standard of Care module has been removed and superceded by the stand alone Hypertension module.


jawalonoski:
@petercyli Thank you for another pull request!

shabiel:
Thank you for a great contribution @petercyli.

PR #557, opened 2019-08-20
Support running from a JAR file

dehall:
Updates the module loading process to support running Synthea from a JAR file (for instance, eclipse IDE supports exporting a project and re-packaging all dependencies into a single JAR). All other resources seem to load fine, only modules needed to be handled differently.  Without these changes, if you try to package synthea into a JAR and run it, you get various exceptions trying to load the files.

This is especially useful for running Synthea in a cluster environment; gradle lockfiles prevent multiple instances from running against the same copy at once, and running the classes directly requires a lot of separate libraries and classpath setup.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Module.java
+++ b/src/main/java/org/mitre/synthea/engine/Module.java
@@ -3,14 +3,15 @@ package org.mitre.synthea.engine;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
-import com.google.gson.stream.JsonReader;
 
 import java.io.File;
-import java.io.FileReader;
-import java.net.URL;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.FileSystemNotFoundException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.spi.FileSystemProvider;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


jawalonoski:
Is there a way to generate the jar from Gradle, rather than eclipse? Even though I personally use eclipse, I do not want to require any specific IDE for any functions.

If the `./gradlew jar` command suffices, this probably warrants an addition to the README and/or Wiki... and we may want to consider adding the javadoc pages to our `gh-pages` branch.

Thoughts on that?

dehall:
Eclipse makes it easy by allowing for repackaging all libraries into the new jar (aka a "fat jar" or "uber-jar"). Gradle doesn't have that out of the box but we could create a new task to build a fat jar.

Alternatively we can get a little creative with the `jar` and `install` tasks to get all the pieces we need. `install` copies all the dependencies onto ./build/install/synthea/lib and `jar` creates the basic jar as ./build/libs/synthea.jar. Then run the App class with the synthea and library jars provided as classpath:

```bash
$ ./gradlew install jar
[...]
$ java -cp "build/libs/synthea.jar:build/install/synthea/lib/*" App -p 0
Scanned 60 modules and 36 submodules.
[synthea runs...]
```
(on Windows it should use a `;` instead of `:` in the classpath list there)

I don't see any harm in putting the javadocs on gh-pages, but I would expect most usage of synthea as a library to focus on the Generator class, so if it's easier to put a brief description of just that class on the wiki that might be enough.

jawalonoski:
Well, if you use one of the `Generator.generatePerson(...)` methods, they both return `Person` objects, so the can of worms is opened, and you might as well have Javadoc for everything. But we can put that aside for another issue to be worked, if you prefer.

dehall:
Yeah that's true. Probably easier to just add a few more lines to the travis build for javadocs anyway. I'm happy to add that in this PR if you'd like or defer that to later.

jawalonoski:
We can defer javadocs until later.

Issue #555, opened 2019-08-19
Blood pressures should return as integers rather than floats


shabiel:
I have not seen any sphygmomanometer that can produce blood pressures such as these:

```
jq '.entry[].resource | select(.code.coding) | select (.code.coding[0].code == "55284-4") | { sys: .component[0].valueQuantity.value, dia: .component[1].valueQuantity.value } ' < output/fhir/Alfonso758_Gálvez271_0270ece3-d93e-49e5-abfa-44d5cf30e988.json 
{
  "sys": 88.29144415948353,
  "dia": 135.23018803524187
}
{
  "sys": 85.38462369227672,
  "dia": 113.6810154467861
}
{
  "sys": 81.13054817674676,
  "dia": 114.95844871239585
}
{
  "sys": 83.45264439371225,
  "dia": 118.82586622254128
}
{
  "sys": 72.77896457826508,
  "dia": 121.73250910823609
}
{
  "sys": 85.14656872532069,
  "dia": 123.21180424227781
}
{
  "sys": 81.30910581978247,
  "dia": 118.51571585691588
}
{
  "sys": 75.86713841285172,
  "dia": 134.6475547640152
}
{
  "sys": 77.74549782225387,
  "dia": 101.45816752987017
}
{
  "sys": 84.91786256626268,
  "dia": 128.25111524136327
}
```

A realistic blood pressure would be something like 128/84 with no decimal point.

I would like to help fix this, if I know where to look. Looks like `src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java` is the right file to look in.

jawalonoski:
I think it is probably the exporters you want to modify, and have them output integers (or truncate the decimals to a few places), rather than messing with the internal data representation.

shabiel:
That unfortunately means you have to change 6 files (maybe more) rather than one. Are you sure?

As far as internal data--I agree we need to retain the decimals until we have the final blood pressure.

jawalonoski:
Well, the alternative is to put special logic into `State.Observation.process(...)` for different types of vital signs and convert to integer when a `new HealthRecord.Observation(...)` is created.

So, add some `if` or `switch` logic inside the block on line 1174 to convert to an `Integer` (it has to be an Object type and not a primitive) when it is a Blood Pressure reading.

The issue with this approach is that we would be putting special logic into a relatively generic class.

[...]

PR #560, opened 2019-08-22
WIP: Fix decimal points on vitals to be realistic

shabiel:
First attempt at #555, and my first Java code in a few years...

This PR doesn't just address Blood Pressures, it also addresses Height, Weight, BMI. It's very preliminary to see if this is the way I should go about changing things.

I didn't address pain yet as I couldn't figure out how to identify an observation is a Pain observation.

The FHIR R4 output generated by this comes out good for BP, Ht, Wt, BMI. BP are exported as Integers/Long, Ht, Wt, BMI as Big Decimals with 1 or 2 decimal points.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ src/main/resources/version.txt
 # VS Code launch settings
 .vscode/launch.test.json
 .vscode/launch.json
+/build/
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -1148,6 +1148,8 @@ public abstract class State implements Cloneable {
     public boolean process(Person person, long time) {
       String primaryCode = codes.get(0).code;
       Object value = null;
+      Double intermediateVital;
+
       if (exact != null) {
         value = exact.quantity;
       } else if (range != null) {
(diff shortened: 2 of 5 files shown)


jawalonoski:
I'm not sure why the Travis build failed, so I'm going to restart it.

@shabiel, please make sure you eventually address the checkStyle complaints when you are ready.

You can ignore the missing Javadoc messages like:
```
Missing a Javadoc comment. [JavadocMethod]
```
But these type of warnings should be addressed... attempting to keep the style relatively consistent.
```
Line is longer than 100 characters (found 103). [LineLength]
WhitespaceAround: 'if' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
WhitespaceAround: '{' is not preceded with whitespace. [WhitespaceAround]
'vitalSign' has incorrect indentation level 10, expected level should be 12. [Indentation]
'}' at column 9 should be on the same line as the next part of a multi-block statement (one that directly contains multiple blocks: if/else-if/else, do/while or try/catch/finally). [RightCurlySame]
'if' construct must use '{}'s. [NeedBraces]
'method def modifier' has incorrect identation level 0, expected level should be 2. [Indentation]
Line contains a tab character. [FileTabCharacter]
'||' should be on a new line. [OperatorWrap]
```

shabiel:
I didn't know about the style check (I just saw it this morning as
well in the Travis build). I will address them. Thank you.

When you ready, can you let me know if you think my approach is
reasonable? The code is kinda messy and there are dead paths right now
(which I found after I pushed it yesterday); but I wanted to know if
this looks like a good direction.

--Sam

jawalonoski:
It looks like StateTest is the culprit... should `120L` (as a Long) and not `120` (Integer) for some reason.

shabiel:
Yes. Math.round(double) returns Long. And HAPI FHIR classes include a
Quantity constructor that takes Long, so in the end I kept Blood
Pressures as Long. I tried to cast it to Integer but the code didn't
compile (because of a warning on loss of precision)--and I was running
out of time... should have fixed it if I had know that it will affect
Travis.

jawalonoski:
👍 Thank you for this inspiring PR, but it is now overcome by events now with #573.

PR #561, opened 2019-08-27
Enhanced Imaging Studies - Random number of Series and Instances

JamesClementsMS:
This PR enhances Synthea's Imaging Study outputs through the following changes:
1. Adds options to uniformly randomly pick the number of series and instances within an ImagingStudy. Random DICOM UIDs will also be added to them.
2. Provides an example of this by modifying the Lung Cancer module.

In more detail:
1. Within the ImagingStudy state inside modules, you can provide the minimum and maximum (inclusive) number of series the ImagingStudy may contain. For this you need to provide at least one series as a reference which will then be duplicated.
Similarily within a series you may provide the minimum and maximum number of instances the series may contain, along with a reference instance.
For example, a ImagingStudy state in JSON format:
```
"State Name": {
  "type": "ImagingStudy",
  "procedure_code": { ... },
  "min_number_series": 1,
  "max_number_series": 3,
  "series": [
    {
      "body_site": { ... },
      "modality": { ... },
      "min_number_instances": 300,
      "max_number_instances": 500,
      "instances": [
        {
          "title": "CT Image Storage",
          "sop_class": { ... }
        }
      ]
    }
  ],
  "direct_transition": "Next State Name"
}
```
When processing the state, the number of series/instances is uniformly randomly picked within the range, and the reference series/instance is duplicated with new UIDs being assigned.

Note: This is backwards compatible. If the new fields are not included in the state, or have invalid values, the provided series/instances will be exported without any changes.

2. In this PR only the Lung Cancer module has been changed. A generated CT ImagingStudy will contain a single series with between 300-500 instances. Most other modules containing imaging studies are for scans producing a single instance (e.g. 2D X-Ray), so they do not need to be changed.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -3,7 +3,9 @@ package org.mitre.synthea.engine;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
 import org.mitre.synthea.engine.Components.Exact;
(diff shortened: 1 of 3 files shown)


JamesClementsMS:
> 
> 
> Looks good. One minor change requested, please.
> 
Thanks! I've made the necessary changes

PR #565, opened 2019-09-05
Updating lat/long in zipcodes.csv

eedrummer:
The current zipcodes.csv file contains the same or similar latitudes and longitudes for cities with multiple zip codes. As examples, New York City, Denver and San Francisco all have a single lat/long assigned to all of their zip codes. Boston has two separate values.

This PR updates the file based on information found at: https://gist.github.com/erichurst/7882666

The claimed source of the data is the US Census Bureau.

The zipcodes.csv file was modified by looking for all entries with zipcodes. If there was a corresponding zip code in the Census data, the lat/long data was used to replace the existing data. Entries in the file without a zipcode we left untouched.

Changes resulted in an average shift of 3.05 miles in points that changed in the file. The minimum was zero and the maximum was 298.23 miles for a point in Anchorage. That point is probably a mistake, but on the whole, the data appears to be less wrong that it was before.

A particular example of this is San Francisco. Currently, all zipcodes in SF are located on the Farallon Islands:
Screen Shot 2019-09-05 at 3 01 04 PM

Updated data seems much more realistic:
Screen Shot 2019-09-05 at 2 59 41 PM

Other updated examples:

Boston Before:
Screen Shot 2019-09-05 at 3 08 31 PM

Boston After:
Screen Shot 2019-09-05 at 3 05 33 PM

NYC Before:
Screen Shot 2019-09-05 at 3 11 32 PM

NYC After:
Screen Shot 2019-09-05 at 3 10 49 PM

Denver Before:
Screen Shot 2019-09-05 at 3 03 40 PM

Denver After:
Screen Shot 2019-09-05 at 3 02 16 PM


mrnosal:
The new file is improved compared to the old, however, some problems remain:

The old zipcodes.csv had 91560 points.
The new zipcodes.csv has 87114 points.
There are 4446 fewer points in the new file compared to the old file.

The old zipcodes.csv had 31403 unique ZCTA5 entries.
There were 60157 duplicate ZCTAs in the old file.
There were 27433 entries without a ZCTA5 value (no zip code).
There were 64127 unique ZCTA5 values.

The new zipcodes.csv has 27541 unique ZCTA5 entries.
There are 59573 duplicate ZCTAs in the new file.
There are 36650 entries without a ZCTA5 value (no zip code).
There are 50464 unique ZCTA5 values.

By comparison, the Census Tiger Line ZCTA shapefile for 2017 includes 31444 unique ZCTA5 entries.

PR #566, opened 2019-09-06
Module overrides

dehall:
Starting point for "module overrides". The idea here is that we can provide a module override file at runtime that defines parameters that override parts of the generic modules. There are a number of possible use cases for this, for example pre-defining alternative populations (think "internationalization", such as "run with this override file to produce a population matching all condition prevalence rates in Germany"). Also includes a task that can generate a base set of override files with some configurability.

The format of the overrides file is a java properties file, with format:
```
(module file path)::(jsonpath to value) = (new value)
```

example:
```
congestive_heart_failure.json\:\:$['states']['Age_70-79-Chance\ of\ CHF']['distributed_transition'][0]['distribution'] = 0.024
congestive_heart_failure.json\:\:$['states']['Age_80-Chance\ of\ CHF']['distributed_transition'][0]['distribution'] = 0.045
atopy.json\:\:$['states']['Initial']['complex_transition'][0]['distributions'][0]['distribution'] = 0.917
atopy.json\:\:$['states']['Initial']['complex_transition'][1]['distributions'][0]['distribution'] = 0.033
```

The way that this feature is used is by setting a config property:
```shell
run_synthea --module_override=path_to_properties_file
```

Note that I also extracted the common "for each module" logic into a new utility method since the exact same boilerplate was used in 5+ places. Apologies for the extra diffs but if you view it in "ignore whitespace mode" it's more obvious what the actual change to those files was.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -162,6 +162,22 @@ task attributes(type: JavaExec) {
   main = "org.mitre.synthea.helpers.Attributes"
 }
 
+task overrides(type: JavaExec) {
+  group 'Application'
+  description 'Create a list of modules parameters in module override format'
+  classpath sourceSets.main.runtimeClasspath
+  main = "org.mitre.synthea.helpers.ModuleOverrides"
+  doFirst {
+    // TODO: is there a way to make this cleaner?
+    String includeFields = project.hasProperty('includeFields') ? project.getProperty('includeFields') : ""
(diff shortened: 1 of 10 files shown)


dehall:
Note to self: once this is eventually complete and merged, update the README on the populations repo: https://github.com/synthetichealth/populations

dehall:
Once this is merged I'll add a wiki page for it as well

PR #582, opened 2019-09-23
Generator client support

swohlever:
This collection of branches includes:
- Support for exporting JSON FHIR records to a thread-safe queue
- A per-generator "years of history" setting
- Per-generator sub-directory for CSV exports
- Use of an alternative class loader when loading modules (to fix an issue with using the Synthea library from within a web application deployed in Tomcat)


shabiel:
Can you tell us more about integrating Synthea with Tomcat? This is
something we (OSEHRA) are interested in. We are currently using a command
line wrapper, which is not ideal.

"to fix an issue with using the Synthea library from within a web
application deployed in Tomcat"

swohlever:
The Tomcat-related update in this request fixes a runtime error I ran into while using the Synthea lib in a Spring-based application in Tomcat. During the module loading phase, the Tomcat runtime did not like the original class loader. Online research led me to a solution: change out the loader. See the modification to Module.java in this pull request for the details.

The addition of the thread-safe queue in the generator object is the major component of this pull request. One can create and start a generator instance in a thread, e.g, created in a server in response to a RESTful or Web Socket request from a client, and then make the generated records available to the client via polling, streaming, file downloads, etc. The CSV file downloads are supported by another change in this pull request that allows a server to create a CSV subdirectory to collect CSV exports for a particular generator.

Whatever web service you create needs to do the heavy lifting related to handling client requests, managing the generator threads, and serving up the results, but that is similar to what any other such service needs to do. The changes to Synthea in this pull request are not extensive but they facilitate creating such a service, wrapped around the Synthea lib.

shabiel:
Thank you.

Can you share code to show how to integrate the Synthea "lib" with the
spring framework? That code is interesting in itself.

--Sam

swohlever:
> 
> 
> The changes to support runtime specification of CSV exporter options (sub directory and years of history) special-case CSVExporter in the Generator class. I think it would be cleaner to use the same pattern as GeneratorOptions to create an ExporterOptions class and then add a constructor to Generator that allows a client to pass in an ExporterOptions instance that is then passed down to Exporter. This will allow runtime customization of other exporters in the future and will reduce the coupling of Generator to any particular exporter.

Yes, the CSV exporter change was a bit app-specific. I wavered on including it in the pull request. It is not a central component of this update so that particular commit could be omitted from the merge.

swohlever:
> 
> Thank you. Can you share code to show how to integrate the Synthea "lib" with the spring framework? That code is interesting in itself.
> […](#)
> --Sam
> On Fri, Oct 4, 2019 at 5:59 AM swohlever ***@***.***> wrote: The Tomcat-related update in this request fixes a runtime error I ran into while using the Synthea lib in a Spring-based application in Tomcat. During the module loading phase, the Tomcat runtime did not like the original class loader. Online research led me to a solution: change out the loader. See the modification to Module.java in this pull request for the details. The addition of the thread-safe queue in the generator object is the major component of this pull request. One can create and start a generator instance in a thread, e.g, created in a server in response to a RESTful or Web Socket request from a client, and then make the generated records available to the client via polling, streaming, file downloads, etc. The CSV file downloads are supported by another change in this pull request that allows a server to create a CSV subdirectory to collect CSV exports for a particular generator. Whatever web service you create needs to do the heavy lifting related to handling client requests, managing the generator threads, and serving up the results, but that is similar to what any other such service needs to do. The changes to Synthea in this pull request are not extensive but they facilitate creating such a service, wrapped around the Synthea lib. — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or mute the thread.

Not at this time. However, using Synthea with Spring was similar to using other third party libs. With the changes in this pull request, a Generator object can be instantiated, configured, and run inside a thread. Another thread can be created to watch for results from the generator's thread-safe queue (which was added to the generator in this pull request).

hadleynet:
OK, if you want to pull the "Added yearsOfHistory generator option" and "Generator-specific CVS exports" commits from the pull request then I can merge that and then I'll refactor the Exporter to support runtime config along the lines I suggested.

swohlever:
> 
> 
> OK, if you want to pull the "Added yearsOfHistory generator option" and "Generator-specific CVS exports" commits from the pull request then I can merge that and then I'll refactor the Exporter to support runtime config along the lines I suggested.

I reverted those two commits.

PR #588, opened 2019-10-07
Synthea as a Library

hadleynet:
This set of commits aims to make Synthea more useable as a library. Together they provide:

- Support for running Synthea from a jar file via java -jar
- Support for building an uber jar with all dependencies bundled
- Support for overriding the bundled configuration file using a local config file to override a subset of settings
- Support for adding modules from a local directory
- Some generalization of the approach from pull request #582 for using a queue to exchange serialized records between threads, e.g. serving generated records from a servlet.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -113,6 +113,22 @@ task graphviz(type: JavaExec) {
   main = "Graphviz"
 }
 
+task uberJar(type: Jar) {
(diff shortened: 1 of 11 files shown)


jawalonoski:
Please add appropriate documentation in the wiki pages.

PR #589, opened 2019-10-08
Physiology

nanowizard:
I know this is a bit late. Noticed an issue with concurrency and wanted to implement a way for `PhysiologyValueGenerator`s to use other generators until there is a sufficient change in input parameters to warrant running the simulation.

SBML model files are placed in `src/main/resources/physiology/models` and generator configurations are in `src/main/resources/physiology/generators`.

The basic synthetic ECG model is there, but it is not currently being utilized since there aren't any corresponding `VitalSign`s for it.

Configurations for running the `physiology` gradle task are currently in `config/simulations`. There is also a README file there which gives a brief explanation of the task and its usage.

I spent some time tweaking the input parameters for the `cardiac_hemodynamics` generator. I based these inputs on a 2015 study providing statistics relating age and BMI to blood pressure (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4553889/). The general trend of BP increase with age  has been shown to be largely due to increased peripheral resistance and increased arterial stiffness. Therefore I developed equations for these two parameters that give results which are close to the statistics presented in the study. Increased BMI is also associated with an increase in BP, so I accounted for this by making arterial resistance related to the deviation of BMI from the center of the ideal BMI range for adults (21.7). From the couple of tests I've performed, Synthea's results appear to be fairly consistent with empirical results.

I'd also like to account for gender in these equations but `ExpressionProcessor`s don't yet support sending non-numeric parameters into expressions.


src/test/java/org/mitre/synthea/engine/StateTest.java
```diff
@@ -1649,4 +1649,42 @@ public void testMultiObservation() throws Exception {
...
+    
+    // The "Final Aortal Volume" attribute should have been set
+    assertTrue(person.attributes.containsKey("Final Aortal Volume"));
+    
+    // The "Arterial Pressure Values" attribute should have been set to a list
+    assertTrue(person.attributes.get("Arterial Pressure Values") instanceof List);
+    
+    //System.out.println("LVEF: " + person.getVitalSign(VitalSign.LVEF, time));
+    //System.out.println("SYS: " + person.getVitalSign(VitalSign.SYSTOLIC_BLOOD_PRESSURE, time));
+    //System.out.println("DIA: " + person.getVitalSign(VitalSign.DIASTOLIC_BLOOD_PRESSURE, time));
```
jawalonoski: You should delete commented out code.

src/main/java/org/mitre/synthea/world/agents/Person.java
```diff
@@ -240,6 +241,24 @@ public Period age(long time) {
...
+    double age = 0.0;
+
+    if (attributes.containsKey(BIRTHDATE)) {
+      LocalDate now = Instant.ofEpochMilli(time).atZone(ZoneId.systemDefault()).toLocalDate();
+      LocalDate birthdate = Instant.ofEpochMilli((long) attributes.get(BIRTHDATE))
+          .atZone(ZoneId.systemDefault()).toLocalDate();
+      age = birthdate.until(now, IsoFields.QUARTER_YEARS) / 4.0;
+    }
+    return age;
+  }
```
jawalonoski: It seems like you are unnecessarily copying code... couldn't you just do something like:

```java
double years = (double) age(time).toTotalMonths() / 12.0;
if (years < 0) {
  years = 0;
}
return years;
```

src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -249,10 +252,21 @@ public static void birth(Person person, long time) {
    * @param person The person to generate vital signs for.
    */
   private static void setupVitalSignGenerators(Person person) {
+    boolean usePhysiology = Boolean.parseBoolean(Config.get("physiology.enabled", "false"));
```
jawalonoski: Typically, we are setting these type of variables once from the Config file at the class level. See `appendNumbersToNames` in the same file.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -252,6 +263,311 @@ public boolean process(Person person, long time) {
       }
     }
   }
+  
+  /**
+   * The CallSubmodule state immediately processes a reusable series of states contained in a
+   * submodule. These states are processes in the same time step, starting with the submodule's
+   * Initial state. Once the submodule's Terminal state is reached, execution of the calling module
+   * resumes.
+   */
```
jawalonoski: This Javadoc comment is copy-pasta.

src/main/java/org/mitre/synthea/helpers/ExpressionProcessor.java
```diff
@@ -0,0 +1,284 @@
...
+   * @param paramTypeMap Map of parameter names to their corresponding CQL types.
+   */
+  public ExpressionProcessor(String expression, Map<String,String> paramTypeMap) {
+    this.cqlParamMap = HashBiMap.create();
+    this.paramTypeMap = paramTypeMap;
+    
+    String cleanExpression = replaceParameters(expression);
+    String wrappedExpression = convertParameterizedExpressionToCql(cleanExpression);
+    // System.out.println("Wrapped Expression:");
+    // System.out.println(wrappedExpression);
```
jawalonoski: Unnecessary commented out prints... remove.

build.gradle
```diff
@@ -48,10 +48,15 @@ dependencies {
...
   compile group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-csv', version: '2.8.8'
-  compile group: 'org.yaml', name: 'snakeyaml', version: '1.19'
+  compile group: 'org.yaml', name: 'snakeyaml', version: '1.25'
   compile group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
   compile group: 'org.apache.commons', name: 'commons-text', version: '1.2'
 
+  compile group: 'org.opencds.cqf', name: 'cql-engine', version: '1.2.20'
+  compile group: 'info.cqframework', name: 'cql', version: '1.2.20'
+  compile group: 'info.cqframework', name: 'model', version: '1.2.20'
+  compile group: 'info.cqframework', name: 'cql-to-elm', version: '1.2.20'
```
jawalonoski: These CQL library versions are out of date by a year.... update to newer version?

build.gradle
```diff
@@ -60,7 +65,15 @@ dependencies {
   // get rid of SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
   // if we switch to a real logging framework we may want to switch this
   compile "org.slf4j:slf4j-api:1.6.1"
-  compile "org.slf4j:slf4j-nop:1.6.1"
+  // compile "org.slf4j:slf4j-nop:1.6.1" // SLF4J seems to already be provided by org.apache.logging.log4j
+  
+  // Physiology simulation dependencies
+  compile files('lib/SimulationCoreLibrary_v1.5_slim.jar')
+  compile group: 'org.sbml.jsbml', name: 'jsbml', version: '1.4'
```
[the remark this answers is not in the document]
nanowizard: Ah yes I remember I tried that initially but I get the following error:

```Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find org.simulator:sbscl:1.5.
     Searched in the following locations:
       - https://jcenter.bintray.com/org/simulator/sbscl/1.5/sbscl-1.5.pom
       - https://jcenter.bintray.com/org/simulator/sbscl/1.5/sbscl-1.5.jar
       - https://repo.maven.apache.org/maven2/org/simulator/sbscl/1.5/sbscl-1.5.pom
       - https://repo.maven.apache.org/maven2/org/simulator/sbscl/1.5/sbscl-1.5.jar
       - https://oss.sonatype.org/content/repositories/snapshots/org/simulator/sbscl/1.5/sbscl-1.5.pom
       - https://oss.sonatype.org/content/repositories/snapshots/org/simulator/sbscl/1.5/sbscl-1.5.jar

Doing a quick search, 'sbscl' doesn't appear to show up in the central repositories.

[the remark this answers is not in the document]
Comment (jawalonoski) [reply build.gradle]: You are correct. I can't find anyway around that.

While poking around, I did notice that the SBSCL library is licensed under LGPL v3.0. https://draeger-lab.github.io/SBSCL/licenses.html

You should probably create a `./lib/sbscl/` folder, where you include the jar along with a copy of the LGPL License (as required by that license, https://www.gnu.org/licenses/lgpl-3.0.txt) and then add a statement in the `NOTICE` file (also required my interpretation of the LGPL).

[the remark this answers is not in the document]
Comment (nanowizard) [reply build.gradle]: Yes good points, I'll do that.

File: build.gradle (hunk 6 above)
Comment (nanowizard) [reply build.gradle]: Will do, thanks Chris!

Comment (brynrhodes) [reply build.gradle]: We have had some packaging issues with the 1.3.9 release, can you use the 1.3.10-SNAPSHOT?

File: build.gradle (hunk 7 above)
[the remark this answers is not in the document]
Comment (nanowizard) [reply build.gradle]: How is this for addition to the NOTICE? 

```
This material includes the Systems Biology Simulation Core Library (SBSCL) as an
unmodified, publicly released version of the software in the form of a Java archive
(JAR) file and makes use of its interfaces under provision of the GNU Lesser
General Public License v3 (https://www.gnu.org/licenses/lgpl-3.0.txt). Releases of
the SBSCL library can be found at https://github.com/draeger-lab/SBSCL/releases/.
```

I'm certainly no expert at legalese but I attempted to explain as clearly as possible exactly how we're using the library and how it falls under the LGPL. I welcome any suggestions for improvement.

File: build.gradle (hunk 6 above)
Comment (nanowizard) [reply build.gradle]: Hi Bryn. Sorry I just saw your comment. Using 1.3.10-SNAPSHOT appears to work, but we get a ton of DEBUG messages in the output.

Comment (jawalonoski) [reply build.gradle]: Try to override the log4j settings.

Comment (jawalonoski) [reply build.gradle]: For example, maybe these suggestions might work: https://stackoverflow.com/questions/4258849/override-log4j-properties

**Comment by nanowizard** [comment]
> * Did you look at the checkstyle output? Does it complain about anything other than "missing javadoc comments"?

Sure it complains about a lot of things, but none of them are against files I've changed. Looks like the classes that have checkstyle issues (besides missing javadoc comments) are:

- org.mitre.synthea.datastore.Datastore
- org.mitre.synthea.engine.Generator
- org.mitre.synthea.export.Exporter

> * Codecov is showing that test coverage of the submitted code is low (only covers 66% of your change). Is there an opportunity to include more unit tests and increase coverage?

Working on this.

> * It seems like you are including/duplicating some of the SBML YML and XML files in both `src/main/resources/physiology` and `src/test/resources`... are you sure you need to duplicate them (you probably shouldn't)? Or are there subtle differences between the files that I'm not seeing?

I duplicated them since I knew these files will probably change in the future (model defaults changes, better input arguments, etc.) and figured we may not want to have to change the tests every time we tweak a parameter. But on second thought, perhaps we *do* want that. If those configuration files change we do want to make sure our tests ensure that the changed files work appropriately.

I'll remove the duplicates.

File: build.gradle (hunk 6 above)
Comment (nanowizard) [reply build.gradle]: I went ahead overriding the logging with log4j.xml

**Comment by nanowizard** [comment]
@jawalonoski I think I've addressed all of the items you mentioned now and increased the code coverage to about 82%. I also noticed I had duplicated a method for getting attribute and `VitalSign` values from `Person` objects so I collapsed them into a single static method on the `ExpressionProcessor` class (commit e603410).

Let me know if you have any other concerns.

**Comment by jawalonoski** [comment]
I am playing around with these changes before merging them, and I'm experiencing some `java.lang.OutOfMemoryError` exceptions when running `./run_synthea -s 0 -p 10000 California`.

I'll want to figure out what is going on here before merging that in.

Also, `physiology.enabled` should default to `false`.

**Comment by nanowizard** [comment]
> I am playing around with these changes before merging them, and I'm experiencing some `java.lang.OutOfMemoryError` exceptions when running `./run_synthea -s 0 -p 10000 California`.
> 
> I'll want to figure out what is going on here before merging that in.

I profiled the code with and without physiology enabled and determined what the main problem was. I was keeping a static map of `SimRunner` instances so that multiple `PhysiologyValueGenerator`s which use the same model would also use the same simulator for each `Person` object. Since that `Map` was `static`, those `SimRunners` would stick around forever, long after the `Person` had completed their life cycle, which resulted in a memory leak.

I rectified this by using a `transient Map` of `SimRunner`s on each `Person` object to keep track of those instead of the `static Map` on the `PhysiologyValueGenerator` class (`transient` so that it won't be serialized with the `Person`). This means when the `Person` goes out of scope, it's corresponding `SimRunner` instances are also garbage collected. The downside to this is it somewhat increases coupling between the `Person` class and the `PhysiologyValueGenerator` class. I'm open to other suggestions if you think there's a better way to ensure those objects are cleaned up properly.

> Also, `physiology.enabled` should default to `false`.

Should be fixed now. That was a mistake on my part.

**Comment by nanowizard** [comment]
Memory Profiling Results:

`master` branch (no physiology simulations):
image

`physiology` branch (with physiology enabled):
image

Memory usage is pretty comparable, though you'll notice there are many more spikes in memory usage on the physiology branch. This is likely due to the temporary simulation data results tables being populated, since they can be quite large.

These were both runs generating a population of 10000. Note that the `master` run executed in about 9 minutes, while the `physiology` run executed in a bit over 13 minutes, suggesting a run time cost of roughly 1.5x for enabling physiology with the current generator settings. This can be tweaked pretty effectively by modifying the input parameter `variance` field in the generator configuration, which causes a trade-off between performance cost and physiological accuracy.

**Comment by dehall** [comment]
I'm very interested in this feature and can take a look tomorrow.

**Comment by jawalonoski** [comment]
During profiling, I noticed two heavy CPU hot spots: calculating Haversine distance for locations, and the use of `java.lang.Enum.valueOf(String)`.

The first is out of scope on this pull request. But for the second hot spot, over 18% of the "Self Time (CPU)" time is spent on `Enum.valueOf` when physiology is enabled.

If you can change `ExpressionProcessor` to precompute whether or not each `param` is a `VitalSign` or an `Attribute` then you won't have to call `VitalSign.fromString` for every single call to `ExpressionProcessor.getPersonValue(...)`.

That might considerably speed up the expression processing, and in turn the entire physiology code. Worth a try.

**Comment by dehall** [comment]
This looks awesome. Just two questions around the `physiology.enabled` config setting:
1) It only affects the PhysiologyValueGenerators that produce the values for vital signs, and not the Physiology state in modules, correct? Could the name be a little more specific to reflect that?
2) What happens if the config setting is disabled, but something else depends on a vital sign that a generator was going to set? Is the idea that we're always going to need a "quick & dirty" version to generate values for that vital sign alongside the more robust but slow model?
(And apologies if these were already discussed; I'm not sure if I'm seeing all the comments on the PR based on the number of emails I had when I got back from vacation a couple weeks ago)

File: config/simulations/README.md [hunk 8]
```diff
@@ -0,0 +1,15 @@
...
+The configuration allows manipulation of the differential equation solver, step
+size, sim duration, and configuration for output charts if desired. Output
+files will be in the `output/physiology` directory.
+
+Note that these are only utilized for the physiology gradle command and are not
+part of the normal Synthea execution procedure.
+
+## Usage
+
+`./gradlew physiology --args="configs/simulations/[config name].yml"`
```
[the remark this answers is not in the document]
nanowizard: Good catch! Thanks!

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -381,11 +505,17 @@ public SetAttribute clone() {
       SetAttribute clone = (SetAttribute) super.clone();
       clone.attribute = attribute;
       clone.value = value;
+      clone.expression = expression;
+      clone.expProcessor = expProcessor;
```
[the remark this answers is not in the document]
nanowizard: To be honest I don't remember... Will look into this.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1206,6 +1365,7 @@ public Observation clone() {
       clone.vitalSign = vitalSign;
       clone.category = category;
       clone.unit = unit;
+      clone.expression = expression;
```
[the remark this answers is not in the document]
nanowizard: Related to your previous comment. Will look into this. I don't believe `ExpressionProcessor` is thread safe so you might be correct.

In [`IoMapper`](src/main/java/org/mitre/synthea/helpers/physiology/IoMapper.java) I use a `ThreadLocal<ExpressionProcessor>` so we might want to do the same here. That would be more efficient than creating a clone for each individual `Person`.

src/main/java/org/mitre/synthea/helpers/Utilities.java
```diff
@@ -315,4 +315,65 @@ private static String randomDicomUidSalt() {
     int saltInt = rand.nextInt(MAX - MIN + 1) + MIN;
     return String.valueOf(saltInt);
   }
+  
+  /**
+   * Utility function to convert from a Java wrapper class to primitive type class.
+   * @param clazz type to convert to primitive type
+   * @return primitive class
+   */
+  public static Class<?> toPrimitiveClass(Class<?> clazz) {
```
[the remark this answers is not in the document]
nanowizard: Oh excellent! I wasn't aware of that function. I'll be happy to use that instead.

nanowizard:
In response to Jay's comment:

> If you can change `ExpressionProcessor` to precompute whether or not each `param` is a `VitalSign` or an `Attribute` then you won't have to call `VitalSign.fromString` for every single call to `ExpressionProcessor.getPersonValue(...)`.

Instead, perhaps I could just add another parameter to indicate whether the string points to a `VitalSign` or attribute, since that's known ahead of time in the physiology configurations. I'll try this out and let you know what the performance difference is.

I only took a superficial look at the CPU profile to make sure there weren't any glaring issues so I didn't dig deep enough  to see that the `VitalSign.fromString` method call could be significantly impacting performance.

In response to Dylan's comment:

> This looks awesome. Just two questions around the physiology.enabled config setting:
> 1. It only affects the PhysiologyValueGenerators that produce the values for vital signs, and not the Physiology state in modules, correct? Could the name be a little more specific to reflect that?

Yes that's a good point. Though I think there are multiple ways we could go about this. One is to rename the config to reflect that it only affects `PhysiologyValueGenerator`s as you suggest. Another option is we could add an additional required field for `Physiology` states to provide an alternative state to go to if `physiology.enabled = false` and have `Physiology` states avoid the simulation in that case.

Since we have `PhysiologyGenerator`s I don't see the `Physiology` state being used too much in modules but I could be wrong.

> 2. What happens if the config setting is disabled, but something else depends on a vital sign that a generator was going to set? Is the idea that we're always going to need a "quick & dirty" version to generate values for that vital sign alongside the more robust but slow model?

Yes I believe so. At least for now I think the idea is to always have a much faster "quick & dirty" backup generator for each `VitalSign` since physiology simulations are relatively computationally expensive.

nanowizard:
> During profiling, I noticed two heavy CPU hot spots: calculating Haversine distance for locations, and the use of `java.lang.Enum.valueOf(String)`.
> 
> The first is out of scope on this pull request. But for the second hot spot, over 18% of the "Self Time (CPU)" time is spent on `Enum.valueOf` when physiology is enabled.
> 
> If you can change `ExpressionProcessor` to precompute whether or not each `param` is a `VitalSign` or an `Attribute` then you won't have to call `VitalSign.fromString` for every single call to `ExpressionProcessor.getPersonValue(...)`.
> 
> That might considerably speed up the expression processing, and in turn the entire physiology code. Worth a try.

I implemented a `VitalSign` cache for the ExpressionProcessor so that it only ever has to call `VitalSign.fromString` once for each unique parameter it receives in `getPersonValue()`. Doing this did indeed have a noticeable impact on performance, reducing total run time on my 10000 patient generation from a bit over 13 minutes to about 12 minutes flat (about 10% faster).

nanowizard:
I don't understand why, but apparently a `Generator` test failed in the CI build:

```
org.mitre.synthea.engine.GeneratorTest > testGeneratorCreatesPeople FAILED
    java.lang.AssertionError: expected:<3> but was:<2>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at org.mitre.synthea.engine.GeneratorTest.testGeneratorCreatesPeople(GeneratorTest.java:34)
```

nanowizard:
> This looks awesome. Just two questions around the `physiology.enabled` config setting:
> 
> 1. It only affects the PhysiologyValueGenerators that produce the values for vital signs, and not the Physiology state in modules, correct? Could the name be a little more specific to reflect that?

I decided to rename the config setting to `physiology.generators.enabled` to be more specific. I figure if we want to be able to turn off `Physiology` state in modules as well that could simply be a separate configuration setting.

jawalonoski:
Thanks for the contribution! 🎉

PR #593, opened 2019-10-11
Updating the zip codes file - Version 3

eedrummer:
This is yet another pass at updating the zip code file. I tried to go about this attempt in a more methodical fashion. All work was done in a Jupyter Notebook using pandas. You can see the complete thing at:
https://nbviewer.jupyter.org/gist/eedrummer/8d226fe5f7d106f9594889df1075fcfb

Sources for checking information included:
US Census Zip Code Gazetteer - https://www2.census.gov/geo/docs/maps-data/data/gazetteer/2019_Gazetteer/2019_Gaz_zcta_national.zip

US Census Place Gazeteer - https://www2.census.gov/geo/docs/maps-data/data/gazetteer/2019_Gazetteer/2019_Gaz_place_national.zip

In a small number of cases, the Open Street Map Geocoding service.

Some information about the new file compared to the old file:
* The old file had 12 lines in it that were exact duplicates. They were removed.
* There were 17399 rows for cities and states that do not have a corresponding city and state in the demographic file. These have been removed as the Synthea code base has no path to access these rows.
* For the remaining entries with zip codes, the data looked pretty good compared to the place gazetteer.
* For entries without a zip code, there was a lot of data munging. I tried to focus on the most heavily populated cities to ensure correctness.  

Some thoughts for future work:
* There are issues in the demographics file. There are counties in the file. Some names are incorrectly truncated. There are several places in PA that should likely end in Township.
* The demographics file seems to split cities that are in multiple counties into separate entries, but that does not impact zip code look up. Handling this would require having a mapping to not just city/state, but city/county/state.
* This change throws out information, but not functionality. The old zip code file had many entries for cities that had odd suffixes which would prevent them from ever being selected. I was hesitant to go through and address all of these cases by hand.


jawalonoski:
First, your Jupyter Notebook is amazing. 🎉 

Second, I am terrible, because I didn't give you any code related to production of the original files, which consists of random scripts on my machine, @dehall machine, and who knows how many others.

I think that in both the demographics and zip files there were suffixes that were removed, because they are inconsistent across sources. Here is one code snippet I found.

```python
regexp = r'( Township)|( township)|( town)|( Town)|( city)|( village)|( Village)|( and borough)|( borough)|( Borough)|( Depot)|( municipality)|( census subarea)|( plantation)|( CCD)|( CDP)|( \(balance\))|( \(pt.\))|( zona urbana)|( zona comunidad)|( comunidad)'
```

This partially corresponds with this line you wrote:
```python
place_g['NAME'] = place_g['NAME'].str.replace('city', '').str.replace('CDP', '').str.replace('town', '').str.rstrip()
```

Additional testing that I did focused on trouble spots encountered in the past:
- Lexington, Kentucky
- Louisville, Kentucky
- Hartsville, South Carolina
- Indian Hill, Ohio (aka "The Village of Indian Hill")
- Nashville, Tennessee
- Bedford, Massachusetts

These changes properly generate each of these known "trouble" spots.

PR #597, opened 2019-10-22
Fixing lat/lon randomization

eedrummer:
Previous settings always added a value to the lat/lon. This will shift
the generated population to the north and east of the desired point.

This change will now allow the generation of negative values for the
random offset. This will generate a population in a circle around the
desired point.

### DIFF
--- a/src/main/java/org/mitre/synthea/world/geography/Location.java
+++ b/src/main/java/org/mitre/synthea/world/geography/Location.java
@@ -270,8 +270,8 @@ public class Location {
       // Precision within 0.001 degree is more or less a neighborhood or street.
       // Precision within 0.01 is a village or town
       // Precision within 0.1 is a large city
-      double dx = person.rand(0.001, 0.1);
-      double dy = person.rand(0.001, 0.1);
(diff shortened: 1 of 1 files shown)


src/main/java/org/mitre/synthea/world/geography/Location.java
```diff
@@ -270,8 +270,8 @@ public void assignPoint(Person person, String cityName) {
       // Precision within 0.001 degree is more or less a neighborhood or street.
       // Precision within 0.01 is a village or town
       // Precision within 0.1 is a large city
-      double dx = person.rand(0.001, 0.1);
-      double dy = person.rand(0.001, 0.1);
+      double dx = person.rand(-0.05, 0.05);
+      double dy = person.rand(-0.001, 0.05);
```
arscan: Is there a reason why `dy` isn't between -.05 and .05 like `dx`?

eedrummer: Ugh. Sorry. I was playing with things and didn't reset. I'll fix that.

PR #600, opened 2019-10-23
Use Path.relativize instead of String substitution

hadleynet:


dehall:
Tested in Windows, now it works in the uberJar but not when running via `run_synthea` ... looking into it

PR #602, opened 2019-10-25
Accelerate geographic searches.

jawalonoski:
## What?
This pull request removes the `org.apache.sis` library dependency and replaces it with a very simple and fast QuadTree class.

## Why?
While reviewing Pull Request #589 I have been conducting a lot of profiling using `jvisualvm`.

The single largest consumption of CPU time has been calls to `java.lang.Math.acos()` required to compute the Haversine distance while querying the QuadTree from the Apache SIS library. These are queries such as the "find nearest provider" query. According to the profiler, this was taking 28 - 30% of the computation time for each entire run of Synthea.

As an experiment, I replaced the Apache SIS QuadTree which uses Haversine distance calculations (which take into account the curvature of the Earth) with a very simple QuadTree that merely considers Euclidean distance on a flat 2D plane.

Considering the required degree of accuracy of this simulation, Euclidean distance is sufficient.

## Test and Results
I ran `./run_synthea -s 0 -p 10000` for `California`, `Alaska`, and `Texas` (large locations) and saw an average of a 30% reduction in overall time. `Alaska` saw the biggest improvement, which went from 10 minutes down to 4.5 minutes on my machine. This is probably because Alaska, and especially the Aleutian Islands, are so spread out and benefited the most from the simplification.

I also ran some seeded runs in Massachusetts before and after the change to determine if the change altered the locations that patients visited for their encounters. I conducted several spot checks and did not find any discrepancies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,9 +38,6 @@ dependencies {
   compile 'ca.uhn.hapi.fhir:hapi-fhir-structures-r4:4.0.3'
   // C-CDA export uses Apache FreeMarker templates
   compile 'org.freemarker:freemarker:2.3.26-incubating'
-  // location processing
-  compile group: 'org.apache.sis.core', name: 'sis-referencing', version: '0.8'
-  compile group: 'org.apache.sis.storage', name: 'sis-storage', version: '0.8'
   compile group: 'com.h2database', name: 'h2', version: '1.4.196'
   // google guava for some data structures
   compile 'com.google.guava:guava:23.0'
--- a/src/main/java/org/mitre/synthea/export/CDWExporter.java
+++ b/src/main/java/org/mitre/synthea/export/CDWExporter.java
@@ -4,6 +4,7 @@ import static org.mitre.synthea.export.ExportHelper.iso8601Timestamp;
 
 import com.google.gson.JsonObject;
 
+import java.awt.geom.Point2D;
 import java.io.File;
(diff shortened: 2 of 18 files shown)


shabiel:
This looks like rocket science to me!

eedrummer:
@jawalonoski I think this looks good, but I just want to make sure that I understand how it works. When searching by degrees, say 2 degrees, if I were to take Boston and use a lat / lon of 42.3, -71, the code will use a "circle" with the northern most point of the circle being at 44.3, -71 and a western most point at 42.3, -73. Is that how it works?

I say "circle" because you're flattening out the earth, so the closer you get to the poles, the less circular the circle will be... If I'm understanding this.

jawalonoski:
@eedrummer Yes, that is how it works.

And yes, you are understanding this. "Circle on a flat plain" is how it works, avoiding curvature of the Earth... Avoiding calculating curvature of the Earth (e.g. by using Haversine distance) is what is increasing the performance so much. Given there aren't a ton of cities at the poles, I think the level of accuracy is sufficient.

Also, accuracy doesn't seem to be too important in my mind anyway, to be honest. The primary purpose of this data structure is to remove hospitals from the search space to the immediate area, and to find the closest emergency room, for example. Caveated of course, with the curvature of the Earth issue, and the fact that we're not taking into account roads, waterways, or other transportation options/restrictions.

PR #611, opened 2019-11-01
Payer loss of care

RobertScalfani:
This Pull Request adds Loss-Of-Care to Synthea when a patient cannot afford treatment and their insurance will not cover it.
The Loss-Of-Care impacts two aspects of a person:

- Their Quality Of Life
-- When a person's QOLS is calculated, based on the percent of care that they have received, their QOLS will fall in a triangular distribution of the best and worst QOLS ranges for their conditions.
- Their likelihood of death
-- When a person has certain life-saving treatments that they did not receive, there is a chance that they will die. Currently the implementation is complete with an example of a patient not receiving chemotherapy for breast cancer resulting in a 60% chance of death. However, this needs to be expanded to more treatments.


nblake989:
I like the concept @dehall referenced regarding “ability to pay” that can rely on demographics and income concepts to evaluate likelihood of outcomes downstream.

jawalonoski:
@dehall I altered this pull request to make the features optional and marked them as experimental. The functionality is disabled by default.

Loss of care is now controlled in `synthea.properties`: `generate.payers.loss_of_care` and `lifecycle.death_by_loss_of_care`.

PR #626, opened 2019-12-02
Health Record Module and Growth Errors

eedrummer:
This PR introduces the concept of a HealthRecordModule. The purpose of these modules is to modify the HealthRecord for a person. The intention is to not simulate clinical events, but to simulate events that would happen with an electronic health record, such as the incorrect entry of information.

This request also contains an implementation of the HealthRecordModule interface, GrowthDataErrorsModule. This is an implementation of the error generation protocol for childhood growth trajectories as described in the supplementary files of https://academic.oup.com/jamia/article/24/6/1080/3767271.


src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -234,6 +231,12 @@ private void init() {
       System.out.println("Modules: " + String.join("\n       & ", moduleNames));
       System.out.println(String.format("       > [%d loaded]", moduleNames.size()));
     }
+
+    if (Boolean.parseBoolean(
+        Config.get("growtherrors.enable", "false"))) {
```
jawalonoski: `growtherrors.enable = false` could probably be simplified to `growth_errors = false`

src/main/java/org/mitre/synthea/engine/HealthRecordModule.java
```diff
@@ -0,0 +1,51 @@
+package org.mitre.synthea.engine;
+
+import java.util.List;
+import java.util.Random;
+
+import org.mitre.synthea.world.agents.Person;
+import org.mitre.synthea.world.concepts.HealthRecord;
+
+/**
+ * The HealthRecordModule offers an interface that can be implemented to modify a Sythea Person's
```
jawalonoski: Spelling error: `Sythea Person's`

src/main/java/org/mitre/synthea/modules/GrowthDataErrorsModule.java
```diff
@@ -0,0 +1,390 @@
...
+      }
+    }
+  }
+
+  private static HealthRecord.Observation weightObservation(HealthRecord.Encounter encounter) {
+    return findObservation(encounter, "29463-7");
+  }
+
+  private static HealthRecord.Observation heightObservation(HealthRecord.Encounter encounter) {
+    return findObservation(encounter, "8302-2");
```
jawalonoski: Please replace magic strings (e.g. `"8302-2"`) with constants.

src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -481,11 +481,9 @@ private static double adjustWeight(Person person, long time) {
         weight = lookupGrowthChart("weight", gender, ageInMonths,
             person.getVitalSign(VitalSign.WEIGHT_PERCENTILE, time));
       } else if (age <= ADULT_MAX_WEIGHT_AGE) {
-        // getting older and fatter
```
jawalonoski: Aww, man.

eedrummer:
There appears to be some impact on performance. Here is some comparison on my laptop

`./run_synthea -p 2000 -a 2-19`

Numbers are seconds reported by gradle.

Code | Run 1 | Run 2 | Run 3
------|------|-------|-------
master | 36 | 38 | 38
HealthRecordModule change in, GrowthDataErrors off | 38 | 40 | 38
HealthRecordModule change in, GrowthDataErrors on | 40 | 39 | 39

One easy change that I will make is that on every simulation step, I'm checking for ant created encounters. If there are not modules registered, I'll just skip all of that.

In cases when modules are registered, I can also check to see if any modules `shouldRun` before creating the list of encounters. This wouldn't impact this case, but could help in other situations, the GrowthDataErrorsModule only operates on people under 20.

src/main/java/org/mitre/synthea/modules/GrowthDataErrorsModule.java
```diff
@@ -0,0 +1,395 @@
...
+   * Potentially mess up heights and weights in the encounters.
+   * @param person The Synthea person to check on whether the module should be run
+   * @param encounters The encounters that took place during the last time step of the simulation
+   * @param time The current time in the simulation
+   * @param random Random generator that should be used when randomness is needed
+   */
+  public void process(Person person, List<HealthRecord.Encounter> encounters, long time,
+                      Random random) {
+    List<HealthRecord.Encounter> encountersWithWeights =
+        encountersWithObservationsOfCode(encounters, "29463-7", "LOINC");
```
jawalonoski: WEIGHT_LOINC_CODE

src/main/java/org/mitre/synthea/modules/GrowthDataErrorsModule.java
```diff
@@ -0,0 +1,395 @@
...
+        recalculateBMI(e);
+      }
+      if (random.nextDouble() <= config.weightCarriedForwardErrorRate) {
+        introduceWeightCarriedForwardError(e);
+        recalculateBMI(e);
+      }
+    });
+
+    List<HealthRecord.Encounter> encountersWithHeights =
+        encountersWithObservationsOfCode(encounters, "8302-2", "LOINC");
```
jawalonoski: HEIGHT_LOINC_CODE

src/main/java/org/mitre/synthea/modules/GrowthDataErrorsModule.java
```diff
@@ -0,0 +1,395 @@
...
+   * @param encounter The encounter that contains the observation
+   */
+  public static void introduceWeightSwitchError(HealthRecord.Encounter encounter) {
+    HealthRecord.Observation wtObs = weightObservation(encounter);
+    HealthRecord.Observation htObs = heightObservation(encounter);
+    if (htObs == null) {
+      // If there is no existing height observation, change the weight observation into a height
+      // one
+      wtObs.unit = "cm";
+      wtObs.codes.get(0).code = "8302-2";
```
jawalonoski: HEIGHT_LOINC_CODE

src/main/java/org/mitre/synthea/modules/GrowthDataErrorsModule.java
```diff
@@ -0,0 +1,395 @@
...
+   * @param encounter The encounter that contains the observation
+   */
+  public static void introduceHeightSwitchError(HealthRecord.Encounter encounter) {
+    HealthRecord.Observation wtObs = weightObservation(encounter);
+    HealthRecord.Observation htObs = heightObservation(encounter);
+    if (wtObs == null) {
+      // If there is no existing weight observation, change the height observation into a weight
+      // one
+      htObs.unit = "kg";
+      htObs.codes.get(0).code = "29463-7";
```
jawalonoski: WEIGHT_LOINC_CODE

src/test/java/org/mitre/synthea/modules/GrowthDataErrorsModuleTest.java
```diff
@@ -0,0 +1,137 @@
...
+  private HealthRecord record;
+  private HealthRecord.Encounter first;
+  private HealthRecord.Encounter second;
+
+  @Before
+  public void setup() {
+    record = new HealthRecord(new Person(1));
+    HealthRecord.Encounter e = record.encounterStart(1000, HealthRecord.EncounterType.OUTPATIENT);
+    // Weight Observation
+    e.addObservation(1000, "29463-7", 50d);
```
jawalonoski: magic codes should reference the constants, right?

eedrummer:
@jawalonoski thanks for the review. There is still more that I would like to change beyond the comments that you made. I also like to push commits frequently so others can see my work and I don't have to worry (much) if my laptop implodes. I didn't want to close this PR and reopen, because I thought that would lose context. I don't see how I can set this back to draft, but I might just be GitHub impaired. I'll comment on the PR when it is ready again for review.

eedrummer:
@jawalonoski This is ready for review again.

I performed a spot check of running with the same seed but switching on and off the GrowthDataErrorsModule. I selected two cases where running with the module on had introduced errors in the record. I didn't see any major changes in conditions between the patients. They had variation in injuries, sore throat and ear infections.

It brings up a larger question about error modules. I would argue that there are cases where you would want errors to change the treatment someone receives in the simulation. 

There is also an odd divide in how this impacts Synthea. Errors generated by GrowthDataErrorsModule won't impact WeightLossModule, because it is looking at vital signs on Person as opposed to HealthRecord. It's probably better that way, as I suspect that a person with a normal weight wouldn't get referred to a clinical weight loss program simply due to a data entry error.

As for performance, testing with Gradle hasn't yielded consistent results for me on my laptop. It appears that the changes don't introduce any major impact. If you have any suggestions on how to more consistently benchmark, I'm open to trying it.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
```diff
@@ -454,6 +465,48 @@ public Encounter(long time, String type) {
...
+     * Add an observation to the encounter.
+     * @param time The time of the observation
+     * @param type The LOINC code for the observation
+     * @param value The observation value
+     */
+    public void addObservation(long time, String type, Object value) {
+      Observation observation = new Observation(time, type, value);
+      observation.codes.add(new Code("LOINC", type, type));
+      this.observations.add(observation);
+    }
```
jawalonoski: This seems duplicative of the `public Observation observation(long time, String type, Object value)` method. Even if you feel like it is significantly different (it only seems to have minor differences), it should be colocated with the existing method.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
```diff
@@ -454,6 +465,48 @@ public Encounter(long time, String type) {
...
+     * @param code The LOINC code to look for
+     * @return A single observation or null
+     */
+    public Observation findObservation(String code) {
+      return observations
+          .stream()
+          .filter(o -> o.containsCode(code, "LOINC"))
+          .findFirst()
+          .orElse(null);
+    }
```
jawalonoski: This also seems duplicative of the `public Observation getLatestObservation(String type)` method.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
eedrummer: They are different and I'm not sure that there is a good way to colocate these.

The root cause of the issue, IMHO, is the use of all of the inner classes in `HealthRecord`.

`addObservation` is for when you have an instance of a `HealthRecord.Encounter` and want to add an `HealthRecord.Observation` to that particular encounter. It is defined on `HealthRecord.Encounter`.

`observation` is defined on `HealthRecord`. You can only add to the current encounter. This prevents historical editing of the record.

I could pull the `observation` method up to just below the `HealthRecord.Encounter` class, but that breaks the pattern of having all of the `HealthRecord` instance methods at the bottom of the class.

src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java
eedrummer: Same comment as above.

PR #607, opened 2019-10-30
Branch to explore serialization support for patient records

hadleynet:
See if it is possible to create a set of patients, serialize the records and then reload them and add additional information at a later time.

PR #630, opened 2019-12-17
Possible solution to year conversion issues

eedrummer:
I'm throwing this up here is a draft, as it may do more harm than good. It is to generate discussion. We should fix this, but I want to make sure that the cure is better than the problem.

It appears that there is a bug in the code that converts years into
milliseconds. When performing the multiplication of value by 365.25,
the 0.25 is getting dropped. That means values will be close, but not
accounting for leap years. The way the test is written, it will work for
most of the year until about mid-December.

This change restores the intention of the code that was there, but that
broke code on the Payer side of things. This patch includes a hack to
get that test to pass, but it is likely that it could cause issues in
that code in real use.

### DIFF
--- a/src/test/java/org/mitre/synthea/world/agents/PayerTest.java
+++ b/src/test/java/org/mitre/synthea/world/agents/PayerTest.java
@@ -528,6 +528,7 @@ public class PayerTest {
     for (int year = 0; year <= 55; year++) {
       ((Map<Integer, Double>) person.attributes.get("QOL")).put(2000 + year, 1.0);
       long currentTime = startTime + Utilities.convertTime("years", year);
+      currentTime = currentTime - (year * 60 * 60 * 6 * 1000);
       healthInsuranceModule.process(person, currentTime);
     }
     int totalYearsCovered = testPrivatePayer1.getNumYearsCovered()
--- a/src/main/java/org/mitre/synthea/helpers/Utilities.java
+++ b/src/main/java/org/mitre/synthea/helpers/Utilities.java
@@ -8,6 +8,7 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.JsonPrimitive;
 
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.net.URL;
 import java.util.Calendar;
 import java.util.Random;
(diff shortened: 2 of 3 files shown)


dehall:
This kind of issue is inevitable when we mix two kinds of date handling - the multiplication-based calculations are fast but imprecise for units of time larger than weeks, and the Java Time API classes are accurate but much slower. 
For the most part the limited accuracy of multiplying out time units is fine (for instance if a module says delay 1 month, precision down to the mlllisecond is unnecessary, and potentially even inappropriate) but the final conversion to formatted date strings uses a high accuracy method. I know @jawalonoski and I talked about a specific issue related to this a while back, but I don't remember specifics.

The change to Utilities.convertTime seems appropriate to me. Maybe extract `new BigDecimal(365.25)` to a constant? I'm more concerned about why the Payer test fails though. I think this needs more investigation.

hadleynet:
I have an alternate fix for this issue in another branch:

https://github.com/synthetichealth/synthea/pull/607/commits/97b57f2c67d674c9955637143209855a2fb573c1

This maintains the multiplication-based time handling but uses Calendar in the Payer test to ensure each iteration of the loop is a full year.

hadleynet:
For now we could just fix the test to ensure each iteration spans an entire year:

https://github.com/synthetichealth/synthea/pull/607/files#diff-4d5d0ba4a57b493369686723dec80772

We can figure out if the rounding issue in the Utilities class is significant later.

src/test/java/org/mitre/synthea/world/agents/PayerTest.java
```diff
@@ -528,6 +528,7 @@ public void payerMemberYears() {
     for (int year = 0; year <= 55; year++) {
       ((Map<Integer, Double>) person.attributes.get("QOL")).put(2000 + year, 1.0);
       long currentTime = startTime + Utilities.convertTime("years", year);
+      currentTime = currentTime - (year * 60 * 60 * 6 * 1000);
```
jawalonoski: I think this is wrong.

- Line 530 says "Set current time to start time plus N years."
- Line 531 says "Set current time to current time minus N years."

Current time is always the start time.

dehall:
@eedrummer @hadleynet Is this one also OBE now that #853 was merged?

hadleynet:
> @eedrummer @hadleynet Is this one also OBE now that #853 was merged?

It does look to be OBE now, I'll wait for @eedrummer to confirm.

eedrummer:
OBE. Fixed by #853

PR #636, opened 2020-01-02
Observation Sampled Data + Attachments

nanowizard:
Adds capabilities for:

1. export of sampled data as part of an Observation from one or more patient attributes. 

2. export of Media elements (images, audio, video) during an Encounter.

All attributes indicated for sampled data must point to a `TimeSeriesData` object. Which is basically just an `ArrayList<Double>` with a corresponding `period` field indicating the time between each data point. These objects are exported automatically from Physiology states when `fromList` is used to populate a patient attribute.

The new Media state for the GMF corresponds to an [FHIR Media element](https://www.hl7.org/fhir/media.html). Static data can be provided to the module inline as Base64 encoded strings. Images may also be generated using the `chart` field which defines Patient attributes to plot on a chart image. Alternatively, a URL may be provided to indicate where the media resource is located.

From the Javadoc associated with the Media State class:

> The Media state type indicates a point in the module where an image, video, or audio recording
> is captured and preserved in the patient's health record. Like Observation states, Media states
> may only be processed during an Encounter, and so must occur after the target Encounter state
> and before the EncounterEnd. See the Encounter section above for more details.
>
>In general, ImagingStudy is the preferred state for diagnostic images. The Media state,
>however, contains functionality for generating inline PNG images of charts generated from data
>captured in Patient attributes. For instance, such data could be results from a prior
>Physiology state which simulated physiological data over time. That data could then be captured
>by a measurement device and plotted on a chart (such as an ECG signal).


jawalonoski:
@nanowizard It looks like merging in #632 caused conflicts with this PR in `PhysiologySimulator`. Can you fix?

jawalonoski:
Also, consider how the sampled data will output in the CSV observations table. ECG data is something we'll want to support in the CSV format as well.

nanowizard:
Hmm... I'm really not sure how to format that data for inclusion in the CSV file. Any way I think about doing it looks very clunky and difficult to decipher. The best I can think of without any information loss is adding some kind of ID field to the observation csv file that references a row in another CSV with sampled data metadata, which in turn will either have a URL or *another* ID field which references yet another CSV file with the data.

This might warrant more discussion before we decide how to proceed.

src/main/java/org/mitre/synthea/export/FhirDstu2.java
```diff
@@ -126,6 +133,7 @@
   private static final String SYNTHEA_EXT = "http://synthetichealth.github.io/synthea/";
   private static final String UNITSOFMEASURE_URI = "http://unitsofmeasure.org";
   private static final String DICOM_DCM_URI = "http://dicom.nema.org/resources/ontology/DCM";
+  private static final String MEDIA_VIEW_URI = "http://hl7.org/fhir/ValueSet/media-view";
```
jawalonoski: This should use just use the SNOMED URI

src/main/java/org/mitre/synthea/export/FhirStu3.java
```diff
@@ -140,6 +146,8 @@
   private static final String SYNTHEA_EXT = "http://synthetichealth.github.io/synthea/";
   private static final String UNITSOFMEASURE_URI = "http://unitsofmeasure.org";
   private static final String DICOM_DCM_URI = "http://dicom.nema.org/resources/ontology/DCM";
+  private static final String MEDIA_VIEW_URI = "http://hl7.org/fhir/ValueSet/media-view";
+  private static final String PROCEDURE_REASON_URI = "http://hl7.org/fhir/ValueSet/procedure-reason";
```
jawalonoski: Both media view and procedure reason should be SNOMED URI.

nanowizard:
I believe all of the requested changes have now been implemented for this branch. Major changes since your review include:

- CSV export of sampled data as a string of values
- Replacement of the ValueSet URLs with the corresponding CodeSystems
- Capability to set time series data manually with SetAttribute state

jawalonoski:
@nanowizard I'm not sure the unit tests are sufficient. Please take a look at the unit test that I am about to commit to this branch called `setAttribute_with_seriesData_Module`

I may be doing something wrong. Consider a module using `SetAttribute.series_data` and then an `Observation` that records that data. I would suspect that the following module would work (this module does not work in the module builder, because the module builder does not support `series_data` yet):

**series_data.json**
```json
{
  "name": "Series Data",
  "remarks": [
    "Simple module to test series data"
  ],
  "states": {
    "Initial": {
      "type": "Initial",
      "direct_transition": "Test Encounter"
    },
    "Terminal": {
      "type": "Terminal"
    },
    "Test Encounter": {
      "type": "Encounter",
      "encounter_class": "ambulatory",
      "reason": "",
      "codes": [
        {
          "system": "SNOMED-CT",
          "code": "1234",
          "display": "Test Encounter"
        }
      ],
      "direct_transition": "Series Data"
    },
    "End Encounter": {
      "type": "EncounterEnd",
      "direct_transition": "Terminal"
    },
    "Series Data": {
      "type": "SetAttribute",
      "attribute": "series_data",
      "direct_transition": "Record Series Data",
      "series_data": "1 2 3 4 4 3 2 1 1 2 3 4"
    },
    "Record Series Data": {
      "type": "Observation",
      "category": "vital-signs",
      "codes": [
        {
          "system": "LOINC",
          "code": "1234",
          "display": "Series Data"
        }
      ],
      "direct_transition": "End Encounter",
      "attribute": "series_data"
    }
  }
}
```

Instead, running this module throws the following exception during export:

```
java.lang.IllegalArgumentException: unexpected observation value class: class org.mitre.synthea.helpers.TimeSeriesData; org.mitre.synthea.helpers.TimeSeriesData@764d2e2a
        at org.mitre.synthea.export.FhirR4.mapValueToFHIRType(FhirR4.java:1556)
        at org.mitre.synthea.export.FhirR4.observation(FhirR4.java:1485)
        at org.mitre.synthea.export.FhirR4.convertToFHIR(FhirR4.java:268)
        at org.mitre.synthea.export.FhirR4.convertToFHIRJson(FhirR4.java:337)
        at org.mitre.synthea.export.Exporter.exportRecord(Exporter.java:188)
        at org.mitre.synthea.export.Exporter.export(Exporter.java:115)
        at org.mitre.synthea.engine.Generator.generatePerson(Generator.java:442)
        at org.mitre.synthea.engine.Generator.lambda$run$2(Generator.java:275)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

jawalonoski:
Or maybe my observation is supposed to look like this:

```json
    "Record Series Data": {
      "type": "Observation",
      "category": "vital-signs",
      "codes": [
        {
          "system": "LOINC",
          "code": "1234",
          "display": "Series Data"
        }
      ],
      "direct_transition": "End Encounter",
      "sampled_data": {
        "factor": 1.0,
        "lower_limit": 0.0,
        "upper_limit": 10.0,
        "attributes": [
          "series_data"
        ]
      }
    }
 ```

**Comment by nanowizard** [comment]
Yes the second one you posted _should_ work. If you are still getting an error then that means I messed up and didn't test it sufficiently.

**Comment by jawalonoski** [comment]
Standby... the second one works. I'm still reviewing.

File: src/main/java/org/mitre/synthea/helpers/TimeSeriesData.java [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TimeSeriesData {
+  /** Provides a collection of time series values as well as sampling metadata
+   *  concerning those values.
+   */
+  private List<Double> values;
+  private double period; // number of seconds between samples
```
jawalonoski: > number of seconds between samples

If this is really seconds, and not say, milliseconds, then the FHIR exporters might be wrong.

In `State.java`, you have the following code:

```java
      // Series data default period is 1.0s
      if (period <= 0.0) {
        period = 1.0;
      }
```

The `series_data.json` example that I uploaded produces the following FHIR R4 `Observation.valueSampledData`

```json
        "valueSampledData": {
          "origin": {
            "value": 0.0,
            "system": "http://unitsofmeasure.org"
          },
          "period": 1.0,
          "factor": 1.0,
          "lowerLimit": 0.0,
          "upperLimit": 10.0,
          "dimensions": 1,
          "data": "1 2 3 4 4 3 2 1 1 2 3 4 "
        }
```

But according to http://hl7.org/fhir/R4/datatypes-definitions.html#SampledData.period, `SampledData.period` is measured in **milliseconds**

nanowizard: Doh! Yes you are correct. I don't know where I got seconds from. It also confirms milliseconds is the correct measurement in https://www.hl7.org/fhir/datatypes.html#SampledData, which is the document I used as a reference.

Will fix that now.

nanowizard: Oh wait I remember this now. My thought was that the TimeSeriesData object should store in seconds since I personally think that's generally easier to work with than milliseconds. The SBSCL simulations use seconds as their form of measurement for example.\

As far as whether the user should input seconds or milliseconds, that can be modified in the State definition. I personally think it is more straightforward to enter seconds rather than milliseconds, but if you think most module developers would be more comfortable with milliseconds, we can change it.

But yes that also means we need to do the appropriate conversion when exporting with FHIR, which I messed up. I'll change that to properly convert to milliseconds.

kgoyal1988:
Hi, I was wondering if it would be possible to integrate ML models into simulation for generation of media elements. There is currently interest in this area, e.g. GAN based models, to generate synthetic medical images. Maybe such models could be integrated into the simulation by some REST API based requests to served models during transactions.

jawalonoski:
@kgoyal1988 We should probably take this comment out of this PR and open a separate GitHub issue if you'd like to have a discussion or have ideas to offer. Some of us have been discussing this topic offline. This is partially one intention or use of the new [HealthRecordEditor](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/engine/HealthRecordEditor.java) interface.

PR #641, opened 2020-01-22
Fix #608 providers with duplicate UUIDs.

### LINKED ISSUE #608: Providers CSV exports UUIDs with duplicates
Hello, it seems that even with a small population of 10 that the id of some providers is not unique in the csv export. Is this possibly the line of code that is the cause for that
https://github.com/synthetichealth/synthea/blob/bcb17a9532645f03f6410b7d5cf67092592842e7/src/main/java/org/mitre/synthea/world/agents/Provider.java#L476
or am I doing something wrong?

jawalonoski:
Fixes #608 providers with duplicate UUIDs.

Through a twist of fate, the `providers.csv` file actually contains `Clinicians`. `Providers` are written into the `organizations.csv` file.

Therefore, the fix to #608 is to increase the variation in the code that assigns UUIDs to `Clinicians`.

### DIFF
--- a/src/main/java/org/mitre/synthea/world/agents/Clinician.java
+++ b/src/main/java/org/mitre/synthea/world/agents/Clinician.java
@@ -54,7 +54,9 @@ public class Clinician implements Serializable, QuadTreeElement {
    */
   public Clinician(long clinicianSeed, Random clinicianRand,
       long identifier, Provider organization) {
-    this.uuid =  new UUID(clinicianSeed, identifier).toString();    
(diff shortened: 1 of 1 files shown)


JLMoszkowicz:
Thank you for addressing this!  Will you be updating the CSV download with the corrected data?

jawalonoski:
> Will you be updating the CSV download with the corrected data?

Sure. Done. See https://storage.googleapis.com/synthea-public/synthea_sample_data_csv_jan2020.zip

PR #644, opened 2020-01-24
Dev

irjudson:


irjudson:
It's early apologies -- I should have written a note. The <p/> element broke Javadoc for me on windows with JDK 1.8. I modified it to be <p></p> with an explicit close tag and that fixed things.

jawalonoski:
Thanks for the pull request!

PR #649, opened 2020-02-05
Improve weight growth trajectories

eedrummer:
TL;DR - This changes how weights are managed in Synthea before age 20. The changes allow for more realistic and dynamic weight growth cures, rather than following the same percentile from birth.

## What was wrong before

1. A weight percentile was selected at birth for an individual. They would then follow that percentile from 0 - 20. This is unlikely to happen in the real world.
1. Synthea uses the LMS parameters from the CDC Growth Charts to determine values from percentiles. [These really shouldn't be used beyond the 97th percentile](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4528342/).
1. Since everyone is stuck in their percentiles from birth, it meant that they only way that an individual would be overweight or obese in Synthea was to be shorter. Conversely, the only way to be underweight was to be tall.

## Changes made by this PR

This shifts the management of weight to focus on BMI. Height still tracks to percentile (I suppose that is for another project). Weight can be calculated from BMI and height. The growth trajectory works using an approach proposed by CDC staff and some adjustments made by me to fill in gaps.

1. Select a [NHANES](https://www.cdc.gov/nchs/nhanes/index.htm) sample for BMI for an individual between 24 and 35 months. Samples are selected adjusting for the NHANES sample weight.
1. From birth until the time of the selected NHANES sample, follow the growth charts to arrive at the appropriate BMI given the individual's selected height percentile.
1. After the NHANES sample, generate a next age year BMI using age year to age year correlations between extended BMI Z Scores (eBMIz) extracted from NHANES data. The eBMIz scores are normal Z Scores for anyone under the 95th percentile. Beyond the 95th percentile, values are modeled with a different, half normal distribution to better capture the distribution than the existing LMS parameters.
1. Repeat this process until age 20.

The next age year BMI generation process is as follows:

1. Determine the current eBMIz from the current BMI
1. Select a next year eBMIz from a normal distribution with the following properties: mean = (current eBMIz * correlation of current age year eBMIz to next age year eBMIz) + difference in population mean eBMIz between age years. Standard deviation = sqrt(1 - (correlation of current age year eBMIz to next age year eBMIz)^2)
1. Convert the new eBMIz back to a BMI

The weight loss module has been updated to work with the change management of weight.

## Results

This approach allows for more dynamic weight growth curves. The following is an example Synthea patient in the SMART Growth Chart app showing fluctuating weight percentiles:
Screen Shot 2020-01-03 at 10 45 32 AM

Further, this change allows trajectories seen in the real world, such as when children between 6-8 rapidly gain weight to change from a normal weight to overweight or obese, as in the following Synthea patient:
Screen Shot 2020-01-03 at 10 03 49 AM

From a population perspective, relative to the older code, this does a better job of capturing the higher end of BMI distributions. This can be seen in comparing the old code to the new code vs. NHANES at age 19:
download

## Areas for improvement

This change doesn't generate as extreme of BMI values as the previous code, which appear to be valid. It doesn't make any changes to height.

The WeightLossModule uses a change in BMI percentile for people under 20, but future work should probably change that to work off of eBMIz. There is a substantial difference between going from the 99th to the 98th percentile vs going from the 95th to the 94th.


src/main/java/org/mitre/synthea/modules/WeightLossModule.java
```diff
@@ -135,65 +132,35 @@ public void manageYearTwoThroughFive(Person person, long time) {
...
           if (person.ageInYears(start) < 20) {
-            if (age >= 20) {
-              height = lookupGrowthChart("height", gender, 240,
-                  person.getVitalSign(VitalSign.HEIGHT_PERCENTILE, time));
-            } else {
-              height = lookupGrowthChart("height", gender, ageInMonths,
-                  person.getVitalSign(VitalSign.HEIGHT_PERCENTILE, time));
-            }
-
+            height = growthChart.get(GrowthChart.ChartType.HEIGHT).lookUp(240, gender,
```
jawalonoski: If they are less than 20, lookup their height as if they were 240 months old?

src/main/java/org/mitre/synthea/world/agents/Person.java
```diff
@@ -127,6 +130,7 @@
   public Person(long seed) {
     this.seed = seed; // keep track of seed so it can be exported later
     random = new Random(seed);
+    mathRandom = new JDKRandomGenerator((int) seed);
```
jawalonoski: I would rather you just change `random` to be an instance of `JDKRandomGenerator`, since it just extends the random class anyway and adds an extra interface definition because apache commons math3 has to be special.

src/main/java/org/mitre/synthea/modules/WeightLossModule.java
jawalonoski: I probably need to checkout the code... so many changes it is hard to follow without looking at it cleanly.

eedrummer: Not exactly.

This branch is to deal with the case where you have someone who lost weight before they were 20, but are now regressing, but their regression will cross age 20. It's kind of a pain to deal with because we don't bother with height changes after that in Synthea, but prior to that, height can change, which will impact BMI.

This code is essentially saying that if you started before you were 20, but you are 20 or older now, use the value for height for 20 year olds, regardless of how far into their twenties the individual is.

src/main/java/org/mitre/synthea/world/agents/Person.java
eedrummer: That makes sense and I can do that.

eedrummer:
Got rid of `mathRandom`. Now `Person.random` is a `JDKRandomGenerator`.

PR #651, opened 2020-02-10
Bug fixes for CPCDS_Exporter Class

mosiermt:
There are a couple minor bug fixes for the class along with updates for Coverage file


src/main/java/org/mitre/synthea/export/CPCDSExporter.java
```diff
@@ -259,28 +254,28 @@ private String patient(Person person, long time) throws IOException {
    * @param careplan    The careplan itself
    * @throws IOException if any IO error occurs
    */
-  private void coverage(String personID, String encounterID, CarePlan careplan,
-      String payerId, String type, UUID groupId, String groupName) throws IOException {
-
+  private void coverage(String personID, Long start, Long stop, String payerId, String type,
```
eedrummer: Please use the primitive `long` for consistency with other code, unless there is a particular reason for using the object form `Long`.

src/main/java/org/mitre/synthea/export/CPCDSExporter.java
```diff
@@ -185,19 +173,26 @@ public static CPCDSExporter getInstance() {
...
+    
 
     for (Encounter encounter : person.record.encounters) {
-
       String encounterID = UUID.randomUUID().toString();
-      String payerId = encounter.claim.payer.uuid.toString();
       UUID medRecordNumber = UUID.randomUUID();
       CPCDSAttributes encounterAttributes = new CPCDSAttributes(encounter);
+      if (Boolean.parseBoolean(Config.get("exporter.cpcds.single_payer"))) {
+        payerId = "b1c428d6-4f07-31e0-90f0-68ffa6ff8c76";
```
eedrummer: Is there some significance to this particular ID? I would suggest either making it a constant or extracting it to a configuration file if it is significant. Otherwise, it should at least be marked with a comment if it is a random UUID.

mosiermt: Updated to be a random GUID for payerID, changed types to primitive longs, and fixed style changes

Issue #660, opened 2020-02-21
BMI Runtime Error


jawalonoski:
```
./run_synthea -s 0 -p 1000
```

```
java.lang.NumberFormatException
        at java.math.BigDecimal.<init>(BigDecimal.java:497)
        at java.math.BigDecimal.<init>(BigDecimal.java:383)
        at java.math.BigDecimal.<init>(BigDecimal.java:809)
        at java.math.BigDecimal.valueOf(BigDecimal.java:1277)
        at org.mitre.synthea.world.agents.Person.getVitalSign(Person.java:402)
        at org.mitre.synthea.modules.LifecycleModule.calculateVitalSigns(LifecycleModule.java:717)
        at org.mitre.synthea.modules.LifecycleModule.process(LifecycleModule.java:116)
        at org.mitre.synthea.engine.Generator.generatePerson(Generator.java:358)
        at org.mitre.synthea.engine.Generator.lambda$run$2(Generator.java:268)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

This rarely happens during `org.mitre.synthea.modules.LifecycleModule.calculateVitalSigns`:
https://github.com/synthetichealth/synthea/blob/e0a0d4db411239ca0b4d6aa6427d7095e600c440/src/main/java/org/mitre/synthea/modules/LifecycleModule.java#L717

My hunch is that the BMI has not been set at this time (possibly birth).

jawalonoski:
Looks like the BMI is getting set to `NaN` somewhere. Debugging shows that this issue is happening to an obese 19 year old (so not at birth) with a weight management plan. It may or may not be relevant, but the attribute `current_weight_length_percentile=99.99992386269027` makes me wonder if the very high percentile has something to do with it.

jawalonoski:
Fixed with #663

PR #661, opened 2020-02-21
Attempt to fix BMI NaN Runtime Error.

jawalonoski:
See #660 

Added some `if` conditionals to not add points after 240 months.


eedrummer:
I believe that I have found the root cause elsewhere.

PR #665, opened 2020-02-25
Add amount field to Counter state

dehall:
Adds an `amount` field to the Counter state, so that counters can count 2 by 2, or 17 by 17, or whatever number you like. Defaults to 1 if the field is not present, so all current usage of the Counter state will work without explicitly adding the field.


jawalonoski:
@dehall Please update the wiki

dehall:
wiki updated: https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-States#counter

PR #669, opened 2020-02-28
Handle weight loss regression for very high BMI percentile kids

eedrummer:
Right now, there is a small impedance mismatch between the pediatric
growth code and the weight loss module. The growth code uses extended
BMI Z Scores and extended percentile values to generate BMIs. This
can result in very high BMIs.

The weight loss module used plain old percentiles. When values hit the
99.8th percentile, it breaks the look up code. This switches the loss
code over to using the extended BMI.

This also contains a small bugfix for making sure that the
adjustBMIVectorForSuccessfulManagement only runs once and some assorted
checkstyle fixes for past transgressions.


jawalonoski:
@eedrummer you should also run `checkstyleTest` you have some transgressions in there as well.

jawalonoski:
@eedrummer You can fix the unit test checkstyle later. This is good. Passed the unit tests and ran locally without issues.

PR #672, opened 2020-03-04
Physiology State Enable/Disable

nanowizard:
Added ability to disable Physiology State simulations in the `synthea.properties` file (and they are now disabled by default). This also involved adding an `alt_direct_transition` field to Physiology state definitions to indicate where the Module should flow in the event Physiology States are disabled.

Note that this also includes updates from the (yet unmerged) `observation_media` branch.


nanowizard:
I added ECG chart and data generation to this branch as part of the gallstones module. Wanted to make sure output was correct and it seems to be in order. Data in the csv file looked correct, and the generated Base64 encoded chart came out well also:

cbimage

PR #684, opened 2020-03-24
Add Device & SupplyList to GMF

dehall:
Adds new state types Device and SupplyList, which get added to the HealthRecord data model but are not (yet) written out to any exported records.
A Device represents a permanent or semi-permanent device such as a prosthetic or a pacemaker. SupplyList represents the generic supplies needed to perform the encounter, like gloves, mask, other PPE, etc

IMPORTANT: I discovered that the new serialization feature doesn't work when the patient's record contains `JsonObject`s. Here I was using one for supplies since we aren't doing anything but exporting yet, but we're also using one already for CarePlan goals, so we should address both of those. I can do that here but wanted to get this PR up asap.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -1688,7 +1688,50 @@ public abstract class State implements Cloneable, Serializable {
       return true;
     }
   }
-
+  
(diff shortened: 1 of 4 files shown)


shabiel:
Prescriptions also use JSON objects.

PR #687, opened 2020-03-28
Mihin covid 19

jeffeastman:
This commit separates covid-19 infection logic from the module which treats that infection.  (I also removed a bunch of stuff from my previous commits)


jawalonoski:
I'm going to make several changes over the next few hours, so if you are planning more, you might want to wait to minimize conflicts.

PR #689, opened 2020-03-29
Converted severity & survival risks to lookup tables.

jeffeastman:
(Not sure about how to address the tables from the module)


jeffeastman:
Curious why this is failing to build. I'm guessing it's because of the lookup_tables but I put them in the same place as the covid19lookuptable branch. But perhaps it would fail too. Any hints from any readers?

eedrummer:
@jeffeastman It looks like you are missing some of the changes necessary to get the time based lookup tables working. Specifically this line: https://github.com/synthetichealth/synthea/pull/692/files#diff-9f3021272ba8ec37bd63780edb6ec9d9R191

Otherwise, it causes a non-serializable object to become associated with the Person, which causes the serialization tests to break.

jeffeastman:
@eedrummer Not sure why this would be an issue because I've merged with covid19 and that's the branch I'm attempting the pull to. The lookuptable mods in covid19lookuptable have not yet been merged into covid19.

eedrummer:
@jeffeastman The test that was failing had to do with new serialization tests that were added within the past month or so. The LookupTableTransition code was well tested, but not used in any mainline Synthea modules prior to this. The code that I modified prevents a `List$subList` instance from being associated with a `Person` because `List$subList` is not serializable. From my reading of the code, I'm not sure why this didn't happen with other lookup tables before. I suspect it has something to do with the number or type of columns in the lookup table. So it is possible that you are hitting this without using time based look up tables.

jeffeastman:
@eedrummer I haven't seen any other lookup tables in master (except maybe in tests) so your explanation makes sense. I'm going to merge with that branch and see what happens...

jawalonoski:
Please do not merge changes into `covid19` without my approval.

jeffeastman:
@jawalonoski Apologies, it won't happen again.

PR #696, opened 2020-04-01
Switch to a look up table for probability of COVID-19 infection

eedrummer:
This PR changes the covid19 module to use a look up table based on time to set the probability of infection. The probability of infection starts at 0.01% on January 20, 2020, the date of the first confirmed infection in the US. The probability doubles every 4 days.

Running a simulation with 10K people yielded the following results:

Screen Shot 2020-04-01 at 12 19 07 PM

Essentially, the disease burns through the entire population by roughly March 11 so new cases start to trail off. The initial rate of infection may be too high.

The lookup table is generated by a python script in `src/main/resources/modules/lookup_tables/covidtimeline.py`. Changing parameters in there and rerunning it will give you a new table to work with.


src/main/resources/modules/covid19.json
```diff
@@ -10,35 +10,32 @@
...
+        {
+          "transition": "Wait Until Exposure",
+          "default_probability": "1",
+          "lookup_table_name": "covid19_prob.csv"
+        },
         {
-          "distribution": 0.2,
-          "transition": "Call Infection Submodule"
+          "transition": "Call Infection Submodule",
+          "default_probability": "0",
```
jawalonoski, line 32: I just want to point out that you have a `default_probability` of zero in both transitions, which means if the patient does not match any row in your `covid19_prob.csv` file, then they won't go to either `Call Infection Submodule` or `Terminal`... you should probably have them default somewhere.

Even if the `covid19_prob.csv` file _currently_ has an exhaustive list of matches, I think it is a "best practice" to have defaults.

eedrummer: It does have a default transition, back to Wait Until Exposure.

jawalonoski:
I've confirmed these changes run locally and pass all the unit tests locally. I'm not sure what to do about this `testAppWithOverflow` failure that keeps occurring on Travis.

```
374 AppTest > testAppWithOverflow FAILED
375    java.lang.AssertionError: expected:<2> but was:<3>
376        at org.junit.Assert.fail(Assert.java:88)
377        at org.junit.Assert.failNotEquals(Assert.java:834)
378        at org.junit.Assert.assertEquals(Assert.java:645)
379        at org.junit.Assert.assertEquals(Assert.java:631)
380        at AppTest.testAppWithOverflow(AppTest.java:137)
```

PR #702, opened 2020-04-06
Better ICD-10-CD code for pneumonia due to coronavirus?

jeffeastman:
But I'm not sure ICD-10 is supported yet?

### DIFF
--- a/src/main/resources/modules/covid19/outcomes.json
+++ b/src/main/resources/modules/covid19/outcomes.json
@@ -15,9 +15,9 @@
       "type": "ConditionOnset",
       "codes": [
         {
-          "system": "SNOMED-CT",
-          "code": "233604007",
-          "display": "Pneumonia (disorder)"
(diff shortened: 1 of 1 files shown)


jawalonoski:
We don't currently support ICD9 or ICD10.

jeffeastman:
I thought I saw some work to add them. Michigan would like me to emit all ICD10 so I’ve added a translation table from the SNOMED condition codes to satisfy that request. Good thing the states have List<Code>. A lot of the new Covid-19 codes seem to be ICD10.

jawalonoski:
@jeffeastman I think I'm going to have to close this for now. We don't support the `ICD-10-CM` code system at the moment. The change you made incorrectly exports as a SNOMED code. We would have to update most of the exporters to handle the `ICD-10-CM` code system and export the appropriate URI (e.g. in each version of the FHIR exporters) or whatever is appropriate for the given export format.

PR #704, opened 2020-04-08
Update codes and displays to fix FHIR validator warnings

dehall:
Updates a bunch of displays, and a few codes that seem to have been wrong, based on running a large set of Synthea patients through the FHIR Validator: https://confluence.hl7.org/display/FHIR/Using+the+FHIR+Validator

The vast majority of these are only display text changes such that we are using a preferred text for the code. There are a only handful of actual codes changed.

Note that this does not address every warning, but some of the other issues identified were not straight 1:1 changes that could be made trivially. I'm not an MD or PharmD but the intent is nothing here represents a material change.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/FhirR4.java
+++ b/src/main/java/org/mitre/synthea/export/FhirR4.java
@@ -2380,7 +2380,7 @@ public class FhirR4 {
     participant.addRole(mapCodeToCodeableConcept(
         new Code(
             SNOMED_URI,
-            "116153009",
+            "116154003",
             "Patient"),
         SNOMED_URI));
     Patient patient = (Patient) personEntry.getResource();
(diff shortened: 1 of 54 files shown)


jawalonoski:
Some of these changes don't match the display values from the terminologies themselves... not sure where the validator is getting them, but they don't match what RxNorm browser says they are.

dehall:
Maybe the validator has an older version of RxNorm loaded and the display values it's suggesting are out of date? There was some discussion on the FHIR chat about another RxNorm code error at https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Validator.20failing.20on.20RxNorm.20code/near/193074346 

I can undo the RxNorm changes here if that's the way we want to go.

dehall:
Closing this and will reopen another one with just the smaller set of actual code changes, no display changes

PR #705, opened 2020-04-10
Add DeviceEnd state

dehall:
Adds a DeviceEnd state to support removal of devices associated to patients. The Device to remove can be referenced by an attribute, the name of the state where the Device was added, or by SNOMED code. Assigning Devices to an attribute was not previously supported so that has been added as well.

I also realized I forgot to update Graphviz for the Device and SupplyList states previously so that's now been added as well.

### DIFF
--- a/src/main/java/Graphviz.java
+++ b/src/main/java/Graphviz.java
@@ -404,6 +404,47 @@ public class Graphviz {
         details.append("SNOMED-CT[").append(bodySiteCode).append("] Body Site: ")
             .append(bodySiteDisplay).append(NEWLINE);
         break;
+      case "Device":
+        JsonObject c = state.get("code").getAsJsonObject();
+        details.append(toCodeString(c, true));
+        
+        if (state.has("manufacturer")) {
+          details.append("Manufacturer: ")
+            .append(state.get("manufacturer").getAsString())
+            .append(NEWLINE);
+        }
+        if (state.has("model")) {
+          details.append("Model: ")
+          .append(state.get("model").getAsString())
+          .append(NEWLINE);
+        }
+        break;
+      case "DeviceEnd":
+        if (state.has("device")) {
+          details.append("Added at: ")
+            .append(state.get("device").getAsString())
+            .append(NEWLINE);
+        }
+        break;
+      case "SupplyList":
+        List<String> supplyQuantity = new ArrayList<String>();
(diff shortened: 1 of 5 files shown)


jeffeastman:
Hi Dylan,

FHIR supports Device, DeviceRequest and DeviceUseStatement and this parallels it's Medication, MedicationRequest, MedicationAdministration in design. The way Synthea's Device is defined it looks a lot more like a Medication than a MedicationRequest or MedicationAdministration. I’ve modeled the FHIR resources in my GMF and currently my implementation of the nascent Synthea Device state is a translation to a DeviceRequest with a used flag which can thus generate all 3 FHIR resources  (Device, DeviceRequest and DeviceUseStatement). I also have a DeviceEnd state. It doesn’t look like the current Device definition is adequate to generate either DeviceRequest or DeviceUseStatements.

Wouldn’t this make more sense?

public class DeviceOrder {
	private List<Code> codes;
	private String reason;
	private String assign_to_attribute;
	private Boolean used;
	private String manufacturer;
	private String model;
}

Jeff

dehall:
Hi Jeff,
The idea makes sense but I'm not sure how all the pieces fit together with your example. Our primary driver for Devices in the GMF right now is ventilators in the COVID19 module, and our conceptual model looks kind of like this on a timeline: associate & start usage are the same event, and remove and end usage are the same event.
```
associate device &           remove device &
start usage                  end usage
| -----------------------------|------------------------>
```

It sounds like what you're proposing would make it more flexible, which is nice, though it feels to me like it needs additional state types. My interpretation:

```
associate     start        end          start            remove  
device        usage1       usage1       usage2           device 
|------------- |------------|------------|-------[...]------|----->
```

That said we've actually had Devices in the HealthRecord data model and FHIR R4 exporter for a little while -- can we take this discussion to a separate issue and more fully flesh out how we think the Device* resources and GMF components should look?

jawalonoski:
@jeffeastman this is great feedback, but I think we will probably review and merge this PR in the short-term for sake of expediency to support the `covid19` branch and then migrate to a richer representation in the near-term.

jeffeastman:
+1 I’ve opened https://github.com/synthetichealth/synthea/issues/706 <https://github.com/synthetichealth/synthea/issues/706> to take this issue into its own thread.

Jeff

PR #708, opened 2020-04-14
Update provider finder code for veterans

jawalonoski:
Update provider finder code so that veterans are more likely to use VA facilities.


src/main/java/org/mitre/synthea/modules/LifecycleModule.java
```diff
@@ -382,6 +383,16 @@ private static boolean age(Person person, long time) {
             person.attributes.put(Person.IDENTIFIER_PASSPORT, identifierPassport);
           }
         }
+        if (person.attributes.get("veteran") != null) {
+          if (person.attributes.get("veteran_provider_reset") == null) {
+            // reset providers for veterans, they'll switch to VA facilities
+            person.attributes.remove(Person.CURRENTPROVIDER);
```
dehall, line 389: Not worth changing right now but just to put it out there, maybe we should make this a function and reset providers at other times too (e.g. pediatric -> adult).

Issue #403, opened 2018-08-16
ICD-10 and/or CPT code inclusion


SQLDataNinja:
Hello,
Is there a chancee that inclusion of ICD-10 and/or CPT Codes to patient encounter and procedure data might be on the roadmap for this awesome project?

Thanks!

jawalonoski:
Sort answer is "only if it is included by the user as part of a value set."

See #347 and #380.

In general, the idea is that inside a [module](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework) a user may specify a FHIR `ValueSet` instead of a single code and the system will randomly use different codes from that value set in the records. That value set can be provided locally or from a service like VSAC (https://vsac.nlm.nih.gov).

We do not plan to embed ICD-10 or CPT in the repository due to licensing issues (both real and/or perceived).

jawalonoski:
Closed with #709.   

In order for ValueSets to be used, you need to specify a FHIR terminology server using the `generate.terminology_service_url` property and have a module with ValueSet specified.

ValueSets can be specified in any [State](https://github.com/synthetichealth/synthea/wiki/Generic-Module-Framework%3A-States) that contains a `code`. Just add a `value_set` attribute in addition to the system, code, and display.

For example:
```json
{
  "system": "SNOMED-CT",
  "code": "2501000087107",
  "display": "MRI of right knee with contrast",
  "value_set": "http://snomed.info/sct?fhir_vs=ecl/<<2491000087104"
}
```

jerrydeng:
The example value_set: "http://snomed.info/sct?fhir_vs=ecl/<<2491000087104" links to a 404 page.  I was wondering if you can give an example of local value_set or what would a terminology server/service link look like.  Does the server need to be specific query URL and give a simple txt/json/xml response?  What would be the lowest hanging fruit approach?  Modify all modules to include some ICD directly on value_set attribute for each encounter codes?

Issue #347, opened 2018-06-20
Multiple Value Sets


jawalonoski:
Request from `#fhirdevdays` to allow a [ValueSet](http://hl7.org/fhir/STU3/valueset.html) to be specified (by canonical URL) instead of a code/coding in a module, so that things like Observations, Conditions, and Procedures can have varied codes.

dehall:
Any recommended value sets that we can use? The FHIR / HL7 ones are very generic and grouped by category so probably aren't useful for our needs. (ex, there's one for "all SNOMED conditions" but not one for "all types of diabetes") VSAC has good ones but those require registration so we can't include them in the modules by default (unless we want people to have to register to use synthea).

jawalonoski:
Step 1 is to create a FHIR `ValueSet` that defines a few diabetes codes. SNOMED, ICD9, etc. Put it in the resources. Reference that valueset by canonical URL in the `ConditionOnset` state. Randomly pick one code from the value set at diagnosis time.

Step 2. Reference a VSAC value set by canonical URL. It isn't found in the resources, but the properties file defines VSAC credentials. Use the credentials to pull the value set and proceed as above.

dehall:
If we are going to use VSAC then I'd propose requiring a default code that it could fall back to if the user isn't online or doesn't have VSAC credentials.
But then it's also important to consider how this will work with `Active Condition` / `Active Medication` logic, and how this will work with claims pricing and disease burden evaluation. Not showstoppers by any means, just important to get right.

jawalonoski:
I agree that a default code is required.

rhdolin:
Greetings. I'm new to Synthea, and am just wondering if there is any more information on capabilities of using value sets when generating populations? Thanks.

jawalonoski:
It is a capability that we have experimented with, especially using VSAC (account required), but have not merged into the master branch. Meaning it isn't part of the software today. We may return to this at some point, but it hasn't been a top priority for us.

lawley:
Note, for SNOMED-based ValueSets you can make good use of the implicit ValueSet idea and avoid the need to explicitly create ValueSets.
For example, http://snomed.info/sct?fhir_vs=isa/12345678 is, __by definition__ the ValueSet for the code 12345678 and all its descendants.

You can also use SNOMED's ECL for more expressive definitions: this is all descendants of 12345678 that have a relationship 11111111 to 87654321 or any descendant
[...]

PR #709, opened 2020-04-17
Add support for ValueSet based code selection

johngrimes:
This change introduces an alternate representation of a code, which looks something like this:

```json
{
  "value_set": "http://snomed.info/sct?fhir_vs=ecl/<<195967001"
}
```

When a code is specified with a [ValueSet URI](https://www.hl7.org/fhir/valueset.html#ident) (instead of system, code and display) a random code will be selected from the expansion of the ValueSet defined by that URI. 

This is dependent upon the user configuring a [FHIR terminology service](https://www.hl7.org/fhir/terminology-service.html) within `synthea.properties` (`generate.terminology_service_url`). The FHIR terminology service must also have knowledge of the ValueSet referred to by the URI, as well as the underlying CodeSystems.

Addresses #347, #403.


src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -796,6 +850,7 @@ public boolean process(Person person, long time) {
       if (type != EncounterType.WELLNESS) {
         person.record.encounterEnd(time, type);
       }
+      materializeCodeField("dischargeDisposition", person.seed);
```
dehall, line 853: I'm guessing the use of reflection for all of these is because the `clone` method isn't doing a deep copy of all the various subfields, and so if you change the Code object in one place it gets propagated throughout every copy of that state? 

Does the use of reflection add any noticeable runtime performance degradation? I've never measured a real-world use case but I've always heard reflection is slow.

johngrimes: Yes, I agree that ideally we wouldn't need reflection.

The challenge I encountered was where to put the hooks to "materialize" the ValueSet based codes.

Initially, I thought about creating a custom serializer for Gson. The problem with this is that we need some sort of seed to ensure reproducibility, and the Person seed only gets passed in when the State is processed.

The reflection is really only there to reduce the duplicated code that would be needed within the State subclasses, with the way it is currently "manually" hooked in within the processing of each individual state.

I am sure there is a better way of doing this, I am just not sure what it is and feel that it might need more drastic changes to the way everything currently hangs together.

jawalonoski:
@johngrimes thank you for the pull request. It looks exciting, but I haven't reviewed it closely (just skimmed the code). Any ideas why the unit tests are failing?

johngrimes:
I fixed an issue where there were illegal access exceptions relating to the reflection within CI. The tests are passing now.

For some reason there is a difference in the test results between running `./gradlew check` and running something like `gradle clean build`, which is what I was running locally.

dehall:
Apologies for the delay in getting back to you - a lot going on at the moment.

This is a solid first step, but I think we need to at least think about the use of Codes in logical conditions as well - specifically the Active Condition, Active Medication, and Active CarePlan logical conditions. The idea of using value sets in those sounds fairly simple but actually is non-trivial. For example we wouldn't want the Active Condition logic to randomly pick one code from a value set to check for, but there are also potential concerns with having it check for *any* code in the value set, because then it casts a much wider net and potentially introduces unintended positives in the logic.

One thought for consideration -- what if instead of choosing a specific code from a value during the generation phase, the specific code was only selected at export time? If we required a "base" code to be included with the value set, then that code is used everywhere during generation, and that would address the concern of the Active Condition/etc logic. I'm sure there are downsides to that approach too, but just want to throw it out there for discussion.

mgramigna:
Just to add on to @dehall's latest comment about the use in logical conditions, I think this same problem applies to `End` states as well. One option when creating an `End` state is to specify a code for the Condition, Encounter, or whatever it is you are ending. I think we would want some kind of support for this still with valuesets, and it's certainly possible that something along the lines of Dylan's proposed solution above accomplishes this.

johngrimes:
@dehall @mgramigna Thanks for the feedback, sounds very sensible.

I've made some changes in the most recent commit, moving the code generation to the exporter (just FHIR R4 for now). Would you mind taking a look?

If the approach looks sound, I will expand it to cover the other exporters and add test coverage.

jawalonoski:
@johngrimes thank you for your pull request, the work, and your patience.

I think the change is good, but as I consider it, I think the change is better located in https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/export/Exporter.java.

If you put all the code replacement in the root `Exporter` you will only have to substitute codes one time, instead of once for each export format. Also, you'll be ensured that the exported codes are the same across formats.

Consolidates the changes and probably makes unit testing easier (or at least, more constrained).

johngrimes:
@jawalonoski Yep, that makes sense.

I've made another update, let me know what you think. I am yet to add tests for this code.

jawalonoski:
@johngrimes Looks very reasonable to me. This is great.

I think we'll need the ability to disable the ValueSet capability, in case a particular user does not have access to a terminology server or they just want to disable it for performance reasons. If disabled, the exporter should use the default codes.

johngrimes:
Another thing I will need to add to this is support for both STU3 and R4 terminology servers, as VSAC are still on STU3.

jawalonoski:
> Another thing I will need to add to this is support for both STU3 and R4 terminology servers, as VSAC are still on STU3.

We can just do R4 (normative) for this PR, and if we decide we need STU3 Terminology servers, we can add that in a separate PR.

johngrimes:
I think this is ready for further review, I think I have addressed the comments provided so far.

jawalonoski:
@johngrimes Thank you very much for the pull request. I merged it in with some minor changes, the biggest being a more aggressive disabling of terminology services when not configured in `synthea.properties` and reducing two of the test fixtures which contained a lot of expansions that weren't being used. Ideally we'd want to keep the ValueSet expansions to a minimum in the mocks.

johngrimes:
Thanks @jawalonoski!

We're already using it to generate synthetic data on trauma-related ED presentations, using some SNOMED CT based ValueSets which represent different categories of trauma.

jawalonoski:
Happy to take a pull request on those trauma modules too, if they are broadly applicable.

PR #710, opened 2020-04-17
Add times to dates on select CSV exports

dehall:
Adds time information to the `DATE` or `START` and `STOP` fields in the following CSV output files: (basically any care-driven information which would be precisely captured in an EHR and where the time/sequence component is potentially relevant within the scope of a single day)
- Observations
- Procedures
- Medications
- ImagingStudies
- Devices
- Immunization (now that I write this I realize it may not have that same single-day relevance, but it shouldn't hurt to include it)

Notably this doesn't include things like Conditions and Allergies which are less likely to have a well-defined start and end, or have their start and end times be meaningful within the context of a single day. If there's a push for it we could also consider just making every date in the CSVs a date+time.


dehall:
All set: https://github.com/synthetichealth/synthea/wiki/CSV-File-Data-Dictionary

PR #715, opened 2020-04-29
Fixes to the encounter module.

hadleynet:
1. Ensure all encounters started by the encounter module are ended.
2. Prevent the encounter module from starting a new encounter in the middle of another encounter from a different module.


hadleynet:
The extra verbose testing output helped to debug an issue but I meant to turn it off. I'll comment out those settings so they are easy to re-enable in the future if needed.

Issue #734, opened 2020-06-21
NullPointerException error


naomizoe:
Hi team,

I have created my own module using the module builder and downloaded the JSON file.
Next, I tried to pass it to Synthea:

C:\Users\naomi\Documents>java -jar synthea-with-dependencies.jar -d modules

However, I can see it scans my local module but I get the following error and no output:

java.lang.NullPointerException
        at org.mitre.synthea.engine.Module.lambda$null$2(Module.java:90)
        at org.mitre.synthea.engine.Module$ModuleSupplier.get(Module.java:344)
        at org.mitre.synthea.engine.Module.lambda$getModules$5(Module.java:170)
        at java.util.concurrent.ConcurrentHashMap.forEach(Unknown Source)
        at org.mitre.synthea.engine.Module.getModules(Module.java:166)
        at org.mitre.synthea.engine.Generator.init(Generator.java:214)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:143)
        at App.main(App.java:144)
Exception in thread "main" java.lang.RuntimeException: java.lang.NullPointerException
        at org.mitre.synthea.engine.Module$ModuleSupplier.get(Module.java:354)
        at org.mitre.synthea.engine.Module.lambda$getModules$5(Module.java:170)
        at java.util.concurrent.ConcurrentHashMap.forEach(Unknown Source)
        at org.mitre.synthea.engine.Module.getModules(Module.java:166)
        at org.mitre.synthea.engine.Generator.init(Generator.java:214)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:143)
        at App.main(App.java:144)
Caused by: java.lang.NullPointerException
        at org.mitre.synthea.engine.Module.lambda$null$2(Module.java:90)
        at org.mitre.synthea.engine.Module$ModuleSupplier.get(Module.java:344)
        ... 6 more

Any idea how to fix this?

Edit: the same error occurs when I save the examplitis JSON file locally and try to pass it to Synthea.

thanks in advance,
Naomi

hadleynet:
Thanks for the error report. Based on the stack trace we've identified that the automated build of the synthea-with-dependencies.jar is not working as expected and the current version is from back on November 2019. Would you be able to try the alternate developer getting started instructions while we try to track this down:

https://github.com/synthetichealth/synthea/wiki/Developer-Setup-and-Running

hadleynet:
[...]

PR #736, opened 2020-06-25
Fix for issue #734 - exception when using a local module directory

hadleynet:
Fixes the issue on MacOS, need to confirm fix on Windows.

Update 6/26 - confirmed to work on Windows.


jawalonoski:
Reviewed and locally tested the following scenarios:

- no über jar (i.e. standard development environment)
- über jar without local module directory,
- über jar with local module directory.

Proper function confirmed.

PR #737, opened 2020-06-26
remove repeated print lines

apalmer03:
Commented out init() in Generator.java to remove second "Running with options:" statement. 
All options minimally tested except for _run_synthea -m metabolic*_


jawalonoski:
@apalmer03 That does the trick. Can you just delete the line instead of commenting it out? There is no need to save it. Git will store the history for us if we ever care to rollback any changes.

apalmer03:
The init line has been deleted and the change has been pushed to the branch.

Issue #735, opened 2020-06-23
Encounters that end after death due to procedures with duration


eedrummer:
There is currently a bug in Synthea that can cause encounter end times to be after a person's death date.

The way to reproduce this is to have a `Death` state after an `EncounterEnd` state, where the encounter contains procedures with duration.

When `HealthRecord` ends an encounter, it will cycle through any procedures with a duration and [append this time to the end of the encounter](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/world/concepts/HealthRecord.java#L998). If a `Death` state is executed in the same time step, it will be before the encounter end because the death will happen at the time step, but the encounter end will be time step + procedure duration(s).

Depending on the number and length of procedures, this can cause a pretty big gap in times. An example of this is the ICU stay in the COVID-19 module. Since this has the chance of several long dialysis sessions, it can push the end date of the stay back by a couple of days.

A quick fix would be to check the encounter end if the patient is dead to ensure it never exceeds the death date. This would be consistent with procedure duration's use in the COVID-19 module, but I'm not sure what side effects it would have on other modules.

dehall:
Closed with #742

PR #742, opened 2020-07-01
Fix for encounters that end after death

eedrummer:
This change addresses issue #735.

**Edit**
The PR has now been changed.  

Procedure now acts like a Delay state. This PR introduces an abstract Delayable state that is now implemented by Delay and Procedure.

**Old PR**
Procedures with duration add time to the end of an encounter. That pushes the end of the encounter beyond the current time step. If an encounter is ended and followed by a death state, the time of death is set to the current time step. This means that the death will happen before the encounter is closed.

This PR changes the Death state to look at the end time of the last encounter. If that happens, it will push the time back to the end of the encounter.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -430,6 +430,24 @@ public abstract class State implements Cloneable, Serializable {
     }
   }
 
+  public static abstract class Delayable extends State {
+    // next is "transient" in the sense that it represents object state
+    // as opposed to the other fields which represent object definition
+    // hence it is not set in clone()
+    public Long next;
+
+    public abstract long endOfDelay(long time, Person person);
+
+    @Override
+    public boolean process(Person person, long time) {
+      if (this.next == null) {
+        this.next = this.endOfDelay(time, person);
+      }
+
+      return ((time >= this.next) && person.alive(this.next));
+    }
+  }
+
   /**
    * The Delay state type introduces a pre-configured temporal delay in the module's timeline. As a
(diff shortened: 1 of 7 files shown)


dehall:
As I look at this -- would it make more sense for the Procedure state with a duration to be handled like a Delay state? I would have expected the time for the procedure to already be accounted for once the Procedure state was finished processing. I see it isn't though, hence the adding in the EncounterEnd state.

eedrummer:
@dehall Overall, that's a better approach. The fix will likely be more invasive though. I guess we can discuss as a team if we want to invest in that.

dehall:
This shouldn't have any impact on the DeathModule right? I'm seeing a lot of instances where the death certification is ~12 days after death, which seems unusually high. But I'll try it again in master to see if that's new.
I know at one point we had talked about how long after death the Death Certification / cause of death would be reported, but I don't remember the details of that discussion.

dehall:
Seems like that's also happening on master so this looks good.
Aside - is COVID19 the only module where people die during an encounter?

eedrummer:
@dehall Death certification does happen later, but I figure that it is fine and probably similar to the real world.

Right now, I believe that the COVID-19 module is the only place where someone will die before the end of an encounter, at least in a noticeable way. The thing about the COVID-19 module is that there is a loop in the ICU where people can have a dialysis procedure for 2 - 4 hours a day. Since patients can be in the ICU for a while, this would add up and push the encounter end over 24 hours past the time of death, where it really becomes noticeable.

This change is more likely to clean up minor issues with other modules. Things that happen following an encounter with a procedure would overlap the encounter before. Now, since they are treated like a delay, they will no longer overlap.

PR #749, opened 2020-07-08
Allow GSON to read the prescription information in MedicationOrder

pete-thompson:
Not the ideal solution, but gets things working. I guess the long term solution is to change the MedicationOrder class to use a component rather than a JsonObject.


hadleynet:
Testing this locally. I added a unit test to make sure the prescription attribute was being populated but it doesn't work unless I remove the transient marker.

hadleynet:
I think I've found an alternate solution that avoids using the defaultWriteObject method for java serialization so we can omit the transient marker that was cause the GSON deserialization to skip the prescription information. I'll push up an alternate pull request.

Issue #752, opened 2020-07-13
Running synthea multiple times with same seed yields different results


tjones-hvn:
Hi, I'm running synthea specifying a seed, and I'm getting different results in `/output` when I'd expect to get identical results multiple times. 
I installed the .jar Synthea file for the first time last week. I'm following the Basic Setup and Running instructions, and the [wiki](https://github.com/synthetichealth/synthea/wiki/Basic-Setup-and-Running) says: `Populations generated with the same seed and the same version of Synthea should be identical`.

Here's what I'm running (note that I'm not specifying a custom `synthea.properties` file):
```
➜ java -jar synthea-with-dependencies.jar -s 9 -p 2
... 
Running with options:
Population: 2
Seed: 9
Provider Seed:1594648111560
Location: Massachusetts
Min Age: 0
Max Age: 140
1 -- Carina240 Hoppe518 (34 y/o F) Boston, Massachusetts
2 -- Perry780 Thompson596 (38 y/o F) Mashpee, Massachusetts
Records: total=2, alive=2, dead=0

➜ java -jar synthea-with-dependencies.jar -s 9 -p 2
...
Running with options:
Population: 2
Seed: 9
Provider Seed:1594648126542
Location: Massachusetts
Min Age: 0
Max Age: 140
1 -- Carina240 Fadel536 (34 y/o F) Boston, Massachusetts
2 -- Perry780 Gusikowski974 (38 y/o F) Mashpee, Massachusetts
Records: total=2, alive=2, dead=0

```

I would like to be able to run Synthea multiple times and get the exact same output files in `/output`. (As you can see from the output above, the last names are different). I've tried specifying a providerSeed as well and that also doesn't seem to help. How can I run Synthea multiple times and get the same result?

Thanks!

jawalonoski:
Other than the last names, is any of the data in the records different?

There is a known issue where date/times in the records may be different, but I'm asking about diagnosed conditions, medications, etc.

Regarding the surnames, my guess based on the age and gender is that each woman got married and ended up with a different spouse... the maiden name should be the same.

This is still an issue though. We'll have to track it down.

tjones-hvn:
Thanks for the prompt response. Yes, the data in the records is different too. This one example was easy to find, but I suspect there are many other differences.
`Carina240 Hoppe518` has three `Observation` resources where `Body temperature` was recorded, with values `37.704`, `37.719`, and `41.233`.
[...]

PR #756, opened 2020-07-21
Recreation

jawalonoski:
This pull request attempts to address #752.

Recreation of data requires the use of a random seed (`-s`), a clinician seed (`-cs`), and a reference date (`-r`).

Things that will not be identical:

- UUIDs: these will change across runs. Sorry.
- Export History: if you run the same parameters next year, and only keep the default 10 years of history, then the data in the record will be filtered differently (i.e. the oldest one year from the first run will not be exported)..

## Summary of Changes

- Randomness
  - I made the `Random` inside `Person` to be `private` and made a few new accessor methods (e.g. `randBoolean()`).
  - Except for a few places (e.g. `Generator.java`), that means `Random` objects are no longer passed around as method arguments, and there should be no use of `Random` in the code.
  - All random number access goes through the `Person#rand*` methods.
  - There should also be no `System.currentTimeInMillis()` calls except for a few spots (e.g. `Generator.java` and sometimes export paths when folder per run settings are used).
- Reference Dates
  - I added an optional `-r` switch for `referenceDate` which can be used to specify when Synthea bases all the birth dates from. In my testing I used `20200320` (or March 20th, 2020) but you can use any date. This defaults to now (`System.currentTimeinMillis()`) if no value is supplied.
- Modules and States
  - All modules and states are completely cloned across the patients now. The modules used to be passed directly, and only the states were cloned. Now the modules are cloned as well. This was done to remove any possibility of multi-thread race conditions on accessing modules. Possibly unnecessary.
- Wiki
  - I added this https://github.com/synthetichealth/synthea/wiki/Contributing#new-sources-of-randomness to the wiki.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Module.java
+++ b/src/main/java/org/mitre/synthea/engine/Module.java
@@ -292,8 +292,8 @@ public class Module implements Cloneable, Serializable {
     clone.name = this.name;
     clone.submodule = this.submodule;
     clone.remarks = this.remarks;
-    clone.states = new ConcurrentHashMap<String, State>();
     if (this.states != null) {
+      clone.states = new ConcurrentHashMap<String, State>();
       for (String key : this.states.keySet()) {
         clone.states.put(key, this.states.get(key).clone());
       }
--- a/src/main/java/org/mitre/synthea/engine/Module.java
+++ b/src/main/java/org/mitre/synthea/engine/Module.java
@@ -52,7 +52,7 @@ import org.mitre.synthea.world.agents.Person;
  * across the population, it is important that States are cloned before they are executed. 
  * This keeps the "master" copy of the module clean.
  */
-public class Module implements Serializable {
+public class Module implements Cloneable, Serializable {
 
   private static final Configuration JSON_PATH_CONFIG = Configuration.builder()
       .jsonProvider(new GsonJsonProvider())
@@ -285,6 +285,23 @@ public class Module implements Serializable {
   }
 
   /**
+   * Clone this module. Never provide the original.
+   */
+  public Module clone() {
+    Module clone = new Module();
+    clone.name = this.name;
+    clone.submodule = this.submodule;
+    clone.remarks = this.remarks;
+    clone.states = new ConcurrentHashMap<String, State>();
+    if (this.states != null) {
+      for (String key : this.states.keySet()) {
+        clone.states.put(key, this.states.get(key).clone());
+      }
+    }
+    return clone;
+  }
+
+  /**
    * Process this Module with the given Person at the specified time within the simulation.
    * 
    * @param person
@@ -423,7 +440,7 @@ public class Module implements Serializable {
       if (fault != null) {
         throw new RuntimeException(fault);
       }
-      return module;
+      return module.clone();
     }
   }
 }
--- a/src/main/java/org/mitre/synthea/world/agents/Person.java
+++ b/src/main/java/org/mitre/synthea/world/agents/Person.java
@@ -13,10 +13,10 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.apache.commons.math3.random.JDKRandomGenerator;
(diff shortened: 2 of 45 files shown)


src/main/java/org/mitre/synthea/engine/Module.java
```diff
@@ -423,7 +440,7 @@ public synchronized Module get() {
       if (fault != null) {
         throw new RuntimeException(fault);
       }
-      return module;
+      return module.clone();
```
dehall, line 443: Have you looked at how much this affects memory usage? The whole set of modules as raw text is 2 MB so broken out into full objects across multiple patients that might be a lot. If the real issue was the race condition in PediatricGrowthTrajectory and this increases memory usage by a lot, we may want to revert this piece.

jawalonoski: image

This is a run of 1000 patients. My max memory is set to 8 GB.

jawalonoski: image

Same run of 1000 patients from the latest `master`

src/main/java/org/mitre/synthea/world/agents/Person.java
```diff
@@ -168,7 +168,7 @@ public Person(long seed) {
   }
 
   /**
-   * Retuns a random double.
+   * Returns a random double.
```
hadleynet, line 171: Rather than having the random number related methods as public Person methods, I think it would be cleaner if they were all gathered together into an interface (e.g. RandomProvider) that Person implements. Methods that need access to randomness but not other aspects of Person could then just accept a RandomProvider instead of Person and this would make it clear which methods need Person access vs just Randomness.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -439,7 +439,8 @@ private Clinician generateClinician(long clinicianSeed, Random clinicianRand,
       long clinicianIdentifier, Provider provider) {
     Clinician clinician = null;
     try {
-      Demographics city = location.randomCity(clinicianRand);
+      Person doc = new Person(clinicianIdentifier);
```
hadleynet, line 442: Should this be Person(clinicianSeed) instead?

src/main/java/org/mitre/synthea/engine/Module.java
dehall: Thanks, doesn't seem like there's a meaningful difference

src/main/java/org/mitre/synthea/world/agents/Person.java
jawalonoski: An interface like RandomProvider is a good idea. I'll work on that.

src/main/java/org/mitre/synthea/world/agents/Provider.java
jawalonoski: No. I admit that it is a little hard to follow, but `clinicianSeed` is the seed for ALL of the clinicians that can be passed in with the GeneratorOptions. The clinicianIdentifier is the seed for THIS particular clinician.

I'll make some changes in this area when I refactor with a `RandomProvider` as suggested above.

src/main/java/org/mitre/synthea/engine/Module.java
```diff
@@ -284,6 +284,23 @@ public Module(JsonObject definition, boolean submodule) throws Exception {
...
 
+  /**
+   * Clone this module. Never provide the original.
+   */
+  public Module clone() {
+    Module clone = new Module();
+    clone.name = this.name;
+    clone.submodule = this.submodule;
+    clone.remarks = this.remarks;
+    clone.states = new ConcurrentHashMap<String, State>();
```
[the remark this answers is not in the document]
jawalonoski: Good catch.

src/main/java/org/mitre/synthea/world/agents/Person.java
jawalonoski: I called the interface `RandomNumberGenerator` because I feel like `Provider` is overloaded with meaning, e.g. healthcare provider.

jawalonoski:
@hadleynet Ready for your re-review.

hadleynet:
Rebased and merged.

PR #761, opened 2020-08-05
Fix DICOM UIDs

dehall:
Discovered via CI that the DICOM UIDs include the current timestamp, which if before 1970 are negative and the minus sign breaks the expected format. These aren't expected to be 100% guaranteed unique in the way UUIDs are so just use the absolute value of the time.


eedrummer:
@dehall Considering that the DICOM spec wasn't released until 1993, I think we should be find with UIDs that I guess could be duplicates prior to 1970.

PR #778, opened 2020-09-02
Add gmfVersion property to GMF JSON, make Synthea check it

eedrummer:
This PR is to start making Synthea aware of versioning of GMF JSON. It does a few things:

* Adds a `gmfVersion` field to the `Module` class, so it can be loaded from the JSON
* Adds a `GMF_VERSION` constant to `Module`
* When a module is loaded, it will check to see if there is a `gmfVersion`. If one is present and it is greater than the version known in `GMF_VERSION`, it will throw an exception. If there is no `gmfVersion` property in the JSON, it will still try to load it.
* Added some JavaScript to recursively process a directory and add `gmfVersion` to all of the JSON files in it
* Updated existing modules


src/main/java/org/mitre/synthea/engine/Module.java
```diff
@@ -276,6 +279,14 @@ public Module(JsonObject definition, boolean submodule) throws Exception {
       }
     }
 
+    if(definition.has("gmfVersion")) {
```
hadleynet, line 282: Any reason you put this below the remarks processing block?

src/main/resources/modules/allergic_rhinitis.json
```diff
@@ -360,5 +360,6 @@
     "Terminal": {
       "type": "Terminal"
     }
-  }
+  },
+  "gmfVersion": 1
```
hadleynet, line 364: I know it makes no difference but, just for eye-balling convenience, it would be nice if this was the first property.

src/main/resources/modules/lung_cancer.json
```diff
@@ -1501,7 +1501,7 @@
       "type": "ImagingStudy",
       "procedure_code": {
         "system": "SNOMED-CT",
-        "code": 16335031000119103,
+        "code": 16335031000119104,
```
hadleynet, line 1504: Did you mean to include these code changes in this pull request?

src/main/resources/modules/gallstones.json
```diff
@@ -1252,42 +1255,50 @@
...
       "unit": "",
-      "codes": [{
-        "system": "SNOMED-CT",
-        "code": 29303009,
-        "display": "Electrocardiographic procedure"
-      }],
+      "codes": [
+        {
+          "system": "SNOMED-CT",
+          "code": 29303009,
```
hadleynet, line 1261: Surprised to see code as a number here instead of a string.

src/main/java/org/mitre/synthea/engine/Module.java
eedrummer: That's a good point. I just threw it in the middle of the constructor. It should go towards the top, because there is no point in processing remarks if we are just going to throw an exception later.

src/main/resources/modules/lung_cancer.json
eedrummer: No... and I don't have any idea how this happened.

dehall: Double precision issue maybe?

src/main/resources/modules/gallstones.json
dehall: That's a limitation of the current module builder, unfortunately. The text field doesn't have any way of indicating what type something is supposed to be, and in the case of attributes it can be anything (boolean, string, number). So for codes that look numeric they get written out as numbers.

src/main/resources/modules/lung_cancer.json
eedrummer: @dehall yes...

```
Welcome to Node.js v14.8.0.
Type ".help" for more information.
> foo = '{"code":16335031000119103}'
'{"code":16335031000119103}'
> parsed = JSON.parse(foo)
{ code: 16335031000119104 }
> Number('16335031000119103')
16335031000119104
> Number.isSafeInteger(16335031000119103)
false
> 
```

src/main/resources/modules/allergic_rhinitis.json
eedrummer: This is kind of a pain to do with `JSON.stringify`

src/main/javascript/add_version.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+    fileInFolder = dirEntry.readSync();
+  }
+}
+
+function addVersion(moduleJSONPath) {
+  console.log(moduleJSONPath);
+  const rawJSON = fs.readFileSync(moduleJSONPath);
+  let module = JSON.parse(rawJSON);
+  if(module['gmfVersion'] === undefined) {
```
dehall, line 43: I know this is super minor but can we call the new field `gmf_version`? All the other fields in the JSON are `snake_case`. The GSON library will transmogrify it automagically into the field `gmfVersion` in the java code.

eedrummer: Done

hadleynet (approved):

PR #779, opened 2020-09-02
Fix procedures with delays to ensure the record entry is only added once

dehall:
This PR tweaks the `Delayable` interface that `Procedure` and `Delay` states and adds a new `processOnce` method, which should only be executed a single time, regardless of how many times the `process` method gets called. The `Procedure` state now overrides this "hook" method instead of having a `process` method which calls back to `super.process`.
(I can envision in the future we might want a "processEveryTimestep" hook method as well, but since we don't need one today and it's trivial to add later, I didn't add it now)

This addresses a bug where Procedures were being added to the record twice because the corresponding state was processing twice.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -438,9 +438,24 @@ public abstract class State implements Cloneable, Serializable {
 
     public abstract long endOfDelay(long time, Person person);
 
+    /**
+     * Process any aspect of this state which should only happen once.
+     * Because of the nature of Delay states, this state may get called
+     * multiple times: once per timestep while delaying. To ensure
+     * any actions which are supposed to happen only happen once,
+     * this function should be overriden in subclasses.
+     *
+     * @param person the person being simulated
+     * @param time the date within the simulated world
+     */
+    public void processOnce(Person person, long time) {
+      // do nothing. allow subclasses to override
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -438,9 +438,24 @@ public boolean process(Person person, long time) {
...
+     * Process any aspect of this state which should only happen once.
+     * Because of the nature of Delay states, this state may get called
+     * multiple times: once per timestep while delaying. To ensure
+     * any actions which are supposed to happen only happen once,
+     * this function should be overriden in subclasses.
+     *
+     * @param person the person being simulated
+     * @param time the date within the simulated world
+     */
+    public void processOnce(Person person, long time) {
```
eedrummer, line 451: Shouldn't this be an `abstract` method as well, like `endOfDelay`?

dehall: Yeah it could be. I saw 2 options -- (1)make this abstract and add an empty implementation in `Delay`, or (2)have an empty implementation here and `Delay` doesn't need any changes. I'm happy to change it if there's a strong preference.

eedrummer: I was hoping to avoid an empty implementation, but I guess we can't do that... so we can leave it as is.

PR #784, opened 2020-09-09
Use “Non Hispanic or Latino” instead of “Not Hispanic or Latino”

### LINKED ISSUE #783: us-core-ethnicity may contain invalid display value
See http://hl7.org/fhir/us/core/STU3.1.1/ValueSet-omb-ethnicity-category.html#expansion "Not Hispanic or Latino" is used instead of "Non Hispanic or Latino".

alexzautke:
Fixes #783


jawalonoski:
Thanks for finding the issue and fixing it!

PR #788, opened 2020-09-11
Support different distribution types for random variables in a new GMF format

eedrummer:
This change supports a new `distribution` property in `Delay`, `Observation`, `Procedure`, `SetAttribute`, `Symptom` and `VitalSign` states. 

For example, an `Observation` of LDL Cholesterol with a Gaussian distribution of results would be:

```json
    "Observation" : {
      "type" : "Observation",
      "category" : "procedure",
      "codes" : [
        {
          "system" : "LOINC",
          "code" : "2089-1",
          "display" : "Cholesterol in LDL"
        }
      ],
      "distribution": {
        "kind": "GAUSSIAN",
        "parameters": {
          "mean": 140,
          "standardDeviation": 15
        }
      },
      "unit": "mg/dL",
    }
```

All states still support the "legacy" format for defining ranges or whatever they call themselves. Current distribution types in this PR are:
* Exact
* Uniform
* Gaussian

The Distribution class should make it easy to add any other distribution types desired.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -464,6 +464,26 @@ public abstract class State implements Cloneable, Serializable {
     }
   }
 
+  public abstract static class LegacyStateWithUnitlessRV extends State {
+    protected Range range;
+    protected Exact exact;
+
+    @Override
+    public State clone() {
+      LegacyStateWithUnitlessRV clone = (LegacyStateWithUnitlessRV) super.clone();
+      clone.range = range;
+      clone.exact = exact;
+      return clone;
+    }
+
+    public boolean isLegacyGmf() {
+      if(this.exact != null || this.range != null) {
+        return true;
+      }
+      return false;
+    }
+  }
+
   /**
    * The Delay state type introduces a pre-configured temporal delay in the module's timeline. As a
    * simple example, a Delay state may indicate a one-month gap in time between an initial encounter
@@ -480,8 +500,12 @@ public abstract class State implements Cloneable, Serializable {
    * step) time.
    */
   public static class Delay extends Delayable {
+    // For GMF 1.0 Support
     private RangeWithUnit<Double> range;
     private ExactWithUnit<Double> exact;
+    // For GMF 2.0 Support
+    private Distribution distribution;
+    private String unit;
 
     @Override
     public Delay clone() {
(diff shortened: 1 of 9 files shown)


dehall (commented):
Haven't tested yet but just looking at the code everything seems good, will test it out shortly. One potential idea - currently this is checking for "isLegagyGMF" every time the given state processes, and falls back to legacy behavior if it is. It's only a simple null check so not computationally intensive, but would it simplify things at all to instead "upgrade" the module at initialize time into the latest format?

hadleynet (changes requested):

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1403,9 +1487,18 @@ public void processOnce(Person person, long time) {
...
+        double durationVal;
+        String procedureDurationUnit;
+        if (duration != null) {
+          durationVal = person.rand(duration.low, duration.high);
+          procedureDurationUnit = duration.unit;
+        } else {
+          durationVal = distribution.generate(person);
+          procedureDurationUnit = this.unit;
+        }
+        this.stop = procedure.start + Utilities.convertTime(procedureDurationUnit,
```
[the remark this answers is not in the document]
hadleynet: Will fix this separately.

eedrummer:
The changes that @hadleynet made look good to me. I like @dehall 's idea about up converting GMF 1.0 states on loading the JSON, as that would allow us to get rid of GMF 1.0 code everywhere but the initialization code, but I don't know if we want to take that on right now.

@jawalonoski I know you don't have a lot of time to look at this, but it would be great to get your thoughts.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1756,27 +1879,49 @@ protected void initialize(Module module, String name, JsonObject definition) {
...
+      if (distribution != null && !distribution.validate()) {
+        throw new IllegalStateException(
+            String.format("State %s contains an invalid distribution", this.name));
+      }
     }
 
     @Override
     public Symptom clone() {
       Symptom clone = (Symptom) super.clone();
+      clone.symptom = symptom;
```
hadleynet, line 1891: Same comment re super.clone doing the work of copying members.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1313,19 +1378,38 @@ public boolean process(Person person, long time) {
...
+      if (distribution != null && !distribution.validate()) {
+        throw new IllegalStateException(
+            String.format("State %s contains an invalid distribution", this.name));
+      }
+    }
 
     @Override
     public Procedure clone() {
       Procedure clone = (Procedure) super.clone();
+      clone.codes = codes;
```
hadleynet, line 1401: I think the super.clone will take care of copying all of the member variables so you can remove lines 1401-1406 inclusive.

hadleynet (changes requested):
Two small changes in your clone methods.

eedrummer:
@hadleynet I cleaned up the clone methods. I think some of that was pulled back in by the rebase I did off of master and not correctly resolving the conflict.

hadleynet (approved):

Issue #768, opened 2020-08-13
Missing locations in Maryland


dehall:
Major locations in Maryland are currently missing from the default `demographics.csv` file, for example Ellicott City (pop. ~ 65k), Columbia (pop ~ 100k), and Towson (pop ~ 55k). 

I suspect the root cause is that Howard County and Baltimore County in MD do not contain any incorporated locations (the 3 examples above are all unincorporated communities and CDPs) and were somehow excluded from the data source(s) that produced the original demographics.csv file.

We should:

- [ ] Add the missing locations in Maryland
- [ ] Try to identify any other major locations that may be missing
- [ ] Ideally, create a clean and easy-to-use script that allows for re-generating the demographics file with updated data

dehall:
Some other examples of CDPs with large populations are
- Paradise, NV (pop > 200k)
- Metairie, LA
- Arden-Arcade, CA

(see [5bd92fcf9129a5b406d3ff16fa3baf58b817.pdf](https://github.com/synthetichealth/synthea/files/5190947/5bd92fcf9129a5b406d3ff16fa3baf58b817.pdf) for a list of CDPs with pop > 50,000 as of 2000. I'd like to figure out their methodology to find CDPs in a more recent census)

In short: it's not just Maryland.

PR #798, opened 2020-09-24
Demographics file entries for CDPs

dehall:
This PR adds new entries to the demographics and zip codes files for Census-designated places (CDPs), which were unintentionally excluded from the previous versions. This is most notable in Maryland, where several large counties contain no incorporated communities. (see issue #768)

These were mashed together from various data sources, including primarily the Census/American Community Survey:
   - S0101: Age and Sex (All Places)
   - S1901: Income in the last 12 months (All Places)
   - S1501: Educational Attainment (All Places)
   - County population demographic distributions (gender, race, age groups) (`cc-est2015-alldata.csv`) 
     - Note that a lot of the demographic information is derived from county-level information, as was done previously in the old Ruby version: https://github.com/synthetichealth/synthea/blob/ruby/lib/tasks/tasks.rake#L356
   - National places list: https://www2.census.gov/geo/docs/reference/codes/files/national_places.txt
   - National Places Gazetteer File: https://www2.census.gov/geo/docs/maps-data/data/gazetteer/2019_Gazetteer/2019_Gaz_place_national.zip

I know there are still some issues, missing entries, etc, so I'm putting this up for WIP in case people want to get an early look, and so I can track the fixes.

Known issues:
 - [x] Missing counties for certain places, which means the demographic info is also missing
 - [x] Missing lat/lon for certain places
 - [x] Need to check that every location in `demographics.csv` has a corresponding line in `zipcodes.csv`
 - [x] Need to check more rigorously for any duplicates
 - [ ] Encoding issues, ex ñ becoming � (edit: turns out this is an issue in the source data. I have cleaned some up, but apparently CSV was not designed to be UTF-8/16 compliant so the source data is all over the place)
 - [ ] Missing zip codes (Since any missing zip codes get replaced with 00000 this is the lowest priority)

For the moment these are separate files so they are easier to review, but once everything looks good in them I'll merge them back into the regular `demographics.csv` and `zipcodes.csv` files. (Potentially by just appending to the end, to minimize the chance of breaking anything)


dehall:
Marking this as ready for review, although because the files are still separate, please don't merge yet.

For posterity, the "script" that I used to create these is attached. (It's in Java not python because I know Java a lot better.)
There are a lot of hacks and manual additions
[CDPDemographicsBuilder.java.txt](https://github.com/synthetichealth/synthea/files/5314231/CDPDemographicsBuilder.java.txt)
 in here to try to get things to line up, apologies in advance to anyone who tries to read it.

dehall:
Also I don't understand why Travis is failing at this point. As it stands the 2 new files aren't even referenced anywhere yet.

eedrummer:
@dehall I started to analyze the new files and a few things jumped out at me:

The new demographics file is substantially smaller than the old one (8746 vs. 34492). Is this intended? 

A few quality checks failed for me using https://nbviewer.jupyter.org/gist/eedrummer/8d226fe5f7d106f9594889df1075fcfb
* The new zip code file has 45 duplicate entries in it
* There are 199 locations in the zip code file with no corresponding entries in the demographics file. So with the way the code works in Synthea, these are essentially dead lines in the file

I think we would need to clean up these issues and know what we are losing before we merge this.

Also, it's worth checking the major cities, to see if the zip code lat/lon are spread around appropriately. 

Finally, it's good to look at a lat/lon difference inside of a city. I had run into cases with clearly wrong lat/lon for cities with multiple zip codes, where you would have points that were 200 miles apart for the same place.

dehall:
The new files are intended to be an addition to the existing ones, for now just kept separate so it's easier to process. So the different size is expected. It just happens to be smaller because there aren't many CDPs as compared to whichever form of geographic location we were using to populate the original file.

I think I know why the duplicate zip codes are happening and that's easy to fix (CDPs can be within multiple counties, so they are listed twice in the census data files)

Not sure why there are files in zipcodes that are not in demographics though, that doesn't really make sense. Will look closer into that. (midway update: it becomes apparent as you step through the code. Most likely those are places with 0 population, which got deleted from the demographics file after they were used to create the zip codes file)

For posterity, this is the basic approach the script took:

 - Pre-processing steps:
   - Load the "S1901: Income in the last 12 months (All Places)" file, remove all places without "CDP" in the name
   - Load the "S1501: Educational Attainment (All Places)" file, remove all places without "CDP" in the name
   - Load the "S0101: Age and Sex (All Places)" file, remove all places without "CDP" in the name
   - Load the "County population demographic distributions" file
     - Create a mapping of county --> { male, female, race/ethnicity }
   - Load the existing Synthea demographics file
     - Create a set of all places already included there
   - Load the "National places list" file
     - Create a mapping of place --> county
     - Add/update some specific entries
   - Load the "National Places Gazetteer File"
     - Create a mapping of place --> { lat, lon }
   - Load the "USA Zip" file
     - Create a mapping of place --> zip
   - Create a manual set of places to exclude for various reasons (some CDPs have been discontinued, or are already included elsewhere)

 - Processing:
   - Create a map of string -> entry to put the data for the new demographics file in
   - For each line in the "income all places" file:
     - Skip it if it's in the excluded places set
     - Skip it if it's already in the existing demographics file
     - Create a new entry in the new demographics list
     - Look up the county name in the place --> county mapping
     - Add the gender/race/ethnicity info from the county to this place's demographics entry
     - calculate the income stats for this place, and add it to this place's demographics entry
   - For each line in the "education all places" file:
     - Skip it if it's not in the new demographics map
     - calculate the education stats for this place, and add it to this place's demographics entry
   - For each line in the "age all places" file:
     - Skip it if it's not in the new demographics map
     - calculate the age stats for this place, and add it to this place's demographics entry
   - Create a list of entries to put the data for the new zipcodes file in
   - For each place in the new demographics map
     - Create a new entry for the zipcodes file list
     - Look up the lat/lon in the place --> lat/lon mapping and add it to this entry
     - Look up the zipcode in the place --> zip mapping and add it to this entry
   - Remove entries from the demographics map if the population is 0
   - Write out the demographics file
   - Write out the zipcodes file

dehall:
@eedrummer updated to remove the mismatches and duplicates. 
Re: checking lat/lon - I only have 1 zip code/coordinate line for each location, which is definitely reductive. I'll see if I can get multiples where appropriate, the challenge is really in aligning data sources of varying quality.

eedrummer:
@dehall I didn't realize that the files were additions. Let me know when they are ready for another look and I'll run some checks on the combined files in the notebook I have.

dehall:
@eedrummer This should be good now. Every line in demographics should have 1 or more lines in zipcodes, and every group of lines in zipcodes should have 1 line in demographics. I went back to an old version of the synthea zipcodes file, before a bunch of "unused" zipcodes were removed, and reintroduced those. So now places like Paradise, NV should have many zip codes with varying coordinates.
https://github.com/synthetichealth/synthea/blob/b3ddfd36acbfcccadbe08b2010a939984d4d2f5a/src/main/resources/geography/zipcodes.csv

eedrummer:
@dehall sorry to be a pain about this, but just looking through `zipcodes_cdps.csv` I'm finding issues.

The file still has duplicates and there are also data issues that will need to be dealt with.

Look at Mount Olive, AL as the first example of duplicates:
https://github.com/synthetichealth/synthea/blob/cdp_demographics/src/main/resources/geography/zipcodes_cdps.csv#L128

The three zip codes for that town appear to be repeated. Further, one of the lat/long combos is off. Comparing the distance between the first lat/long and the last lat/long for the town is about an 80 mile drive according to Google Maps.

I can check this stuff pretty quickly with Pandas in python. I think having a library that offers Data Frame-like capabilities or able to execute SQL on a CSV file in java would be handy here if you don't want to go that route. You can do things like group by zip code and calculate outliers from the average lat/long without even getting into actual geographic distance just to filter out crazy values. Unfortunately, I've found some pretty crazy values in the underlying data from census.

dehall:
@eedrummer No worries! That's why we do PRs. And it's my fault for bringing back rows from an old version of the file without double checking them closely.
I can clean up dups super easily, but in general I'm not sure what to do about weird lat/lons. 

You highlighted these 3 lines:
```
,Alabama,AL,"Mount Olive",35072,33.093498,-86.066947
,Alabama,AL,"Mount Olive",35071,33.694975,-86.848078
,Alabama,AL,"Mount Olive",35117,33.667778,-86.892599
```

According to well-known beacon of truth https://www.unitedstateszipcodes.org/ -- Mount Olive, AL, appears to only have 2 zip codes: 35071 and 35117. 35072 is in fact 80 miles away in Goodwater, AL. So the issue was we somehow associated 35072 to Mount Olive incorrectly. I'll revisit all our data sources but if there's weirdness in data from the census or other official sources I don't really want to have to be the arbiter of what's valid.

dehall:
Quick update: I have no idea where the original zip codes list came from. This was the PR where it originated, way back in the Ruby version:
https://github.com/synthetichealth/synthea/pull/254

Note that it already contained the 3 rows for Mount Olive:
```
815,Alabama,Mount Olive,35072
816,Alabama,Mount Olive,35071
817,Alabama,Mount Olive,35117
```

Now when I do a little research, I found this source, which says that 35072 is associated to Mount Olive, plus a few other places:
https://www.uszip.com/zip/35072

> 35072 is the zip code for
> Goodwater, AL
> 
> Also in this location:
> Hollins, AL
> Mount Olive, AL
> Hackneyville, AL
> Millerville, AL

So the latest I just pushed should have removed all the duplicates, but any "weird" zip code lines like the Mt Olive one will still be there. Any suggestions?

eedrummer:
@dehall My suggestion is to drop the weird zip code in this case.

The last time that I went through the zip code file, I grouped it on city, because that's how it is really used in Synthea. I would then compare the lat / long values to see if there are cases with any outliers. The outliers were causing me problems earlier, I think mostly because there was a zip code for Denver that is way out by the airport that was causing a bunch of people to be generated there, which isn't realistic.

So, for cases where we have multiple zip codes associated with a single place, my argument is that we should have those agree from a lat / long perspective. If we have to drop one, it's not a big deal, because the town is still in Synthea and all we lose is a potential portion of the town ending up in a zip code that may or may not be valid, which I don't think anyone will care about.

Finally, it may be that 35072 really is a valid zip code for Mount Olive. We're treating zip codes as polygons and getting a lat / long value for the centroid of that polygon. In reality, zip codes are a list of addresses. They may not even represent a constant physical address (my understanding is that there are ships with zip codes). So, there may be an address or block in Mount Olive in the 35072 zip code that is some weird outlier of postal service history. I just think that in our case, it makes the most sense to drop those values.

dehall:
@eedrummer Updated one more time. Hopefully the last change to the zip codes file since:
 - there should be no duplicate entries
 - there should be no instances of blank zip code lines, followed by other lines for the same city that do have zip codes
 - every line in demographics should have at least one corresponding line in zipcodes, and vice versa
 - there should be no instances of the same city having locations > 0.3 degrees apart. Even if you set the threshold at 0.2 degrees there are only a couple instances, which look correct on the map: Pahrump, NV and The Woodlands, TX. I can't link directly to a search result but this site has a nice map: https://www.unitedstateszipcodes.org/

Not every line in zipcodes_cdps.zip has a zip code, but this should be ok since the synthea engine will just use 00000 in those cases. It will still use the lat/lon from the zipcodes file based on the city/state name.

And just a reminder, this is intended to be a supplement to the existing demographics and zipcodes files, just separated here for ease of reviewing. If you want to combine them (unsorted), run these 2 commands:

```
tail -n +2 src/main/resources/geography/demographics_cdps.csv >> src/main/resources/geography/demographics.csv
tail -n +2 src/main/resources/geography/zipcodes_cdps.csv >> src/main/resources/geography/zipcodes.csv
```
In the meantime I'm working on the command to run to sort them afterward

eedrummer:
@dehall Looks great! Once you consolidate the files, I'll merge this PR.

dehall:
@eedrummer merged the CDP files into the main files, and they should be in the proper order. I did notice there are some existing quirks in the main files, such as places having multiple lines where one has a zip code and one doesn't. (For example, Albion ID) I didn't address those here, partly to make it clear what this change was, and partly because my script wouldn't be a plug&play fix there) but I can if you'd like.

PR #817, opened 2020-11-10
WIP: Improving/expanding Cardiovascular Disease modules

dehall:
Work in progress to break out the currently Java-based CardiovascularDiseaseModule into a number of GMF modules for specific conditions/cardiac events.

The basic idea is to create a centralized "cardiovascular risk" module, which will calculate various risk scores, and then a set of modules that will use those risk scores to drive when events like MI, stroke, etc, occur. Step 1 of this will keep the risk score calculation in a Java module, and potentially we can try to convert that to GMF as well later. This also introduces the ACC pooled cohort equation for cardiovascular risk as a supplement to the Framingham risk scores. We're likely to support multiple risk score systems/frameworks to allow for different use cases and comparisons between them.

See https://github.com/synthetichealth/synthea/projects/2 for the full list of what we're aiming to do


dehall:
Closing this since active development has switched over to `cabg_cardio_gmf`

PR #820, opened 2020-11-12
Build Update

hadleynet:
This branch updates the Gradle build to:

1. Ensure generated project artifacts are compatible with Java 1.8
2. Fix Gradle warnings to ensure future compatibility with Gradle 7.0
3. Update dependencies to most recent compatible version


build.gradle
```diff
@@ -1,13 +1,16 @@
...
- * This generated file contains a sample Java project to get you started.
- * For more details take a look at the Java Quickstart chapter in the Gradle
- * user guide available at https://docs.gradle.org/3.5/userguide/tutorial_java_projects.html
+ * This file defines the steps to build Synthea using Gradle. Full documentation of the syntax
+ * can be found at https://docs.gradle.org/6.5/userguide/java_plugin.html
  */
 
 // Apply the java plugin to add support for Java
 apply plugin: 'java'
+
```
dehall, line 8: Can I request the two compatibility lines also be added outside the `compileJava` scope as well? Not sure if it's an eclipse bug, but if those aren't included, it picks the wrong target JDK when you import, and the build fails because some XML stuff isn't included as of Java 9. I don't think it should have any other impact.

i.e,
```diff
+sourceCompatibility = '1.8'
+targetCompatibility = '1.8'
+
+compileJava {
+    sourceCompatibility = '1.8'
+    targetCompatibility = '1.8'
+}
```

dehall (commented):
One other thought - is this a good time to update `.travis.yml` to run tests on openjdk15? We're currently using 13 which I think is considered out of date in this new rapidly-iterating-java world.

dehall:
Oh I am getting one issue with the `uberJar` command. I wouldn't even have thought of it if not for the "archiveBaseName" change. On master it works as expected, but here I get an error when I try to run the created jar. (Note that running as usual with `run_synthea` it's fine)

```
$ java -jar build/libs/synthea-with-dependencies.jar -p 0
Exception in thread "main" java.lang.NoClassDefFoundError: com/google/gson/TypeAdapterFactory
	at org.mitre.synthea.world.geography.Location.loadAbbreviations(Location.java:403)
	at org.mitre.synthea.world.geography.Location.<clinit>(Location.java:26)
	at org.mitre.synthea.engine.Generator.init(Generator.java:196)
	at org.mitre.synthea.engine.Generator.<init>(Generator.java:171)
	at org.mitre.synthea.engine.Generator.<init>(Generator.java:156)
	at App.main(App.java:199)
Caused by: java.lang.ClassNotFoundException: com.google.gson.TypeAdapterFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 6 more
```

Size comparison:
master: 
`-rw-r--r--  1 dehall  staff  138471557 Nov 23 10:35 synthea-with-dependencies.jar`
PR: 
`-rw-r--r--  1 dehall  staff  31704638 Nov 23 10:34 synthea-with-dependencies.jar`

At a glance I can't tell why it seems not to be including things it previously did. Maybe something about this line `     from { configurations.compile.collect { it.isDirectory() ? it : zipTree(it) } }` but I don't know what other configurations there are besides `compile`. "implementation" doesn't work at all, and changing it to "runtime" based on a stackoverflow post didn't change anything that I can see.

build.gradle
```diff
@@ -154,7 +159,7 @@ task uberJar(type: Jar) {
             'Build-OS'       : "${System.properties['os.name']} ${System.properties['os.arch']} ${System.properties['os.version']}"
         )
     }
-    baseName = 'synthea-with-dependencies'
+    archiveBaseName = 'synthea-with-dependencies'
     from { configurations.compile.collect { it.isDirectory() ? it : zipTree(it) } }
```
dehall, line 163: Update to the previous comment, this seems to work based on https://stackoverflow.com/a/4894308 and a quick local test

```suggestion
    from { configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) } }
```

hadleynet: I confirmed the issue and your suggested fix. Thanks!

PR #826, opened 2020-12-08
MEND Program Module

RobertScalfani:
This Pull Request adds a module to simulate the MEND (Mind, Exercise, Nutrition, Do It!) program, described here - http://www.mendfoundation.org. MEND is one of the world's largest evidence-based healthy lifestyle programs that empowers children, with the support of their families, to reach and maintain a healthy weight. While the program is targeted at 7 - 13 year olds, it is applied more widely in practice. This module allows children between 5 - 17 to participate in the program. This module depends on attributes set by Synthea's  java `WeightLossModule`. It adds and uses the `mend_program.json` module as well as the `weight_loss/mend_week.json` and `encounter/vitals.json` submodules.

### DIFF
--- a/src/main/resources/modules/encounter/vitals.json
+++ b/src/main/resources/modules/encounter/vitals.json
@@ -6,7 +6,6 @@
   "states": {
     "Initial": {
       "type": "Observation",
-      "target_encounter": "current_encounter",
       "category": "vital-signs",
       "vital_sign": "Height",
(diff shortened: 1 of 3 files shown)


eedrummer (approved):
While the typo fix at the end isn't necessary, it doesn't really impact this PR

hadleynet (approved):

PR #831, opened 2020-12-10
Migrate to GitHub Actions

dehall:
Migrates the CI process from Travis CI to GitHub Actions.

This uses 2 "workflows" to achieve what we currently do today with Travis.
1. A "Build & Test" workflow to check push/pull requests on all branches:
   - the app compiles correctly on the latest JDK (15 as of now; uses the `gradle assemble` task)
   - all tests pass on JDK 8 (uses the `gradle check` task) then checks code coverage with CodeCov
   - both of these use caches to speed up compile time
   - status is reported to our Slack
2. A "Deploy" workflow which deletes and re-creates the `master-branch-latest` tag, and deploys the following built artifacts, only on pushes to the master branch:
   - uberJar (to the `master-branch-latest` release)
   - javadocs (to gh-pages)
   - graphviz (to gh-pages)
   - if one of the steps fails, an error is reported to our Slack

(note step 2 uses a potentially sketchy Action to delete the old tag, but so far I haven't found many better options. I'm open to suggestions)


hadleynet (approved):
This looks great. Anything need to be done before we merge?

dehall:
I don't think we need to do anything else - this should remove every reference to Travis (which btw, travis-ci.org is going away in favor of .com soon anyway, according to a banner at the top of every page) so we'll be all set

PR #836, opened 2020-12-18
CI Build Fix

hadleynet:
- Temporarily disable testYears test
- Simplify test module (only used to check reading modules from different location)
- Turn off export.split_records for generator tests that don't explicitly test that option


dehall:
There are still some intermittent failures, in particular the FHIR R4 test seems to fail a lot with memory issues, but this fixes the persistent issues. I'm going to merge this, and continue to investigate the memory issues. (Maybe the R5 HAPI JARs use a lot more memory for resource validation?)

dehall (approved):

PR #842, opened 2021-01-11
testYears Fix

ak27842:
The specific test of focus was the testYears test from the UtilitiesTest class.  Due to the existence of leap years, the previous designation of a year as 365 days led to the test returning the incorrect year when traversing back 75 years, occurring from 12/15 to the end of the year.  To change this, a year was changed to be 365.25 days.  

Below is a table depicting the days this test will fail based on the days/year scalar:
Days/Year | When testYears Test Fails
365 | 12/13 0:00 - 12:31 23:59
365.2 | 12/28 0:00 - 12:31 23:59
365.22 | 12/29 0:00 - 12:31 23:59
365.24 | 12/31 0:00 - 12:31 23:59
365.25 | 12/31 0:00 – 12:31 23:59
The current value for this scalar is 365.25.

### DIFF
--- a/src/main/java/org/mitre/synthea/helpers/Utilities.java
+++ b/src/main/java/org/mitre/synthea/helpers/Utilities.java
@@ -43,7 +43,7 @@ public class Utilities {
       case "days":
         return TimeUnit.DAYS.toMillis(value);
       case "years":
-        return TimeUnit.DAYS.toMillis((long) (365.24 * value));
+        return TimeUnit.DAYS.toMillis((long) (365.25 * value));
       case "months":
         return TimeUnit.DAYS.toMillis(30 * value);
       case "weeks":
--- a/src/main/java/org/mitre/synthea/helpers/Utilities.java
+++ b/src/main/java/org/mitre/synthea/helpers/Utilities.java
@@ -43,7 +43,7 @@ public class Utilities {
       case "days":
         return TimeUnit.DAYS.toMillis(value);
       case "years":
-        return TimeUnit.DAYS.toMillis(365 * value);
+        return TimeUnit.DAYS.toMillis((long) (365.24 * value));
       case "months":
         return TimeUnit.DAYS.toMillis(30 * value);
       case "weeks":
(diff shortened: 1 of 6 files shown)


dehall:
To elaborate a little more - the current approach of having a "quick & dirty" time conversion method means that there are always going to be gaps somewhere when compared to the rigorous but slower built-in Java time handlers. 

Setting the "days/year" in this function to 365.25 causes a couple tests around the HealthInsuranceModule to fail. That module is designed to process things on a monthly and yearly basis, but expects to be called on the usual timestep, ie one week, so it's able to be called more often than intended with no downside. On the other hand, it's not designed to "catch up" if it's not called in a given time frame.

The last piece of this is that the `Utilities.convertCalendarYearsToTime` function is expected to return 1/1/XXXX at 00:00, ie the first millisecond of the year.  So if the calculation is off by one millisecond then it returns 12/31 of the previous year. Then for 3/4 years (ie, non-leap years) 12/31/X + 365.25 days = 1/1/(X+2), ie, a whole year in the middle gets skipped. And for the leap years, 1/1/X + 365.25 = 12/31/X, it's not the next year yet. So in some cases, things will fail because they got counted twice. Some things will fail because they didn't happen at all.
I proposed 2 options for a fix to Aditya:
1) Always run the HealthInsuranceModule tests on a weekly basis over the intended timeframe, to mirror how they are actually called in practice.
2) Shift the start date, so that instead of having testing dates right around the year boundary, they happen at a time where the leap year math doesn't matter.

hadleynet:
I was wondering if we could just pick a start date for the testYears function rather than starting with the current date - is that what you are proposing in 2?

dehall:
Not what I had in mind, but that's an interesting approach. I was thinking we should change the days/year to 365.25, then fix whatever falls out from that. Never considered the alternative of just picking a specific start date in the one test that fails.
Either way we're changing how a test works because what it's testing has edge cases. Not sure there's any way around that...

hadleynet:
Fixed by #853 instead.

PR #845, opened 2021-01-22
Include/Exclude for CSV exporter files

dehall:
Adds configuration options for included and excluded files in the CSV exporter. Since many use cases don't require every single CSV file, we can save some space and computing time by not exporting them in the first place. (I'm looking at you, observations.csv...)

If you specify a list of files in `exporter.csv.included_files`, then only those files will be exported. (and if you don't include `patients.csv` in the list, it will yell at you but still process as expected)
If you specify a list of files in `exporter.csv.included_files`, then all other files will be exported and the named files will not be.
If you specify both config options, it will ignore both and do the default, export every file.

The basic approach here is to use a "no-op" implementation of an OutputStream, so that we only have to change the initialization of the file writers, not how they are called. This does mean though that the string content of the CSV is still created, but then thrown away. Since these work one line at a time it shouldn't mean a whole lot in terms of memory usage, but I can revisit if people think it's an issue.

Another minor change here is to move the actual initialization of the CSVExporter to an `init()` method so that it can be re-initialized if needed, for example in unit tests with different config settings

### DIFF
--- a/src/main/java/org/mitre/synthea/export/CSVExporter.java
+++ b/src/main/java/org/mitre/synthea/export/CSVExporter.java
@@ -15,12 +15,16 @@ import java.math.RoundingMode;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
+import org.apache.commons.io.output.NullOutputStream;
 import org.apache.commons.lang3.StringUtils;
 import org.mitre.synthea.helpers.Config;
 import org.mitre.synthea.helpers.RandomCodeGenerator;
@@ -121,6 +125,8 @@ public class CSVExporter {
    * Writer for payerTransitions.csv
    */
   private OutputStreamWriter payerTransitions;
+  
+  public static final int NUMBER_OF_FILES = 16;
 
   /**
    * Charset for specifying the character set of the output files.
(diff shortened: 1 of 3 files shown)


src/main/java/org/mitre/synthea/export/CSVExporter.java
```diff
@@ -121,6 +125,8 @@
    * Writer for payerTransitions.csv
    */
   private OutputStreamWriter payerTransitions;
+  
+  public static final int NUMBER_OF_FILES = 16;
```
eedrummer, line 129: This is only used by the test suite. Suggest moving it to `CSVExporterTest`.

src/main/java/org/mitre/synthea/export/CSVExporter.java
```diff
@@ -150,58 +160,80 @@ private CSVExporter() {
...
+            System.err.println("WARNING! CSV exporter is set to not include patients.csv!");
+            System.err.println("This is probably not what you want!");
+          }
+
+        } else {
+          List<String> files = Arrays.asList(excludedFilesStr.split(","));
+          // normalize filenames -- trim, lowercase, add .csv if not included
+          files = files.stream().map(f -> f.trim().toLowerCase())
+              .map(f -> f.endsWith(".csv") ? f : f + ".csv").collect(Collectors.toList());
+          excludedFiles.addAll(files);
```
eedrummer, line 191: This is essentially the same a lines 175 - 179. You could just create a lambda to encapsulate the repeated functionality right in the method.

src/main/java/org/mitre/synthea/export/CSVExporter.java
```diff
@@ -150,58 +160,80 @@ private CSVExporter() {
...
+        }
+      }
 
       File patientsFile = outputDirectory.resolve("patients.csv").toFile();
-      boolean append =
-          patientsFile.exists() && Config.getAsBoolean("exporter.csv.append_mode");
-      patients = new OutputStreamWriter(new FileOutputStream(patientsFile, append), charset);
+      boolean append = patientsFile.exists() && Config.getAsBoolean("exporter.csv.append_mode");
```
eedrummer, line 196: Probably an odd corner case, but if someone excluded the patient file, but wants to append to other files, it would over write, because it can't find the patient file.

I think `patientsFile.exists() &&` can be dropped because `getWriter` does `append = append && file.exists()`

eedrummer (commented):
Some minor nits.

src/main/java/org/mitre/synthea/export/CSVExporter.java
dehall: I tried it out and lambdas in Java are just ugly enough that I went for a full helper method instead

PR #848, opened 2021-01-25
Hypertension medication module: fix medication tagged as condition and remove redundant logic

rutemartins:
* Fixed transition bug in a couple of conditional transitions: changed `Condition Active` to `Medication Active` on amlodipine.
* Eliminated redundant logic for proteinuria and microalbuminuria (which should put the patient on lisinopril 10 MG Oral Tablet regardless of whether they're on other meds) for state `Step 2 High`.

### DIFF
--- a/src/main/resources/modules/medications/hypertension_medication.json
+++ b/src/main/resources/modules/medications/hypertension_medication.json
@@ -221,34 +221,19 @@
       "complex_transition": [
         {
           "condition": {
-            "condition_type": "And",
+            "condition_type": "Or",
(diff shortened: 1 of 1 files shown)


dehall (approved):
It took me a long time to figure out what was going on in this module; for some reason there are a lot of seemingly duplicate states that only vary in naming convention, ex "Step 2 High" vs "Step_2_High". But they aren't exact duplicates, they do vary in certain places. We should probably fix that sometime, but it's outside the scope of this PR

PR #849, opened 2021-01-27
Even more assorted bugfixes

dehall:
This PR intends to fix a few issues, mostly in unit tests, relating to memory usage and weird exceptions. This gets pretty involved so please don't hesitate to ask if the fixes aren't clear.

Note: when looking at the diff I highly recommend using "ignore whitespace changes" as the 3 FHIR exporter test diffs are mostly indentation changes

1. Occasionally I've seen `NotSerializableException`s in `GeneratorTest.testUpdateAfterCreationAndSerialization` because of FixedRecordGroup not being Serializable. At a glance this doesn't make sense because this test doesn't use fixed records. But deeper analysis revealed that our approach with `clone()` methods left some dangling references. 
   - First issue is that we aren't even using the `clone()` method correctly. The top-level `Object.clone()` already does a shallow copy of all fields on the object (see https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#clone-- ), so it's totally unnecessary to do it manually afterward. Instead we should make sure any object references or true "state" like when the State was entered/exited are explicitly nulled out. For the most part, there's nothing wrong with manually shallow copying fields to the clone, but I changed every State subclass to remove them anyway so we don't continue to use the unnecessary practice.  (Many say `clone()` is a bad practice by itself, but I'm not as worried about that for now)
   - So with that in mind, the one place where the shallow copy is a problem is on `ObservationGroup.observations`.  When the ObservationGroup (for example a DiagnosticReport state) gets processed, it calls `process` on each sub-observation, and that sets true state on the sub-observation, specifically the `entry` field. So what wound up happening is every instance of DiagnosticReport states shared the same instances of the subobservations, and things were stomping over each other all the time. I suspect there are a lot of ways this caused weirdness, but it was only really exposed in that unit test - the FixedRecordGroupTest ran and created a patient that had a FixedRecordGroup in their attributes, then later this test created a patient, and through a long chain of object references they linked back to that old patient. 
     - An actual example of a sequence of fields that got serialized:
person -> attributes -> list of one module's history -> state object -> module object -> list of all states in that module -> a diagnostic report state -> list of observations -> one observation -> healthrecord entry -> healthrecord -> encounters list -> one encounter -> claim -> person (not the same person we started with!) -> ...
---
2. `OutOfMemoryError`s tend to happen on the FHIRR4ExporterTest, and this is also one of the longest-running tests we have. The basic approach the 3 FHIR exporter tests take is:
   - for 1..10, generate a new patient
   - export the patient as FHIR JSON
   - parse the JSON back into a Resource
   - validate the Resource
   - if the Resource isn't valid (has errors), re-parse it as a Bundle, then validate each Resource in the Bundle individually, to check for known issues

It turns out that the way the FHIR validator works is to convert a Resource into a (XML) string, then parse that into an object tree of Elements. That parsing and the subsequent size of the tree is where OutOfMemoryErrors happen. (And when it's not an OOM, it often takes a long time because of memory pressure)
So my suggestion is to skip the validation of the Bundle as a whole, and just validate the individual Resources. For any patients where the bundle doesn't pass initial validation, this is definitely faster because it's just removing that first pass. For patients where the initial pass does succeed, I think this will probably be roughly the same amount of time, because the full bundle validation has to iterate over every resource anyway.  The main downside of this approach would be that it doesn't validate references within the Bundle, but actually I don't think the validator checks that as it is today. 

---
3. Some memory usage minutiae: 
   - Cleared out the static list Exporter.deferredExports once the records are actually exported, so they can be GC'ed
   - Changed FixedRecordTest.generator to be an instance variable instead of a static, again so it can be GC'ed

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -1559,9 +1559,9 @@ public abstract class State implements Cloneable, Serializable {
       // we need to ensure we deep clone the list
       // (otherwise as this gets passed around, the same objects are used for different patients
       // which causes weird and unexpected results)
-      List<Observation> cloneObs = new ArrayList<>(observations);
-      for (int i = 0; i < cloneObs.size(); i++) {
-        cloneObs.set(i, cloneObs.get(i).clone());
+      List<Observation> cloneObs = new ArrayList<>(observations.size());
+      for (Observation o : observations) {
+        cloneObs.add(o.clone());
       }
       clone.observations = cloneObs;
       
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -1554,7 +1554,6 @@ public abstract class State implements Cloneable, Serializable {
 
     public ObservationGroup clone() {
       ObservationGroup clone = (ObservationGroup) super.clone();
-      clone.codes = codes;
       
       // IMPORTANT: because each observation gets process()ed when the state gets processed,
       // we need to ensure we deep clone the list
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -121,14 +121,17 @@ public abstract class State implements Cloneable, Serializable {
    * clone() should copy all the necessary variables of this State so that it can be correctly
    * executed and modified without altering the original copy. So for example, 'entered' and
    * 'exited' times should not be copied so the clone can be cleanly executed.
+   * Implementation note: the base Object.clone() copies over all fields automatically
+   * (as a shallow copy), so we don't need to do that ourselves. Instead, we should 
+   * 1. explicitly null out any fields that should not be copied, such as entered/exited
+   * 2. deep copy mutable reference types, if necessary.
    */
   public State clone() {
     try {
       State clone = (State) super.clone();
-      clone.module = this.module;
-      clone.name = this.name;
-      clone.transition = this.transition;
-      clone.remarks = this.remarks;
+      clone.entered = null;
+      clone.exited = null;
+      clone.entry = null;
       return clone;
     } catch (CloneNotSupportedException e) {
       // should not happen, and not something we can handle
@@ -232,7 +235,6 @@ public abstract class State implements Cloneable, Serializable {
     @Override
     public CallSubmodule clone() {
       CallSubmodule clone = (CallSubmodule) super.clone();
-      clone.submodule = submodule;
       return clone;
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -590,14 +588,6 @@ protected void initialize(Module module, String name, JsonObject definition) {
     @Override
     public SetAttribute clone() {
       SetAttribute clone = (SetAttribute) super.clone();
```
hadleynet, line 590: Based on the comment in the original should we be setting threadExpProcessor to null?

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1462,13 +1414,6 @@ public void processOnce(Person person, long time) {
     @Override
     public VitalSign clone() {
       VitalSign clone = (VitalSign) super.clone();
```
hadleynet, line 1416: Same question re threadExpProcessor. I won't raise this again but if this is a valid change then apply everywhere else it is needed.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1622,8 +1554,17 @@ public boolean process(Person person, long time) {
...
     public ObservationGroup clone() {
       ObservationGroup clone = (ObservationGroup) super.clone();
-      clone.codes = codes;
-      clone.observations = observations;
+      
+      // IMPORTANT: because each observation gets process()ed when the state gets processed,
+      // we need to ensure we deep clone the list
+      // (otherwise as this gets passed around, the same objects are used for different patients
+      // which causes weird and unexpected results)
+      List<Observation> cloneObs = new ArrayList<>(observations);
```
hadleynet, line 1562: Should this be new ArrayList<>(observations.size()) instead?

hadleynet (changes requested):

src/main/java/org/mitre/synthea/engine/State.java
dehall: Good catch - I meant to explicitly call that out but forgot. Actually the original comment was wrong. We definitely don't want to copy thread-specific variables, but the whole point of the `ThreadLocal` class is that it *is* safe to pass around between threads, because it's a wrapper around the object, and it uses an internal map of Thread->object to ensure that each thread only sees its own version of the object.

Leaving it as-is means that the underlying Object.clone is copying the field and the object is passed around, just as it was before. It won't break anything if we do set it to null, but I think the intent of using the ThreadLocal in the first place was that we want to reuse the underlying object because it's expensive to create, but the libraries it uses aren't threadsafe, so we only want to use it within the same thread. So if we set `threadExpProcessor = null` in clone(), I don't think there's any reason to use ThreadLocal at all. I looked at this in detail a few months ago, trying to make CQL processing less resource-intensive, but I don't remember the details. I think it's worth revisiting, though my vote would be to leave this as is today

src/main/java/org/mitre/synthea/engine/State.java
dehall: Yeah I think I had chosen the complicated way to implement deep copying. Updated

hadleynet (approved):

PR #856, opened 2021-02-12
Remove internal H2 database and associated features

dehall:
The internal H2 database is a nice idea, but in practice it's too slow and resource-heavy to be useful. Our own documentation says not to use it: 
image

This PR removes the DB entirely, along with the 3 reports that could be generated if it was enabled:
 - Prevalence Report
 - Cost/Access/Outcomes Report
 - Custom SQL Report

Other than us using the prevalence report for some internal projects, where we're now moving over to using the CSV export, I don't think anyone has used any of these in the past few years. (The custom SQL report is actually broken for certain use cases because it runs a toUpperCase on the provided commands, which means you can't query by lower case attributes, for example.)

There's no other direct functionality impact here, but since this is a feature removal, I think we can hold off a while on merging this, especially if someone says they are using the database for something. 

Do we think anyone is using the DB? / Is there any other way we should announce this change?


jawalonoski:
I agree that this feature can be safely removed.

To the community -- if you use this feature, please raise your concerns here.

jawalonoski (approved):

PR #857, opened 2021-02-17
Hk 2 bb export

hkong:
work in progress --- initial commit for ExportConfig for bb2rif export using TSV file; beneficiary, beneficiary_history, inpatient, and outpatient are workng; carrier and prescription not yet complete


hkong:
I'm opening a new and updated PR that branches from `bb-export`

PR #860, opened 2021-02-19
Hk 4 bb export

hkong:
ExportConfig for bb2rif export using resources/exporters/cms_field_values.tsv file; beneficiary, beneficiary_history, inpatient, outpatient, carrier, and prescription are all working.


src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -173,35 +177,47 @@ final void prepareOutputFiles() throws IOException {
...
   
   /**
    * Export a beneficiary details for single person.
    * @param person the person to export
    * @param stopTime end time of simulation
+   * @param useConfig flag to use ExportConfig; use false to use original code, true to use ExportConfig and ExportBuilder
    * @throws IOException if something goes wrong
    */
-  private void exportBeneficiary(Person person, long stopTime) throws IOException {
+  private void exportBeneficiary(Person person, long stopTime, boolean useConfig) throws IOException {
```
hadleynet, line 196: I'm not sure how useful the useConfig parameter will be in the future. Are there cases where we wouldn't want to use the config file?

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -798,13 +864,76 @@ private void exportPrescription(Person person, long stopTime) throws IOException
...
+
+  public interface GetEnumValueInterface {
+    public int getEnumValue( String field );
+  }
+
+
+  /**
+   * Exporter Data Builder
+   */
+  public class ExportDataBuilder {
```
hadleynet, line 883: Seems like this class could be safely merged with ExportConfig so that all of the TSV processing and resulting information is encapsulated in one class?

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -1146,11 +1275,13 @@ private String capitalizeWords(String str) {
     }
   }
 
+  interface AbstractFields {}
```
hadleynet, line 1278: What's the purpose of this interface? If it is just a marker couldn't you use java.lang.Enum instead?

hadleynet (changes requested):

src/main/java/org/mitre/synthea/export/ExportDataBuilder.java
```diff
@@ -0,0 +1,71 @@
+// package org.mitre.synthea.export;
```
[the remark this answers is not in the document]
hkong: that was an error.  The content was moved to BB2RIFExporter.java, but I accidentally committed the original.  Removed.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hkong: All of these are for regression testing (I run with the config set to false to get the original output, then set it to true to compare the new output using the config TSV file).  I like to keep these for one more PR, until we no longer need these regression tests.  These, by the way, are the bulk of the coverage test errors.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hkong: You're right!  I'll do it with ExportConfigEntry as well.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hkong: It's just so that they have a common ancestor.  I'll try java.lang.Enum.

hkong:
refactored code based on changes suggested by @hadleynet.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hkong: refactored code based on comments

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -798,13 +864,19 @@ private void exportPrescription(Person person, long stopTime) throws IOException
...
-    //    prescription.flush();
+    outpatient.flush();
+    carrier.flush();
+    prescription.flush();
   }
 
+
+  public interface GetEnumValueInterface {
+    public int getEnumValue( String field );
+  }
```
jawalonoski, line 877: What is this? It doesn't seem to be used anywhere that I can see.

src/main/java/org/mitre/synthea/export/BFDExportBuilder.java
```diff
@@ -0,0 +1,167 @@
...
+  // private List<BFDExportConfigEntry> hospiceConfigs = new ArrayList<BFDExportConfigEntry>();
+  // private List<BFDExportConfigEntry> medicareBeneficiaryIdConfigs = new ArrayList<BFDExportConfigEntry>();
+  private List<BFDExportConfigEntry> outpatientConfigs = new ArrayList<BFDExportConfigEntry>();
+  // private List<BFDExportConfigEntry> pdeConfigs = new ArrayList<BFDExportConfigEntry>();
+  private List<BFDExportConfigEntry> prescriptionConfigs = new ArrayList<BFDExportConfigEntry>();
+  // private List<BFDExportConfigEntry> snfConfigs = new ArrayList<BFDExportConfigEntry>();
+  
+  /** constructor */
+  public BFDExportBuilder() {
+    this.configFile = new File( "src/main/resources/exporters/cms_field_values.tsv" );
```
jawalonoski, line 56: Should probably get this as a resource. This won't work in the JAR release.

src/main/java/org/mitre/synthea/export/BFDExportConfigEntry.java
```diff
@@ -0,0 +1,94 @@
...
+  private String beneficiary_history;  
+  private String inpatient;
+  private String outpatient;
+  private String carrier;
+  private String prescription;
+
+  /** trims any newlines, tabs, spaces from value 
+   *  @param value the string to evaluate for newlines, tabs, etc.
+  */
+  public String trimAllWhitespace( String value ) {
```
jawalonoski, line 21: Should probably be private.

jawalonoski: And since it is a single line of code -- not sure why you even need the function whatsoever.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
jawalonoski: I still don't understand this.

jawalonoski (commented):
There is stuff in here I'm confused by, but it mostly makes sense. I didn't run it through checkStyle, but I'm sure it will have a lot of complaints.

As far as I can tell, all it does now is add fixed values -- correct?

jawalonoski:
FYI, I just attempted to fix the merge conflicts with GitHubs web interface, which generated this `Merge branch 'bb-export' into hk_4_bb_export` commit.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hkong: nice catch, I'll delete it

src/main/java/org/mitre/synthea/export/BFDExportConfigEntry.java
hkong: I may need to extend this.  It used to use something that is not available in Java 8.  I like to keep this until the next PR.

src/main/java/org/mitre/synthea/export/BFDExportBuilder.java
hkong: > There is stuff in here I'm confused by, but it mostly makes sense. I didn't run it through checkStyle, but I'm sure it will have a lot of complaints.
> 
> As far as I can tell, all it does now is add fixed values -- correct?

In this PR, that is correct.  In the new work in hk_5_bb_export branch, it does distributions and other things (like encounter dates as a start)

PR #862, opened 2021-03-01
Added HL7 ADT generation to generate HL7V2 test data

qpointsystems:
I think this is the 2nd time I've written this code over the years and I think this turned out halfway decent -- for my purposes anyways. This request here is also less of a real Pull Request as it is a ping to the Mothership as to whether this type of addition might be of interest to the community of Synthea users. If so I could make the effort to better follow project guidelines/style/etc, spiff it up a bit and put up a real pull request. Either way appreciate any feedback. 

I currently generate HL7V2 'fat' ADT messages and CDA, side by side, for comparison once the data is imported into my larger project. It's quite a useful process.  It's not a perfect match but nice to have some HL7V2 test data.

Also, can't thank those behind this project enough.  Synthea has been helpful over the years.  Cheers.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/HL7V2Exporter.java
+++ b/src/main/java/org/mitre/synthea/export/HL7V2Exporter.java
@@ -2,6 +2,8 @@ package org.mitre.synthea.export;
 
 import ca.uhn.hl7v2.HL7Exception;
 import ca.uhn.hl7v2.model.DataTypeException;
+import ca.uhn.hl7v2.model.GenericGroup;
+import ca.uhn.hl7v2.model.Group;
 import ca.uhn.hl7v2.model.v251.datatype.CE;
 import ca.uhn.hl7v2.model.v251.datatype.CWE;
 import ca.uhn.hl7v2.model.v251.datatype.CX;
@@ -16,12 +18,14 @@ import ca.uhn.hl7v2.model.v251.datatype.XPN;
 import ca.uhn.hl7v2.model.v251.datatype.XTN;
 import ca.uhn.hl7v2.model.v251.group.ADT_A01_INSURANCE;
 import ca.uhn.hl7v2.model.v251.group.ADT_A01_PROCEDURE;
+import ca.uhn.hl7v2.model.v251.group.ORM_O01_ORDER;
 import ca.uhn.hl7v2.model.v251.message.ADT_A01;
 import ca.uhn.hl7v2.model.v251.segment.AL1;
 import ca.uhn.hl7v2.model.v251.segment.DG1;
 import ca.uhn.hl7v2.model.v251.segment.EVN;
 import ca.uhn.hl7v2.model.v251.segment.IN1;
 import ca.uhn.hl7v2.model.v251.segment.MSH;
+import ca.uhn.hl7v2.model.v251.segment.OBR;
 import ca.uhn.hl7v2.model.v251.segment.OBX;
 import ca.uhn.hl7v2.model.v251.segment.PD1;
 import ca.uhn.hl7v2.model.v251.segment.PID;
@@ -71,9 +75,6 @@ public class HL7V2Exporter {
     }
 
     public String export(Person person, long time) {
-        // create a super encounter... this makes it easier to access
-        // all the Allergies (for example) in the export templates,
-        // instead of having to iterate through all the encounters.
         Encounter superEncounter = person.record.new Encounter(time, "super");
         for (Encounter encounter : person.record.encounters) {
             if (encounter.start <= time) {
@@ -91,10 +92,8 @@ public class HL7V2Exporter {
             }
         }
 
-        // The export templates fill in the record by accessing the attributes
-        // of the Person, so we add a few attributes just for the purposes of export.
         person.attributes.put("UUID", UUID.randomUUID().toString());
-        person.attributes.put("ehr_encounters", person.record.encounters);
+        person.attributes.put("ehr_encounters", person.record.encounters);        
         person.attributes.put("ehr_observations", superEncounter.observations);
         person.attributes.put("ehr_reports", superEncounter.reports);
         person.attributes.put("ehr_conditions", superEncounter.conditions);
@@ -189,7 +188,6 @@ public class HL7V2Exporter {
         PID pid = adt.getPID();
         pid.getPid1_SetIDPID().setValue("1");
         Map<String, Object> pattrs = person.attributes;
-//        System.out.println("\tGenerating PID: " + pattrs.get("name"));
 
         XPN patientName = pid.insertPatientName(0);
         patientName.getFamilyName().getSurname().setValue(getStrAttr(pattrs, "last_name"));
@@ -253,9 +251,8 @@ public class HL7V2Exporter {
     }
 
     private void processEncounters(Person person) throws DataTypeException, HL7Exception {
-        List<HealthRecord.Encounter> encounters = (List<HealthRecord.Encounter>) person.attributes.get("ehr_encounters");
+        List<HealthRecord.Encounter> encounters = (List<HealthRecord.Encounter>) person.record.encounters;
         Encounter e = encounters.get(0);
-//        System.out.println("\tGenerating PV1: " + e.name);   
         PV1 pv1 = adt.getPV1();
         pv1.getPv11_SetIDPV1().setValue("1");
         IS eClass = pv1.getPv12_PatientClass();
@@ -333,15 +330,13 @@ public class HL7V2Exporter {
         Integer ac = 0;
         List<String> seenAllergenCodes = new ArrayList();
         for (Entry entry : allergies) {
-//            System.out.println("\tGenerating AL1: " + entry.name);
             AL1 a = new AL1(adt, adt.getModelClassFactory());
             a.getAl11_SetIDAL1().setValue(String.valueOf(ac + 1));
             if (entry.codes != null && entry.codes.size() > 0) {
                 Code c = entry.codes.get(0);
                 if (seenAllergenCodes.contains(c.code)) {
-//                    System.out.println("\t\tSkipping Dup Allergy:" + c.display);
+                    //Skip it
                 } else {
-//                    System.out.println("\t\tAdding Allergy:" + c.display);
                     seenAllergenCodes.add(c.code);
                     a.getAl12_AllergenTypeCode().getCe1_Identifier().setValue(c.code);
                     a.getAl12_AllergenTypeCode().getCe2_Text().setValue(c.display);
@@ -360,14 +355,12 @@ public class HL7V2Exporter {
         Integer pc = 0;
         List<String> seenConditionCode = new ArrayList();
         for (Entry entry : conditions) {
-//            System.out.println("\tGenerating PRB: " + entry.name);
             PRB p = new PRB(adt, adt.getModelClassFactory());
             if (entry.codes != null && entry.codes.size() > 0 && StringUtils.containsIgnoreCase(entry.codes.get(0).display, "finding")) {
                 Code c = entry.codes.get(0);
                 if (seenConditionCode.contains(c.code)) {
-//                    System.out.println("\t\tSkipping Dup Diag:" + c.display);
+                    //Skip it
                 } else {
-//                    System.out.println("\t\tAdding Diag:" + c.display);
                     seenConditionCode.add(c.code);
                     p.getProblemID().getCe1_Identifier().setValue(c.code);
                     p.getProblemID().getCe2_Text().setValue(c.display);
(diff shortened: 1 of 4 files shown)


shabiel:
+1

jawalonoski:
@qpointsystems Thank you for the pull request.

I do think this capability would be interesting to the broader development community. I've anecdotally heard of others creating HL7 v2 messages as well using Synthea, so there must be an unmet need that this PR could fill.

As you humbly acknowledge, there could be some clean up. In particular there seems to be some copy & paste unnecessary code in your exporter (for example, setting some of the attributes that are used by the Apache Freemarker template-based exporters).

I do think this could be a nice addition to the project.

qpointsystems:
@jawalonoski, thanks a lot for your comments! I'd like to basically do a clean pull request that just includes  the following:

  * HL7V2Exporter.java as the initial version of a HL7V2 exporter 
  * Code added to Exporter.java to optionally invoke HL7V2Exporter if the flag from synthea.properties is set
  * synthea.properties to add the **exporter.hl7v2.export** property
  * build.gradle for the HAPI 2.51 dependency

The version of HAPI I've chosen is definitely debatable. Another version like 2.7 or even 2.8 might be more appropriate for some and not others. Ideally a user could set the version they wanted.  But I was trying to avoid adding a lot of dependencies directly to the project.  There are ~10 versions of HAPI aside from 'hapi-base'.   Definitely a bit of a conundrum. 

Regarding my Copy&Paste habit -- guilty as charged of course. But could you please point me to a place in HL7V2Exporter where I'm setting some Freemarker aspects?  Definitely agree that, since my current exporter is not doing any templating, I shouldn't be implying with code that it is. I'd like to nix that for sure. I'm sure once you point an example out I'll get what you mean.

Finally: for my next attempt, do you want it in the form of a pull request, even if it's just to facilitate feedback?  

Cheers!

Gary Teichrow

src/main/java/org/mitre/synthea/export/HL7V2Exporter.java
```diff
@@ -0,0 +1,626 @@
...
+        person.attributes.put("ehr_allergies", superEncounter.allergies);
+        person.attributes.put("ehr_procedures", superEncounter.procedures);
+        person.attributes.put("ehr_immunizations", superEncounter.immunizations);
+        person.attributes.put("ehr_medications", superEncounter.medications);
+        person.attributes.put("ehr_careplans", superEncounter.careplans);
+        person.attributes.put("ehr_imaging_studies", superEncounter.imagingStudies);
+        person.attributes.put("time", time);
+        person.attributes.put("race_lookup", RaceAndEthnicity.LOOK_UP_CDC_RACE);
+        person.attributes.put("ethnicity_lookup", RaceAndEthnicity.LOOK_UP_CDC_ETHNICITY_CODE);
+        person.attributes.put("ethnicity_display_lookup", RaceAndEthnicity.LOOK_UP_CDC_ETHNICITY_DISPLAY);
```
jawalonoski, line 110: I'm fairly certain that this big block of code is all related to the Freemarker exporters. They use these attributes in the templates, but you don't need that here.

qpointsystems: Perfect! All that code can go. I should obviously be using **superEncounter** and **person.record.encounters**, etc. I mean, it's all right there in memory (the longitudinal model that is).

qpointsystems:
One thought I had as I was writing this HL7 V2 Exporter is this: my exporter is clearly very use-case specific, frankly a bit of a kludge, and unlikely to be of much use as is. It's specific to my use-case which is generating large volumes of test data for a messaging pipeline and it's been amazingly helpful to my project.  Synthea has been a tremendous help to my efforts -- can't say that enough.

I spent a good chunk of time on a different approach before writing this exporter.  That was attempting to take CCDA's generated by Synthea and converting them into one or more HL7 V2 messages. I basically gave up, but it was an interesting exercise. 

The **Exporter aspect in Synthea**, which has access to the entire patient health graph that's been generated in memory, is simply the perfect place to differentiate that graph out to some use-case specific format (CCDA, FHIR,  HL7, ????, etc) as esoteric as that use-case might be. 

Currently the only way to add in some new ExportXXXX is to _splice_ a call to that new functionality into Exporter.java, add any appropriate dependencies, and finally a configuration option to toggle it.  Easy peasy, but what if nobody really needs/wants that ExportXXXX functionality?  There is no current mechanism to, via some wiring configuration, cause Synthea to invoke some Exporter that implements some _Exporter Interface_ from a separate library. Much looser coupling to the larger efforts of the project but still leverages the KEY thing that Synthea does which is building these amazing patient health graphs of realistic clinical data. 

I suggest the question here is how could this kind of functionality be added to Synthea without dragging in a lot of unnecessary complexity (e.g. Spring, etc). 

I think my effort at this point should be to suggest/implement a mechanism for this sort of functionality and **not** to add my far to use-case specific exporter to the greater project.

dehall:
@qpointsystems to your point of generalizing exporters - this is something I've been thinking a lot about recently as well, and experimented around a bit with a couple approaches.

What I would propose as a first step is to create a generic interface for exporters, and a mechanism for these to be registered with the Exporter class at runtime. Nothing complex at this point like dependency injection(Spring), just a way in code to say `Exporter.registerExporter(....)` . Then a developer who wants to add a new exporter would provide their own `main` method, that registers an exporter and then invokes the existing Synthea main method. Eventually, but not a high priority, we could consider migrating the existing exporters to use that interface as well.

As you note, anything that wires Synthea to be aware of arbitrary additional libraries is going to either be pretty heavyweight, or not compatible with existing aspects of the project, so I wouldn't recommend starting there. (For example I tried reading jars and scanning for classes implementing interfaces with "javautil": https://github.com/bmc/javautil / http://software.clapper.org/javautil/api/org/clapper/util/classutil/ClassFinder.html  but it's incompatible with another library we're using, I forget which)

hadleynet:
Use of `java.util.ServiceLoader` looks like a possibility. Essentially you could package up a new exporter as a jar with an appropriate `META-INF/services` file that declares support for a Synthea-defined exporter interface. We'd have to add something to Synthea to use `ServiceLoader` to look for new exporters on the class path at runtime.

qpointsystems:
> Use of `java.util.ServiceLoader` looks like a possibility. Essentially you could package up a new exporter as a jar with an appropriate `META-INF/services` file that declares support for a Synthea-defined exporter interface. We'd have to add something to Synthea to use `ServiceLoader` to look for new exporters on the class path at runtime.

Very cool. Nice and simple -- KISS principle at work.

So perhaps there is an **exporter** sub-directory where these implementation jars go. Then Synthea shortly after startup scans that directory for jars with an appropriate manifest and have a class that implements a corresponding IExporter interface AND is further called out/named in synthea.properties (we shouldn't load them all, just once that the user wants to use).  Each loaded exporter implementation in then is 'invoked' for each patient.  

The output folder name might need to be configurable but no reason that aspect could not be interrogated from the implementation of the particular exporter.  Along with some other attributes perhaps:

String export(Person person, long time);
String getExportPath() or getExportPath()  (e.g., "ccda", "hl7v2", etc)

jawalonoski:
> Use of `java.util.ServiceLoader` looks like a possibility. Essentially you could package up a new exporter as a jar with an appropriate `META-INF/services` file that declares support for a Synthea-defined exporter interface. We'd have to add something to Synthea to use `ServiceLoader` to look for new exporters on the class path at runtime.

Let's keep that separate from the v2 pull request. That will be a larger refactor that I would like to keep separate. We can open an issue for that if we want to track that.

qpointsystems:
I think we can close this pull. I'm quite pleased with the notion of adding a more flexible Export capability -- perhaps via the ServiceLoader notion mentioned above. Thanks for the feedback on my code and, when there is a more 'modular' Export functionality, i'll make my code (FWIW) available via a github project.  As with your ability to add disease states, I having an ability to support custom export formats might be fairly useful.

Cheers!

dehall:
Not to reignite this discussion, just for posterity: 

Don't ask what reminded me of this, but I just remembered what the issue with ServiceLoader would be -- the classpath and the different ways that we can run Synthea. There are (at least) 2 ways to launch Synthea, either via gradle (`gradle run` / the `run_synthea` script), or by the standalone jar `java -jar synthea-with-dependencies.jar`.

Now for the ServiceLoader to find our "module" jar, it has to be on the classpath. For the jar version this is easy, though we have to rewrite our run command: `java -cp synthea-with-dependencies.jar:module.jar App` . I think I tested something like this and it did work as expected. But for running from gradle/run_synthea, there's no easy way to add to the classpath. You can add new Classloaders that read from external JARs, but when you start mixing classloaders you can get all sorts of trouble so I'm not sure that's a road we want to go down.

Dmytro-Hasilin:
Hi @qpointsystems ,

What happened over time with this ADT exporter?

gteichrow:
@Dmytro-Hasilin, well, I used the capability back in '21 to pressure test a part of my system that supports HL7V2.  

For that purpose it worked out nicely. I honestly don't think it'd meet the grade for inclusion in this project;  a bit of a hack.  

That said, for my purposes where I required syntactically valid HL7 ADT messages that were content wise the same as the CDA messages Synthea was generating (i.e., contained the same demographic and clinical information in a different format), I got what I needed. 

Going forward, when I periodically update Synthea to the latest master, I then re-merge my work.  So far It's been easy to merge.  

I've you'd like a 'stash' of the code I'd be more than happy to provide.

g

Dmytro-Hasilin:
@gteichrow  got it. I've copy pasted and merged with master this PR and for my simple use case it also worked well. 
Its a pity that this effort is unused. I would definitely use it in future, so thanks. 
I believe it is worth adding to repo as a preview feature 👍

dehall:
Hi @gteichrow and @Dmytro-Hasilin , fyi we did at some point in the past few years enable the ServiceLoader approach I mentioned above to allow for custom drop-in JARs - more info is here: https://github.com/synthetichealth/synthea/wiki/Custom-Exporters and https://github.com/synthetichealth/custom-exporter-template
As far as I know, nobody outside of our code development group has used it yet so there may be quirks, but we'd love for someone to try it out for something like this

Dmytro-Hasilin:
@dehall  I would love to contribute, this project saved me a lot of time. But @gteichrow is not sure this code is not sure this PR is up to the standards to contribute, and I understand only on high level what is done here. 

I could wrap it in the jar. But does that adds value? Would it be approved?
I would say yes, it is better that keeping it as code 'stash', it produces way better output than random adt messages from google search results.

dehall:
Yeah, your update is what I was thinking - this approach would allow you to build a jar from a separate repo and use that in synthea without any additional code changes. So you could update synthea without having to reapply/rebase changes, you could share it with people who don't want to make code changes, etc.

qpointsystems:
I'm currently on holiday, but when I get back I can port/refactor my existing functionality to one of these ServiceLoader compatible 'plugin' jars.  

But, I do have an integration question: 

At this point I don't think there is an interface I'd implement so that my 'plugin' can implement?  (maybe it's there?)

What I did with my changes is, I wrote my **HL7V2Exporter** and it has a single public method:
public String export(Person person, long time);

Then in **Exporter.java**, where the various configuration flags like **'exporter.*.export'** are evaluated, I added this chunk of code:

    //Caterwaul
    if (Config.getAsBoolean("exporter.hl7v2.export")) {
      String hl7V2Msg = HL7V2Exporter.getInstance().export(person, stopTime);
      File outDirectory = getOutputFolder("hl7v2", person);
      Path outFilePath = outDirectory.toPath().resolve(filename(person, fileTag, "hl7"));
      writeNewFile(outFilePath, hl7V2Msg);
    }

Proposal: if services could be examined for that service's implementation of the interface IExporterService and expose one ore more methods (e.g., initialize, getConfigFlag, export(person), shutdown) in that interface, particularly the configuration property that correlates to this Service (e.g., getConfigFlag() returns 'exporter.hl7v2.export') that would allow Exporter.java to 'bind' a given exporter Service IExporterService to a given configuration property.

This would turn Exporter.exportRecord() into a method that **iterates** through 3 or more invocations of exporter IExporterService.export() methods and let the services control all that logic.

Anyways, I bring this up as it is one thing to move my HL7V2Exporter exporter to a ServiceLoader compatible jar.  It's another to wire the invocation of my service into the larger program; specifically a call in. Exporter.export().

(I'll be checking email on vacation like a good geek, but can't get away with coding, haha)

dehall:
The interface does exist, it's `org.mitre.synthea.export.PatientExporter` with one method: `public void export(Person person, long stopTime, ExporterRuntimeOptions options)`
It's not quite as guided as your proposed interface, but it allows pretty much unlimited flexibility within that export method. For example it doesn't assume the output will be written to a file - our demo writes to a postgres database. And it's called for every patient record so it's up to the exporter itself to check if it should run.

So you will need to migrate some file write logic and the config check into the exporter itself, but other than that it should hopefully be a straightforward copy&paste of your existing code. If you get a chance, take a look at https://github.com/synthetichealth/synthea/wiki/Custom-Exporters

gteichrow:
That's fabulous. Thanks for the comments and looking forward to porting my code.  

:)

PR #870, opened 2021-03-17
CSV Claims

jawalonoski:
Work in progress for Claims and claim Transactions in CSV.


hadleynet (approved):
Looks good to me. Ran locally and CSV and CPCDS output files look reasonable.

I did notice that CPCDS claim amounts have a lot of decimal places and should probably be rounded to two decimal digits. Not sure if that is a new issue or not.

Longer term I suggest we migrate all of the CSV writing code to use the approach we developed for BFD. The enum-value-per-column-based approach is less error prone and self documenting.

PR #873, opened 2021-03-25
SDoH Updates

jawalonoski:
This PR will have several SDoH updates.

- The first update is a demographic change which accounts for Native Hawaiian and Pacific Islanders in the population. The approximation used here matches the expected value, where roughly 10% of the population of Hawaii identified as Native Hawaiian and Pacific Islander in US Census in 2000, 2010, and 2015 (est), and the other states are significantly less.
- There is an update to "Distributions" that adds a "round" property which will round the resulting random number to be a whole number. This is useful where observations should be integer values (e.g. "how many people live in your household?") versus a floating point.
- This PR also allows `SetAttribute` to take a code as a value, and that attribute can then be used as the value for an Observation. This eliminates duplication of huge diagnostic report states, where the only differences are Observation values that differ by answer codes.
- This PR adds a set of Social Determinant of Health attributes to patients. These attributes can be used in modules to determine pathways or outcomes. At the present, they are used solely for responding to Health Related Social Needs screenings.
- The PR contains a number of other screenings (e.g. anxiety, depression) that may (probabilistically) be executed as submodules from a wellness encounter. The SDoH screening is executed in the same manner.

The SDoH file is county-level data. It is derived from the 2018 Social Vulnerability Index and the 2021 Community Health Rankings. The file has the following headers:

```
FIPS_CODE,COUNTY_CODE,COUNTY,ST,STATE,FOOD_INSECURITY,SEVERE_HOUSING_COST_BURDEN,UNEMPLOYED,NO_VEHICLE_ACCESS,UNINSURED
```

The first five columns (`FIPS_CODE,COUNTY_CODE,COUNTY,ST,STATE`) are required and fixed.

The remaining columns (`FOOD_INSECURITY,SEVERE_HOUSING_COST_BURDEN,UNEMPLOYED,NO_VEHICLE_ACCESS,UNINSURED`), plus any others that might be added in the future, get added to the patient attributes based on the probability listed in the table cell.


src/main/java/org/mitre/synthea/world/concepts/RaceAndEthnicity.java
```diff
@@ -5,56 +5,20 @@
...
   private static Map<String, String> createCDCRaceLookup() {
     Map<String, String> result = new HashMap<String, String>();
     result.put("white", "2106-3"); // white
-    result.put("hispanic", "2106-3"); // white
+    result.put("hispanic", "2106-3"); // hispanic
     result.put("black", "2054-5"); // black
     result.put("asian", "2028-9"); // asian
     result.put("native", "1002-5"); // american indian or alaska native
-    result.put("other", "2076-8"); // native hawaiian or pacific islander
+    result.put("hawaiian", "2076-8"); // native hawaiian or pacific islander
```
eedrummer, line 20: Adding a new code to `CDCRaceLookup` has a cascading effect on `Demographics.languageFromRaceAndEthnicity`. It works, in that there is a `switch` statement and individuals with the Hawaiian race code will hit the default block, which returns English as the language spoken.

This may be the best we can do for now. For other race / ethnicity combinations, we used data from the census to determine the percentage of individuals who state that they do no speak English well. For that slice of individuals, we select languages based on census data for languages spoken by race and ethnicity.

eedrummer (approved):
Looks good. Minor issues that aren't show stoppers.

See the comment about generating languages.

Also, the new SDoH file has minor issues with the existing demographics file:
* There are 158 entries in the demographics file with no SDoH info
* There are 13 entries in the SDoH file that will never be used because there is no corresponding entry in demographics.

src/main/java/org/mitre/synthea/world/concepts/RaceAndEthnicity.java
jawalonoski: Regarding the misalignment between demographics and the SDoH file... I'm not sure I can do anything about that, if the county files I used do not have data. I either lose SDoH data, or I lose Demographic data. The way the code should work here is that demographic data takes precedence. If the SDoH data is absent, then (unfortunately) the people in those counties do not get any of those attributes. Not ideal, but everything continues to function.

jawalonoski: Regarding language... good catch, I'll fix that.

PR #879, opened 2021-04-27
Odds and Ends

dehall:
This PR includes a mishmash of various fixes and tweaks pulled from multiple active working branches, in order to reduce the changes left on those branches to what's really important, and ensure these fixes are available throughout.
Where possible, these commits were cherry-picked from other branches to help git figure out how to merge and rebase later.

Details of changes:

 - `src/main/java/Graphviz.java`: fix a bug in rendering graphviz for modules that use the "Vital Sign" type logic 
 - `src/main/java/org/mitre/synthea/engine/State.java`: 
   1.  allow calling Java modules via the "CallSubmodule" state, and ensure that module history isn't duplicated (this caused an issue when a Java module was called repeatedly and the module history grows exponentially)
   2. fix a bug with Delays occuring inside submodules, where upon returning from the submodule, the parent module wasn't aware of any "time rewinding" that may have occurred. The approach here is to set the exited time of the CallSubmodule state based on the exited time of the previous state from history.
   3. require units on observations with numeric values (if the unit is missing it breaks CCDA)
 - `src/main/java/org/mitre/synthea/engine/Transition.java`: fix a bug involving null attributes in table based transitions
 - `src/main/java/org/mitre/synthea/helpers/Utilities.java`: add a function `convertRiskToTimestep` to allow converting risk between arbitrary timesteps (not just the currently configured `generate.timestep`)
 - `src/main/java/org/mitre/synthea/modules/LifecycleModule.java`, `src/main/java/org/mitre/synthea/world/agents/Provider.java`, `src/main/java/org/mitre/synthea/world/concepts/Names.java`:  remove an implicit and unintended link between the Provider class and the modules by extracting the logic to pick names to a new class. Previously, loading providers would call `LifecycleModule.fakeFirstName` etc, which would require loading the Module class, which spins up all the modules. In most cases there will be no difference, but this can help prevent some weird bugs in certain situations.
 - `src/main/resources/modules/hypothyroidism.json`: fix a range with low/high flipped
 - `src/main/resources/modules/sepsis.json`: replace VitalSign states that were setting Systolic and Diastolic blood pressure, and inadvertently overriding the built-in BloodPressureValueGenerators for blood pressure, causing a patient's BP to fluctuate wildly between 40-120 for the rest of their life. Now the blood pressure observation here uses the original range from the module, but doesn't override the underlying BPVG. Also adds another reading of blood pressure using the patient's regular vital signs, per clinical feedback.

More to come in future PRs:
 - Review a few tweaks made to exporters to support different classes of observation values (booleans, HealthRecord.entrys, etc) -- these work fine in FHIR but CCDA is trickier
 - Break out the metabolic syndrome standards of care module into multiple submodules. No functional changes intended, just for readability and maintainability

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -14,6 +14,8 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.math.ode.DerivativeException;
 import org.mitre.synthea.engine.Components.Attachment;
@@ -1603,6 +1605,26 @@ public abstract class State implements Cloneable, Serializable {
     @Override
     protected void initialize(Module module, String name, JsonObject definition) {
       super.initialize(module, name, definition);
+      validate(module, name);
+    }
+
+    protected void validate(Module module, String name) {
+      if (exact != null || range != null) {
+        // units are required
+        if (unit == null || unit.isEmpty()) {
+          throw new RuntimeException(
+              "Observations with numeric quantities must contain units matching pattern [^\\s]+: "
+          + "Module \"" + module.name + "\": State \"" + name + "\": Unit missing or empty");
+        }
+
+        String regex = "[^\\s]+";
+        if (!Pattern.matches(regex, unit)) {
+          throw new RuntimeException(
+              "Observations with numeric quantities must contain units matching pattern [^\\s]+: "
+          + "Module \"" + module.name + "\": State \"" + name + "\": Unit \"" + unit + "\"");
+        }
+      }
+
       if (distribution != null && !distribution.validate()) {
         throw new IllegalStateException(
             String.format("State %s contains an invalid distribution", this.name));
@@ -1686,6 +1708,14 @@ public abstract class State implements Cloneable, Serializable {
   private abstract static class ObservationGroup extends State {
     protected List<Code> codes;
     protected List<Observation> observations;
+
+    @Override
+    protected void initialize(Module module, String name, JsonObject definition) {
+      super.initialize(module, name, definition);
+      for (Observation observation : observations) {
+        observation.validate(module, name);
+      }
+    }
 
     public ObservationGroup clone() {
       ObservationGroup clone = (ObservationGroup) super.clone();
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -185,9 +185,13 @@ public abstract class State implements Cloneable, Serializable {
       // to indicate when the state actually completed.
       if (this instanceof Delayable) {
         this.exited = ((Delayable)this).next;
+      } else if (this instanceof CallSubmodule) {
+        this.exited = ((CallSubmodule)this).submoduleExited;
       } else {
         this.exited = time;
       }
+    } else if (this instanceof Terminal) {
+      this.exited = time;
     }
 
     return exit;
@@ -231,6 +235,7 @@ public abstract class State implements Cloneable, Serializable {
    */
   public static class CallSubmodule extends State {
     private String submodule;
+    private transient long submoduleExited;
 
     @Override
     public CallSubmodule clone() {
@@ -250,6 +255,14 @@ public abstract class State implements Cloneable, Serializable {
       boolean completed = submod.process(person, time);
 
       if (completed) {
+        // keep track of when the submodule exited, in case it was "rewinding time" when it completed
+        if (person.history.get(0).exited == null) {
+          // temporary hack. figure out why we need this for the unit tests
+          this.submoduleExited = time;
+        } else {
+          this.submoduleExited = person.history.get(0).exited;
+        }
+        
         // add the history from the submodule to this module's history, at the front
         if (moduleHistory != person.history) {
           // if the submodule is a java module, it didn't create its own history
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -257,7 +257,8 @@ public abstract class State implements Cloneable, Serializable {
       if (completed) {
         // keep track of when the submodule exited, in case it was "rewinding time" when it completed
         if (person.history.get(0).exited == null) {
-          // temporary hack. figure out why we need this for the unit tests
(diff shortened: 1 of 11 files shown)


src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -61,6 +62,9 @@
       Config.get("generate.providers.selection_behavior", "nearest").toLowerCase();
   private static IProviderFinder providerFinder = buildProviderFinder();
 
+  public static final boolean appendNumbersToNames =
+      Config.getAsBoolean("generate.append_numbers_to_person_names", false);
+
```
jawalonoski, line 67: This should probably just be inside the new `Names` class.

jawalonoski:
Metabolic syndrome module refactoring should probably be a separate PR.

src/main/java/org/mitre/synthea/world/agents/Provider.java
dehall: Every instance of calling `fakeFirstName` or `fakeLastName` also checked this boolean immediately after, so I moved this and the subsequent check and adding numbers into those 2 methods.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -1586,6 +1605,26 @@ public boolean process(Person person, long time) {
...
+    protected void validate(Module module, String name) {
+      if (exact != null || range != null) {
+        // units are required
+        if (unit == null || unit.isEmpty()) {
+          throw new RuntimeException(
+              "Observations with numeric quantities must contain units matching pattern [^\\s]+: "
+          + "Module \"" + module.name + "\": State \"" + name + "\": Unit missing or empty");
+        }
+
+        String regex = "[^\\s]+";
```
eedrummer, line 1620: @dehall Where did the regex come from? I'm having a hard time understanding exactly what it is trying to do. Is it essentially ensuring that the string is not blank?

dehall: That's from this commit on `cabg_cardio_gmf`: https://github.com/synthetichealth/synthea/commit/22c116bdada0024b087db61025a194a24b7a8d02

Looking at it now though it is a roundabout way of doing things. `\s` means any whitespace character, the double backslash is just for java string escaping, `^` means negate, the brackets `[]` mean "anything from this group", but I'm not sure it's meaningful when the one thing inside it is already a group, and `+` means one or more. So this will match anything that is one or more non-whitespace characters. 
It's a little roundabout because it should be equivalent to `"\\S+"` where the `\S` class means non-whitespace. 
( https://docs.oracle.com/javase/tutorial/essential/regex/pre_char_classes.html )
If we were using Java 11 we could use `String.isBlank` instead here, but that's not in 8. I can change this to use Apache `StringUtils.isBlank` if that's more clear. We already use StringUtils in a few other places so it's not a whole new library just for this

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -250,8 +257,20 @@ public boolean process(Person person, long time) {
       boolean completed = submod.process(person, time);
 
       if (completed) {
+        // keep track of when the submodule exited, in case it was "rewinding time" when it completed
```
eedrummer, line 260: I know this is silly, but the line length violates the check style rules.

src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -250,8 +257,20 @@ public boolean process(Person person, long time) {
...
+        if (person.history.get(0).exited == null) {
+          // this happens when the patient dies in the submodule, so processing is going to stop anyway
```
eedrummer, line 262: Another line length issue.

src/main/java/org/mitre/synthea/engine/Transition.java
```diff
@@ -283,13 +283,13 @@ public String follow(Person person, long time) {
...
-          String personsAttribute
-              = (String) person.attributes.get(currentAttribute.toLowerCase());
-          if (personsAttribute == null) {
+          if (!person.attributes.containsKey(currentAttribute)) {
             throw new RuntimeException("LOOKUP TABLE ERROR: Attribute '"
                 + currentAttribute + "' in CSV table '" + this.lookupTableName
                 + "' does not exist as one of this person's attributes.");
           }
+          String personsAttribute
+              = person.attributes.get(currentAttribute).toString();
```
eedrummer, line 292: Question on using `toString()` here... Is this the behavior we want?

It could be "dangerous" as an attribute that is not a `String` in the `attributes` map, but some other object, could get "flattened" into a `String`. Would it be better to cast to `String` and potentially crash if the object is not a `String`?

eedrummer (changes requested):
Very minor check style issues.

Some questions, but might not require any changes.

src/main/java/org/mitre/synthea/engine/Transition.java
dehall: This is what I get for pulling in features that I didn't write myself. 😁  As I look closer at this now, I think I misread where things would be different before vs after.  (This was the original commit on `cabg_cardio_gmf` btw: https://github.com/synthetichealth/synthea/commit/8f2a471347982462d0d0659ce3abdfc335df64f3)

The real difference is the removal of the `toLowerCase()` on the attribute name. There's not much changed on nulls, because previously the explicit null check would throw an exception, now a NullPointerException will be thrown on the `toString`. I don't mind the "flattening to a string" aspect too much, because I would expect it to be useful in cases like booleans and integers. Seems like Jay was using booleans in lookup tables in that original commit. So at this point I think the options are a) leave the toString (or a_ii is to replace it with `String.valueOf()` to intentionally convert `null` into `"null"`), or b) consider a list of types that are allowed to be here, and crash if it's not one of those types. Then that immediately raises the question of should floating point numbers be allowed? Comparing floats as strings sounds like a bad idea, but I think in many cases we use `double`s to be more generic, but they actually contain a whole number.

I feel like Jay's preference is to generally not crash. I would lean toward leaving the toString more because it might allow future things to work with fewer code changes.

jawalonoski: Yes, Booleans in the table-based transition table. Casting the attribute as a (String) was causing exceptions.

And, in the end, the attribute needs to be a String, because of how the LookupTableKey class works. 

This change intended to make the table-based transitions function for boolean data, without refactoring the entire implementation (which I didn't have time for).

eedrummer: Makes sense.

PR #880, opened 2021-04-27
Womens health analysis - WIP 

carrils:
To reflect most effective long-acting reversible contraceptives, in documentation https://www.cdc.gov/reproductivehealth/unintendedpregnancy/pdf/contraceptive_methods_508.pdf. 

General information on contraceptives by age can be found here https://www.cdc.gov/nchs/data/databriefs/db173_table.pdf#1
Pregnancy rates get more realistic with commit 5fb92a0 yet still need further modifications. 

Changes are guided by results of an analysis jupyter notebook, ran with synthea output.

### DIFF
--- a/src/main/resources/modules/contraceptives.json
+++ b/src/main/resources/modules/contraceptives.json
@@ -234,11 +234,11 @@
           "transition": "Using_Withdrawal"
         },
         {
-          "distribution": 0.117,
+          "distribution": 0.0,
           "transition": "Using_IUD"
         },
         {
-          "distribution": 0.00,
+          "distribution": 0.117,
           "transition": "Using_Implant"
         },
         {
--- a/src/main/resources/modules/contraceptives.json
+++ b/src/main/resources/modules/contraceptives.json
@@ -234,11 +234,11 @@
           "transition": "Using_Withdrawal"
         },
         {
-          "distribution": 0.07,
+          "distribution": 0.117,
           "transition": "Using_IUD"
         },
         {
-          "distribution": 0.047,
+          "distribution": 0.00,
           "transition": "Using_Implant"
         },
         {


carrils:
This PR attempts to retune certain module probability values to address previous concerns that rates may be too high or unrealistic. The rates are very high when compared to reference statistical information. The method used to guide changes to the probability distribution values is as follows.
	1. Using verified NCHIS and CDC information, compare what matches reference information and what does not.
	2. Implement this change to the probability distribution value to model a correction or a theory. 
	3. Take .csv file output from synthea and process in a jupyter notebook, this notebook is used throughout to obtain metrics and statistics.
	4. Determine whether the modification helped or hurt the accuracy of Synthea's output. 

There are sometimes a small population that gets their age calculated incorrectly when looking at metrics, this is only seen when the age is above 50. Presumably, this is caused by something rooted in the matching logic (nested for's) that looks at PIDS from one csv file to the other to calculate ages metric in the jupyter notebook.  For some reason the date in the conditions.csv is read in as the current time, which would make the age calculation inaccurate. 

Populations generated with default property values (other than enable csv exports) and:
`./run_synthea -g F -p 1000`


### Unmodified
image
[<= 19]: 85
[<= 24]: 155
[<= 29]: 203
[<= 34]: 168
[<= 39]: 136
[<= 44]: 93
[<= 50]: 96
Pregnancies: 
1479
Total number of patients (deceased + living): 
1121


### Adjusting  probability of not using contraceptive
image
Target rates from: https://www.cdc.gov/nchs/data/hestat/pregnancy/pregnancy_fig1.png
[<= 19]: 87, should be 58.9
[<= 24]: 150, should be 144.6
[<= 29]: 200, should be 157.1
[<= 34]: 147, should be 136.5
[<= 39]: 145, should be 76.5
[<= 44]: 107, should be 19.4
[<= 50]: 109
Pregnancies: 
1727
Total number of patients (deceased + living if generate dead patients is on): 
1195

[contraceptives.json]
"transition": "Using_No_Contraceptive"
0.283 -> 0.386
First I tried upping the value for not using contraceptives even though we are trying to get less pregnancies since the reference material said a greater percentage of women are not using contraceptives. This predictably made our pregnancy rate issue even worse than it is already so I reverted that. 



### Testing an 'extreme' 
image
Target rates from: https://www.cdc.gov/nchs/data/hestat/pregnancy/pregnancy_fig1.png
[<= 19]: 86, should be 58.9
[<= 24]: 141, should be 144.6
[<= 29]: 127, should be 157.1
[<= 34]: 107, should be 136.5
[<= 39]: 84, should be 76.5
[<= 44]: 107, should be 19.4
[<= 50]: 101
Pregnancies: 
1251
Total number of patients (deceased + living if generate dead patients is on): 
1128

[contraceptives.json]
"transition":"Using_IUD"
0.07 -> 0.283
"transition": "Using_No_Contraceptive"
0.283 -> 0.07
This was more so done to ascertain whether or not the rates I was looking at can be brought down. This ended up taking our rate and decreasing it by too much so this was reverted. 



### Shift probability value from IUD to Implant 
image
Target rates from: https://www.cdc.gov/nchs/data/hestat/pregnancy/pregnancy_fig1.png
[<= 19]: 76, should be 58.9
[<= 24]: 156, should be 144.6
[<= 29]: 175, should be 157.1
[<= 34]: 224, should be 136.5
[<= 39]: 151, should be 76.5
[<= 44]: 139, should be 19.4
[<= 50]: 97
Pregnancies: 
1673
Total number of patients (deceased + living if generate dead patients is on): 
1137

[contraceptives.json]
"transition": "Using_Implant"
0.07 -> 0.117
Long-acting reversible contraceptives include implants and intrauterine devices, so shift probability from implant into IUD since from the literature they are grouped together anyway. Implant specifically is the most effective form of contraception. Right now Long-acting reversible contraceptives are used at a rate of 11.7% in middle age group (25-34).  This was the change that only made improvements to the output that also made sense logically, so this was kept.

eedrummer:
@carrils Where do we stand with this PR? It looks like the current version flips two distribution probabilities, but that results in more pregnancies than the unmodified version. Do we want to try the "extreme" version?

PR #881, opened 2021-04-29
Add ability to select patients to keep/export using a GMF module

dehall:
This PR introduces the ability to specify a special GMF module that determines criteria for a patient to be exported.

Because the GMF doesn't have the concept of "returning a value" explicitly, this takes the approach of using a named state to determine whether a patient record should be kept or not. (I don't love magic names, but we already require the "Initial" state name.) If a patient finishes in a Terminal state named "Keep" then the record is kept, otherwise the record is discarded, the seed rotated, and another record is generated and tested.
(Another possible approach, if people don't like requiring a specific name, would be to set a specific attribute to true/false)

Use the feature with `./run_synthea -k (path_to_module_file)`

Note that this uses a reduced version of Module.process to run through the module. The "keep module" should primarily include Initial, Terminal, and Simple states, with transitions using whatever logic to select patients. Use of other state types may produce unexpected results.

Two sample "keep modules":

1. Keep patients with Attribute: diabetes == true
image
[keep.json.txt](https://github.com/synthetichealth/synthea/files/6401701/keep.json.txt)

2. Keep patients that have an active diagnosis of either diabetes or hypertension:
image
[keep.json.txt](https://github.com/synthetichealth/synthea/files/6401708/keep.json.txt)

In the future we may want to add some kind of "population attributes" to allow for filtering in the context of the population vs just the individual.

⚠️  IMPORTANT ⚠️ 
It is very easy to build a "keep module" that only keeps impossible patients. (For instance, no patient under the age of 18 can get diabetes in synthea. Trying to generating a population with ages 1-100 where every patient has diabetes will never finish, because if it tries to generate a patient under 18 it will never succeed and keep looping forever as it tries with a different seed. To successfully generate a dataset of "only diabetes", make sure the age range is at least 18+)
Use at your own risk. Make sure the demographics settings that you are using align with the patient characteristics you are expecting.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Generator.java
+++ b/src/main/java/org/mitre/synthea/engine/Generator.java
@@ -71,6 +71,7 @@ public class Generator implements RandomNumberGenerator {
   private boolean onlyDeadPatients;
   private boolean onlyVeterans;
   private Module keepPatientsModule;
+  private Long maxAttemptsToKeepPatient;
   public TransitionMetrics metrics;
   public static String DEFAULT_STATE = "Massachusetts";
   private Exporter.ExporterRuntimeOptions exporterRuntimeOptions;
@@ -214,6 +215,18 @@ public class Generator implements RandomNumberGenerator {
       Config.set("generate.only_alive_patients", "false");
       this.onlyDeadPatients = false;
       this.onlyAlivePatients = false;
+    }
+
+    try {
+      this.maxAttemptsToKeepPatient = Long.parseLong(
+        Config.get("generate.max_attempts_to_keep_patient", "1000"));
+
+      if (this.maxAttemptsToKeepPatient == 0) {
+        // set it to null to make the check more clear
+        this.maxAttemptsToKeepPatient = null;
+      }
+    } catch (Exception e) {
+      this.maxAttemptsToKeepPatient = null;
     }
 
     this.onlyVeterans = Config.getAsBoolean("generate.veteran_population_override");
@@ -442,6 +455,7 @@ public class Generator implements RandomNumberGenerator {
       boolean patientMeetsCriteria;
 
       do {
+        tryNumber++;
         person = createPerson(personSeed, demoAttributes);
         long finishTime = person.lastUpdated + timestep;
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -519,12 +514,15 @@ public Person generatePerson(int index, long personSeed) {
...
    * If a patient does not meet the criteria the process will be repeated so a new one is generated
+   * @param person the patient to check if we want to export them
+   * @param finishTime the time simulation finished
    * @param isAlive Whether the patient is alive at end of simulation.
    * @param providerCount Number of providers in the patient's record
    * @param providerMinimum Minimum number of providers required
    * @return true if patient meets criteria, false otherwise
    */
-  public boolean patientMeetsCriteria(boolean isAlive, int providerCount, int providerMinimum) {
+  public boolean patientMeetsCriteria(Person person, long finishTime, boolean isAlive,
```
eedrummer, line 524: This is fine and maybe I think too much in OO, but I'm wondering if we can clean things up by introducing a class for the result of a patient simulation. Call it `SimulationResult` and it can have the person, finish time, whether they are alive and the provider count.

The `patientMeetsCriteria` could be moved to that class and get passed the options for the overall simulation or work with `Config`. Right now, things like `onlyDeadPatients` get set as fields on `Generator`, but are really only used in `patientMeetsCriteria`, outside of checking that `onlyDeadPatients` and `onlyAlivePatients` conflict.

eedrummer (commented):
Some suggestions on code changes.

src/main/java/org/mitre/synthea/engine/Generator.java
dehall: Where this gets tricky is in all the various methods that return a Person object. And in `Person generatePerson(int index, long personSeed)` it's extra tricky because that function generates many but only returns one. 
I think you're probably right, we can clean things up (do we really need all those different functions?) but the delta to do that will be nontrivial.
I've made a couple additional tweaks to reduce the number of variables that get passed around, so hopefully it seems a little cleaner now.

carrils (commented):
Hey Dylan, Looks very good! Not sure I can offer much here in the way of changes/improvements, maybe a sanity check in patientMeetsCriteria? I would be willing to help or work on this as well, I am just unsure of technical difficulty to get that done.

hadleynet (approved):

dehall:
Just pushed a new setting per @carrils 's suggestion. It will attempt to generate a matching patient up to a configurable number of tries (default 1000), then if it fails it will throw an exception with a long message to explain why. Hopefully the message is clear enough to reduce the number of questions we get about this feature, since I think this is something a lot of people have asked for and may use.

dehall:
Update: I just discovered an issue while testing this merged onto cabg_cardio_gmf. The `if (!isAlive && !onlyDeadPatients && this.options.overflow)` case in `patientMeetsCriteria` now loops with a new seed, where previously it didn't. The effect of this is that with default settings it won't produce dead patients anymore.
Need to think of a clean way to fix this.

PR #899, opened 2021-06-11
Support for de/serializing Random class in GSON

### LINKED ISSUE #897: GSON InaccessibleObjectException / Illegal Access exceptions while running Synthea with JDK 16
I installed JDK 8 and then later 16.0.1 on Mac OS and encountered Null pointer exceptions when attempting to run Synthea (both in development and basic setup modes).
`java.lang.reflect.InaccessibleObjectException: Unable to make field private final java.util.concurrent.atomic.AtomicLong java.util.Random.seed accessible: module java.base does not "opens java.util" to unnamed module @40f9161a`

`caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private final java.util.concurrent.atomic.AtomicLong java.util.Random.seed accessible: module java.base does not "opens java.util" to unnamed module @40f9161a`

I installed JDK 15.0.2 and it worked as expected. Just reporting in case if it's not a known issue.

dehall:
This PR fixes #897 and makes Synthea work in Java 16 by introducing a new TypeAdapter for GSON, to allow serializing and deserializing classes that are in core Java packages. (ie, `java.*.*`). In practice it seems like the only core class that we try to serialize is java.util.Random, but I made this generic so that if we need to add more in the future this should make it easy. The basic approach is to leverage the existing java Serializable functionality to write the object to a string, then base64 it. (It turns out for the `java.util.Random` class all the relevant fields are private, so there aren't many other options to serialize it)

Note that I think the only time this will actually get called is when someone does population snapshotting. But we need the TypeAdapter present for loading all the GMF modules, because GSON walks the entire class hierarchy starting from State, and via State.entry you can get to Person.random. If the TypeAdapter for Random isn't present, it will try to create its own via reflection which is no longer possible.

Also I bumped up some of the relevant library versions here just in case that might help too

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ jacoco {
 
 dependencies {
   // This dependency is found on compile classpath of this component and consumers.
-  implementation 'com.google.code.gson:gson:2.8.6'
+  implementation 'com.google.code.gson:gson:2.8.7'
   implementation 'com.jayway.jsonpath:json-path:2.4.0'
   implementation 'ca.uhn.hapi.fhir:hapi-fhir-base:5.2.0'
   implementation 'ca.uhn.hapi.fhir:hapi-fhir-structures-dstu3:5.2.0'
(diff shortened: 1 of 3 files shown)


jawalonoski (approved):
Tested with Java 8 and Java 16. Also tested snapshot capability to make sure it continued to function there.

PR #915, opened 2021-07-06
Metadata exporter

dehall:
Currently in the FHIR exporter we include patient seed, population seed, and Synthea version, but these are not available in any other data formats. Rather than add them individually to the other formats, this metadata exporter outputs info about the most recent run as a JSON file.

Notes:
 - I set the new config setting to default to true, because it's a lot better to have it and not need it, than to create a large CSV population and forget the seed you used when you want to recreate it. If people disagree I'm fine disabling it by default though
 - Each run gets a separate metadata file, just to minimize the chance of weirdness in case people have multiple instances running at the same time.
 - The filename is `{timestamp} _ {population size} _ {state} _ { city? } _ {run ID}` to make it more immediately apparent what goes to what, and ensure uniqueness

From the slack channel, 2 earlier suggestions I haven't yet added are gradle version (there doesn't seem to be a trivial way to do it programmatically, so I'll see what out there makes sense) and total run time. 

Any suggestions on additional fields we may want to add here?

### DIFF
--- a/src/main/java/App.java
+++ b/src/main/java/App.java
@@ -76,7 +76,10 @@ public class App {
             options.clinicianSeed = Long.parseLong(value);
           } else if (currArg.equalsIgnoreCase("-r")) {
             String value = argsQ.poll();
-            SimpleDateFormat format = new SimpleDateFormat("YYYYMMDD");
+            // note that Y = "week year" and y = "year" per the formatting guidelines
+            // and D = "day in year" and d = "day in month", so what we actually want is yyyyMMdd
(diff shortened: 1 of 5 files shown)


dehall:
Note that in updating how the reference date is exported, I discovered a bug in the parsing of the reference date on the command line. It was using `YYYYMMDD` instead of `yyyyMMdd` which can produce surprising results: https://ideone.com/88QamU
So that is also fixed here

hadleynet (approved):

Issue #909, opened 2021-06-28
Patient.communication.language sometimes contains invalid system


emesar:
Per the FHIR R4 spec for the Patient resource, the [Patient.communication.language](https://www.hl7.org/fhir/patient-definitions.html#Patient.communication.language) element has a "preferred" binding with the [Common Languages](https://www.hl7.org/fhir/valueset-languages.html) ValueSet, but a "required" binding with the [All Languages](https://www.hl7.org/fhir/valueset-all-languages.html) ValueSet. Both ValueSets use the following system URI exclusively: `urn:ietf:bcp:47`.

However, there are 5 languages provided in [Synthea's language lookup file](https://github.com/synthetichealth/synthea/blob/master/src/main/resources/language_lookup.json) which instead provide `http://id.loc.gov/vocabulary/iso639-1` as the system URI. An example generated by Synthea:

```json
"communication": [{
    "language": {
        "coding": [{
            "system": "http://id.loc.gov/vocabulary/iso639-1",
            "code": "vi",
            "display": "Vietnamese"
        }],
        "text": "Vietnamese"
    }
}]
```

Can these be changed to use the correct system URI (`urn:ietf:bcp:47`)?

jawalonoski:
Technically bcp47 is not a Code System, it is a ValueSet.

It does not define any of its own codes, instead it includes codes from other systems, e.g. iso639-1 which defines the "vi" code (among others).

I don't care if we update the language lookup file so that FHIR validates (assuming that it will actually fix FHIR validation), but the `system` in that example is technically correct (and if FHIR says it is invalid, then FHIR is wrong).

Feel free to submit a pull request with the changes, and steps to verify the changes are correct and validate properly.

emesar:
@jawalonoski thanks for the quick response. After researching this a bit further, I see your point; BCP-47 doesn't define the actual meaning of the various language codes, just a best practice structure.

FHIR is *definitely* treating BCP-47 as a CodeSystem, though; it's provided as such in the [HL7 Code System Registry](https://www.hl7.org/fhir/terminologies-systems.html).

I've asked for some clarity in the [FHIR Zulip](https://chat.fhir.org/#narrow/stream/179202-terminology/topic/BCP-47/near/244213266).

jawalonoski:
Well, given the resolution over at FHIR Zulip, I'm happy to take a pull request to fix this issue if you want to make a contribution.

If we don't get one in a few days we'll get around to correcting it.

[...]

PR #916, opened 2021-07-07
Update language lookup to always use BCP: 47

emesar:
Per #909, FHIR requires the use of the system `urn:ietf:bcp:47` in the [Patient.communication.language](https://www.hl7.org/fhir/patient-definitions.html#Patient.communication.language) element. A handful of languages (5) in language_lookup.json currently use `http://id.loc.gov/vocabulary/iso639-1` as the system URI, which raises a validation error in some FHIR server implementations. 

[BCP: 47, Section 2.2.1](https://tools.ietf.org/search/bcp47#section-2.2.1) designates ISO 639 language subtags as conforming with BCP: 47. As such, the only change made is to update the systems of the 5 language entries to `urn:ietf:bcp:47`.

### DIFF
--- a/src/main/resources/language_lookup.json
+++ b/src/main/resources/language_lookup.json
@@ -51,7 +51,7 @@
   },
   "vietnamese": {
     "code": "vi",
-    "system": "http://id.loc.gov/vocabulary/iso639-1",
+    "system": "urn:ietf:bcp:47",
     "display": "Vietnamese"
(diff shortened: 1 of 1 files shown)


jawalonoski:
All the unit tests passed, but the Slack notifications failed. This looks good to go.

jawalonoski:

PR #929, opened 2021-08-19
Add helper scripts

SimonUzL:
I have created a script for the creation of PostgreSQL tables for the import of synthesized CSV files by Synthea. However, it slightly differs from the CSV file data dictionary from the wiki. For instance, a claims.csv or claims_transactions.csv is not generated by Synthea 2.7.0. I think, as a lot of people have to create such a script manually, we should share them. Maybe in this repository or somewhere else?

### DIFF
--- a/scripts/synthea_postgresql.sql
+++ b/scripts/synthea_postgresql.sql
@@ -1,0 +1,219 @@
+CREATE TABLE patients (
+	id uuid primary key,
+	birthdate date,
+	deathdate date,
+	ssn text,
+	drivers text,
+	passport text,
+	prefix text,
+	first text,
+	last text,
+	suffix text,
+	maiden text,
+	marital text,
+	race text,
+	ethnicity text,
+	gender text,
+	birthplace text,
+	address text,
+	city text,
+	state text,
+	contry text,
+	zip text,
+	lat numeric,
+	lon numeric,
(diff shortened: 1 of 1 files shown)


dehall:
@SimonUzL thank you for the pull request.  Personally I've tended to use pgfutter ( https://github.com/lukasmartinelli/pgfutter ) in the past to load the CSV files into postgres, automatically creating the tables as it goes, but I can see the value in specifying the DDL as well.

However we don't generally accept external systems integrations in PRs. See https://github.com/synthetichealth/synthea/wiki/Contributing#contributions-to-avoid

How did you produce this script? This makes me think maybe we should have a more formal way of declaring what's in the CSV, and make it easy for people to produce DDL for whichever DB they might use. The data dictionary wiki is a nice start but only represents what's currently on master, if you're using an older version of Synthea for whatever reason then you can't easily get the CSV data dictionary for the version you're using.

SimonUzL:
@dehall I should have searched for such a tool before I manually write scripts. Thank you for the link and your reply! I thought, that the latest tag in the repository is the most recent stable version of Synthea and thus producing csv files as described in the wiki. A more formal way of declaring for what's in the CSV is a great idea. Next time, before I create a pull request, I go through the Contributing article in the wiki.

PR #938, opened 2021-09-15
Add Coverage resource to top level of each generated Patient

JoshuaASmith:
For our needs with the data we were generating, we added a function to create Coverage objects for each Patient that was generated. To do this, we took examples of other resource generation functions and used those to make the coverage() function. We also looked at how Coverage resources were created when they are nested in ExplanationOfBenefits, and used that as a guide. 

From our testing, this is creating multiple valid Coverage resources for each Patient. Another thought we had was to leverage this new resource generation function inside the EoB function for the coverage resources created there, not sure if that would work. 

I did not see a Contributing guide to see details on when to add tests, let me know if there is any other work you would like done to this PR. I wouldn't consider myself a Java expert but happy to make an attempt. Also if this is not something you all are interested in, please close and we will continue to use/maintain our fork. 

CC: @bzimbelman

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Generator.java
+++ b/src/main/java/org/mitre/synthea/engine/Generator.java
@@ -48,6 +48,8 @@ import org.mitre.synthea.modules.LifecycleModule;
 import org.mitre.synthea.world.agents.Payer;
 import org.mitre.synthea.world.agents.Person;
 import org.mitre.synthea.world.agents.Provider;
+import org.mitre.synthea.world.concepts.CoverageRecord;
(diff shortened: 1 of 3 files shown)


jawalonoski (changes requested):
We're still willing to take this PR if you can update it with the requested changes...

There should be more than one top-level coverage with `Coverage.period` filled out, because the coverage will change over time.

Finally, you should also update the `explanationOfBenefit` of method to actually reference the specific top-level Coverage you are generating, that is applicable for the date range, rather than the contained coverage with the ID `#coverage`.

PR #947, opened 2021-10-19
IHS Providers

eedrummer:
This PR implements changes and additions for providers related to the [Indian Health Service](https://www.ihs.gov).

* There is now a new significant value for `Provider.type` of `IHS Facility`. Upon merging this PR, I'll update the associated information in the wiki.
* IHS Hospitals were pulled out of the main `hospitals.csv` and assigned a type of `IHS Facility`
* Using the spreadsheet found on [Find Health Care](https://www.ihs.gov/findhealthcare/) added primary care locations that are `IHS Facility`. Places used were either health centers or health locations that have addresses that can be successfully geocoded using Google's API for doing so.
* Updated the `ProviderFinderNearest` to exclude `IHS Facilities` if the person in the simulation does not have a race code corresponding to "American Indian or Alaska Native"


src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -318,6 +326,9 @@ public static void loadProviders(Location location, long clinicianSeed) {
         String hospitalFile = Config.get("generate.providers.hospitals.default_file");
         loadProviders(location, hospitalFile, servicesProvided, true, clinicianSeed);
 
+        String ihsHospitalFile = Config.get("generate.providers.ihs.hospitals.default_file");
+        loadProviders(location, ihsHospitalFile, servicesProvided, true, clinicianSeed);
```
dehall, line 330: One comment that was raised in a separate discussion was what happens when we apply another location from https://github.com/synthetichealth/synthea-international ? Turns out it's an exception when we try to read from file path `null` when the config setting isn't set. 
Rather than update every single other country from that repo, could we just add a null check on `ihsHospitalFile` here and on `ihsPCFile` on line 341 below? I'm wondering if this concept is really applicable to any other countries?

This also makes me wonder if this approach of a new file per provider type is the right way to do things, vs having just a "type" column in the file. Maybe that's a discussion topic for later.

eedrummer: @dehall I can add `null` checks where I load IHS related providers. How is this not an issue with the veterans providers? Shouldn't that throw an NPE too? Do all other countries have a similar way of providing healthcare to veterans?

We should have a conversation about how to better address this in the long term. Do we want other countries to be able to have ways of influencing provider finding behavior based on provider types in some sort of code free way?

dehall: Yes I think conceptually the same question exists for Veterans and the va_faciltiies file, but in practice that config setting was in place a couple years before all the synthea-international locations were created, so all the config files there should already have that setting present. The other thing is that Veterans only seek care at VA facilities for certain encounter types, so the va file should be required if there are any Veterans in the population:  https://github.com/synthetichealth/synthea/blob/ihs-providers/src/main/java/org/mitre/synthea/world/agents/behaviors/ProviderFinderNearest.java#L30

@jawalonoski and I talked a little about the basics of cleaning up and updating the provider files yesterday, and we said it would definitely be nice, but the impact to all those international locations means that even that is a potentially significant effort. I mean I like the idea but prioritization is up to Jay. Happy to chat further though.

PR #955, opened 2021-11-10
Prevent duplicate codes in HealthRecord Entries

eedrummer:
This PR addresses cases where `HealthRecord.Entry`s can have multiple instances of the same `Code` in their `codes` property. In cases where a `Person` in the simulation passes through the same `ConditionOnset` state multiple times, Synthea will just keep appending the same condition codes to the `Entry`. This change now checks to see if the `Code` already exists in the `Entry`.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -836,7 +836,7 @@ public abstract class State implements Cloneable, Serializable {
             ClinicianSpecialty.GENERAL_PRACTICE, null);
         entry = encounter;
         if (codes != null) {
-          encounter.codes.addAll(codes);
+          encounter.mergeCodeList(codes);
         }
         person.setCurrentEncounter(module, encounter);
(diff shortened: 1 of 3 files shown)


eedrummer:
Let's hold off on this for now. I would mark it as draft, but I'm not sure you can do that after it is a "real" PR.

I'm running into an issue with the FHIR R4 test after rebasing this on master.

eedrummer:
Fixed the issue that I was seeing. Still not 100% sure what was causing it.

`PersonTest.testPersonFhirR4Recreation` would fail for me running `gradle clean check` on the command line. The test would pass without issue if I ran that single test, or the whole `PersonTest` in IntelliJ IDEA.

After some debugging, what I was finding is that the FHIR R4 record being created would have a double code in a `CarePlan` resource. The last commit in this PR addresses that by applying the same merge code logic to reasons as it did to `Entry.codes`. 

Why the double code would get added only when running the whole test suite is still a mystery to me, but does not seem worth spending the resources to track down.

PR #959, opened 2021-11-18
HIV Module

jawalonoski:
HIV Module.

This module should be ready for review.

The infection rates are low, and the mortality rates are high (especially in the early years), so the number of patients in the general population will not be high. A Keep Module is needed, and possibly adjustments to the infection rate look up tables, if you want to study the population in bulk.

There is plenty of medication diversity as well, that changes over time... e.g., 
image

Not included are any changes related to encounter service types, e.g., specifically visiting HIV clinics. I recommend we incorporate those changes later, once a solution has been built.


src/main/resources/modules/hiv_diagnosis.json
```diff
@@ -0,0 +1,273 @@
...
+          "lookup_table_name": "hiv_diagnosis_later.csv"
+        },
+        {
+          "transition": "Wait_Until_Next_Diagnosis_Check",
+          "default_probability": 0.999903,
+          "lookup_table_name": "hiv_diagnosis_later.csv"
+        }
+      ]
+    },
+    "Diagnosis Check": {
```
eedrummer, line 246: I think we need a Guard state that will wait until 1979. My reading of this is that if anyone is born prior to 1979, when this runs, it will just go to terminal. I ran 10K patients and none of those diagnosed with HIV was born before 1979.

eedrummer (changes requested):
See comment on the diagnosis module.

src/main/resources/modules/hiv_diagnosis.json
jawalonoski: I've made the change, now I'm testing it and running the data analysis.

jawalonoski:
After guard update... meds....
image

jawalonoski:
Causes of Death... red is AIDS/HIV, yellow is COVID-19, all others greyscale...

image

jawalonoski:
This was 100K patients, general population, no other restrictions.

src/main/resources/modules/hiv_diagnosis.json
jawalonoski: Guard added and checked in. Resulting graphs from below.

eedrummer:
I think something odd happened here. The only commit I see after I last commented is the "Merge" one that GitHub claims has no changes. I don't see a `Guard` state in `hiv_diagnosis.json`.

jawalonoski:
Probably user error. I pushed the change.

PR #965, opened 2021-12-10
Allow ISO formatted dates in look up table transitions

eedrummer:
Look up table transitions previously allowed for decisions to be made on the time in the simulation. The way that this works is that time ranges can be defined using milliseconds since the epoch. This works, but is not great for human readability.

This PR preserves that capability, but also allows for date ranges to be defined as ISO Dates. This means what was previously:

`1599264000000-1627948800000`

can now be represented as:

`2020-09-05-2021-08-03`

For those who prefer the old format (there is probably someone), it still works.

### DIFF
--- a/src/main/java/org/mitre/synthea/helpers/Utilities.java
+++ b/src/main/java/org/mitre/synthea/helpers/Utilities.java
@@ -11,13 +11,20 @@ import java.io.IOException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.Calendar;
 import java.util.Random;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.Range;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mitre/synthea/helpers/Utilities.java
```diff
@@ -338,6 +345,46 @@ private static String getSyntheaVersion() {
...
+        Pattern.compile("^(\\d{4}\\-\\d{2}\\-\\d{2})\\-(\\d{4}\\-\\d{2}\\-\\d{2})$");
+
+    Matcher matcher = dateRangeRegex.matcher(range);
+
+    Range<Long> parsedRange;
+    if (matcher.matches()) {
+      parsedRange = Range.between(
+        LocalDate.parse(matcher.group(1), DateTimeFormatter.ISO_LOCAL_DATE).atStartOfDay()
+            .toInstant(ZoneOffset.UTC).toEpochMilli(),
+        LocalDate.parse(matcher.group(2), DateTimeFormatter.ISO_LOCAL_DATE).atStartOfDay()
```
hadleynet, line 377: I wonder if it would be better for the upper bound to be the end of the day so that the range is fully inclusive instead of inclusive of the start and exclusive of the end? Either way, the Javadoc should specify inclusive vs exclusive.

hadleynet (changes requested):
Looks good, just one minor nit.

eedrummer:
@hadleynet I made changes in response to your comments. Also rebased on master.

hadleynet (approved):

PR #971, opened 2022-01-04
Add more usage info to the README

lmsurpre:
And cleanup extraneous paren in usage info.

In my particular case, I was looking for an option to generate more patients with the same set of clinicians/providers, which I now assume is the `-cs` option (which wasn't documented here previously).

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Generating the population one at a time...
 
 Command-line arguments may be provided to specify a state, city, population size, or seed for randomization.
 ```
-run_synthea [-s seed] [-p populationSize] [-m moduleFilter] [state [city]]
+run_synthea [-s seed] [-p populationSize] [state [city]]
 ```
 
 Full usage info can be printed by passing the `-h` option.
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ cd synthea
 ./gradlew build check test
 ```
 
-### Changing the default properties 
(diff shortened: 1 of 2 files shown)


jawalonoski:
Thanks! Could you remove the `-m` lines from your changes?

I do not want to advertise that feature to the world. It causes so many issues and emails. Massive time sink.

See https://github.com/synthetichealth/synthea/wiki/The--M-Feature

lmsurpre:
Sure.  You want it removed from the usage info too `-h` or just the README?
Personally, I think it makes sense to remove it from the example command (https://github.com/synthetichealth/synthea/pull/971/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R57) but leave it in the usage info below.  I'll go with that for now, but happy to make further tweaks as needed.

README.md
```diff
@@ -53,19 +53,46 @@ Generating the population one at a time...
 ```
 
 Command-line arguments may be provided to specify a state, city, population size, or seed for randomization.
-
-Usage is 
 ```
 run_synthea [-s seed] [-p populationSize] [-m moduleFilter] [state [city]]
```
lmsurpre, line 57:
```suggestion
run_synthea [-s seed] [-p populationSize] [state [city]]
```

jawalonoski (approved):

PR #975, opened 2022-01-06
BFD FIPS Codes

jawalonoski:
Adds FIPS county codes to the BFD Beneficiary file.

Also fixes an unrelated unit test where the cost adjustment factor was not properly set (which will cause random unit test failures).


src/main/java/org/mitre/synthea/world/geography/CMSStateCodeMapper.java
```diff
@@ -246,6 +248,38 @@ public String zipToCountyCode(String zipcode) {
...
+    List<LinkedHashMap<String, String>> csvData;
+    try {
+      String csv = Utilities.readResourceAndStripBOM("geography/fipscodes.csv");
+      csvData = SimpleCSV.parse(csv);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    for (LinkedHashMap<String, String> row : csvData) {
+      String zipcode = row.get("zip");
+      if (zipcode.length() > 3) {
```
hadleynet, line 272: This isn't needed, just pull out the nested if

hadleynet (changes requested):
Looks good, just one nit.

src/main/java/org/mitre/synthea/world/geography/CMSStateCodeMapper.java
jawalonoski: I fixed this, and added in another column in the carrier table.

hadleynet (approved):

PR #978, opened 2022-01-07
BFD Add external diagnosis codes.

jawalonoski:
Adds external diagnosis codes to all applicable tables.

**Discussion required about RATE of external codes.**


src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -181,6 +184,26 @@ private BB2RIFExporter() {
...
+        String primary = row.get("primary");
+        String external = row.get("external");
+        double count = Double.parseDouble(row.get("count"));
+        if (!data.containsKey(primary)) {
+          data.put(primary, new RandomCollection<String>());
+        }
+        data.get(primary).add(count, external);
+      }
+    } catch (Exception e) {
+      return null;
```
hadleynet, line 202: Emit a warning if this happens?

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -716,7 +739,16 @@ private void exportOutpatient(Person person, long stopTime)
           fieldValues.put(OUTPATIENT.PRNCPAL_DGNS_CD, mappedDiagnosisCodes.get(0));
         }
       }
-
+      // Check for external code...
+      String primary = fieldValues.get(OUTPATIENT.PRNCPAL_DGNS_CD);
```
hadleynet, line 743: This code block is repeated for each of the claim files with external codes. Can you DRY it out into a single method that takes fieldValues, source field enum value, external field enum value and version field enum value?

jawalonoski: It also needs to take the person (as a source of randomness), and in one case the presentOnAdmission set and presentOnAdmission enum value. So that is seven arguments, which is kind of a lot, but I'll give it a try.

hadleynet (approved):

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -957,12 +1027,22 @@ private void exportInpatient(Person person, long stopTime)
           fieldValues.put(INPATIENT.PRNCPAL_DGNS_CD, mappedDiagnosisCodes.get(0));
         }
       }
+
+      // Check for external code...
+      setExternalCode(person, fieldValues,
```
hadleynet, line 1032: You could move these two calls to setExternalCode inside the if block starting on line 1039 since they both rely on there being a INPATIENT.PRNCPAL_DGNS_CD set previously.

hadleynet (changes requested):
Just one small nit and I think you need to rebase this branch on master.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
jawalonoski: OK, good idea.

hadleynet (approved):

PR #984, opened 2022-01-12
Ignore death dates beyond the end of the simulation in the RIF exporter

hadleynet:
Fixes an issue where a death date in the future would be used in the RIF exporter.


jawalonoski:
Looks good, just running a test now...

jawalonoski (approved):

Issue #983, opened 2022-01-11
Build fails | 71fe525 - Merge pull request #973 from synthetichealth/thread-pool-sizing


IoTFier:
I updated my master repo to the latest commit at master - 71fe525 - Merge pull request #973 from synthetichealth/thread-pool-sizing

The build fails with the following error. 

<img width="1119" alt="image" src="https://user-images.githubusercontent.com/26844611/149017161-4ad2d654-3b84-4731-b12e-7b03643e9675.png">

IoTFier:
Hi @eedrummer, build fails after updating master to the latest commit - 0f63ca4 - Merge pull request #970 from synthetichealth/java17

Please see the error below where 3 tests are failing.

> Task :compileTestJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :test

org.mitre.synthea.engine.FixedRecordTest > fixedDemographicsImportTest FAILED
    java.lang.AssertionError: expected:<4> but was:<5>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at org.mitre.synthea.engine.FixedRecordTest.fixedDemographicsImportTest(FixedRecordTest.java:58)

org.mitre.synthea.export.CCDAExporterTest > testCCDAExport FAILED
    java.lang.AssertionError: Validation of exported CCDA failed: Illegal operation (self.realmCode->isEmpty())|index=54, size=54|index=42, size=42 expected:<0> but was:<3>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.mitre.synthea.export.CCDAExporterTest.testCCDAExport(CCDAExporterTest.java:60)

org.mitre.synthea.world.geography.DemographicsTest > languageFromRaceAndEthnicity FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at org.mitre.synthea.world.geography.DemographicsTest.languageFromRaceAndEthnicity(DemographicsTest.java:78)

[...]

PR #985, opened 2022-01-13
Fix for intermittent test failures of DemographicTest

eedrummer:
Should address #983

Seeds the `Random` used for this test. Switches the town from Philly to Bedford as that is the default.


jawalonoski:
Andy, you shouldn't hard-code it.

For example, this is from the ProviderTest.java

```java
    TestHelper.loadTestProperties();
    String testState = Config.get("test_state.default", "Massachusetts");
    String testTown = Config.get("test_town.default", "Bedford");
    city = new Location(testState, testTown);
```

PR #1008, opened 2022-02-14
Allow end time of simulation to be specified

hadleynet:


jawalonoski (changes requested):
I tested it a couple of different ways using different parameter combinations including '-r`, `-s`, -`cs` and others. Everything works as desired.

Please add the `-e` date to the `./output/metadata/*.json` files.

jawalonoski (approved):

PR #1012, opened 2022-02-22
Demographics fix

jawalonoski:
- Fixes a rare exception related to demographic distributions that are all zeroes (e.g., if every income bracket for a location has a zero probability -- then a person cannot exist within any of the income brackets, and a runtime exception is thrown).
- Rather than patch the demographics file, part of this fix is to detect all zero distributions and use a default uniform distribution (which was already being applied in the case where the distributions were null instead of zero). Don't like that? Override the demographics file.
- I also made partial fixes to the demographics file (fixed one county in New Mexico that was especially egregious in this regard).
- One other fix related to Unicode accented n's

### DIFF
--- a/src/main/java/org/mitre/synthea/world/geography/Demographics.java
+++ b/src/main/java/org/mitre/synthea/world/geography/Demographics.java
@@ -521,12 +521,7 @@ public class Demographics implements Comparable<Demographics>, Serializable {
    */
   private static void nonZeroDefaults(Map<String, Double> map) {
     // Any null or nan values should be zero
-    for (String key : map.keySet()) {
-      Double value = map.get(key);
-      if (value == null || value.isNaN()) {
-        map.put(key, Double.valueOf(0));
-      }
-    }
+    map.replaceAll((key, value) -> (value == null || value.isNaN()) ? 0.0 : value);
     // Now check if all values are zero
     boolean allZero = true;
     for (Double value : map.values()) {
@@ -535,11 +530,10 @@ public class Demographics implements Comparable<Demographics>, Serializable {
         break;
       }
     }
+    // If all values were zero, apply a uniform distribution.
     if (allZero) {
-      Double value = 1.0 / Double.valueOf(map.size());
-      for (String key : map.keySet()) {
-        map.put(key, value);
-      }
+      Double value = 1.0 / map.size();
+      map.replaceAll((key, oldValue) -> value);
     }
   }
 
--- a/src/main/java/org/mitre/synthea/world/geography/Demographics.java
+++ b/src/main/java/org/mitre/synthea/world/geography/Demographics.java
@@ -252,7 +252,6 @@ public class Demographics implements Comparable<Demographics>, Serializable {
      */
 
     String pickedRange = incomeDistribution.next(random);
-
     String[] range = pickedRange.split("\\.\\.");
     // TODO this seems like it would benefit from better caching
     int low = Integer.parseInt(range[0]) * 1000;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mitre/synthea/world/geography/Demographics.java
```diff
@@ -502,12 +506,43 @@ private static Demographics csvLineToDemographics(Map<String,String> line) {
...
+   * @param map The map to check.
+   */
+  private static void nonZeroDefaults(Map<String, Double> map) {
+    // Any null or nan values should be zero
+    for (String key : map.keySet()) {
+      Double value = map.get(key);
+      if (value == null || value.isNaN()) {
+        map.put(key, Double.valueOf(0));
+      }
+    }
```
hadleynet, line 529: You could replace this for loop with: map.replaceAll((key, value) -> value == null || value.isNaN() ? 0.0 : value)

src/main/java/org/mitre/synthea/world/geography/Demographics.java
```diff
@@ -502,12 +506,43 @@ private static Demographics csvLineToDemographics(Map<String,String> line) {
...
+      if (value != 0)  {
+        allZero = false;
+        break;
+      }
+    }
+    if (allZero) {
+      Double value = 1.0 / Double.valueOf(map.size());
+      for (String key : map.keySet()) {
+        map.put(key, value);
+      }
```
hadleynet, line 542: You could replace this for loop with map.replaceAll((key, oldValue) -> value)

src/main/java/org/mitre/synthea/world/geography/Demographics.java
```diff
@@ -502,12 +506,43 @@ private static Demographics csvLineToDemographics(Map<String,String> line) {
...
+    // Now check if all values are zero
+    boolean allZero = true;
+    for (Double value : map.values()) {
+      if (value != 0)  {
+        allZero = false;
+        break;
+      }
+    }
+    if (allZero) {
+      Double value = 1.0 / Double.valueOf(map.size());
```
hadleynet, line 539: I don't think you need the Double.valueOf here

src/main/java/org/mitre/synthea/world/geography/Demographics.java
```diff
@@ -502,12 +506,43 @@ private static Demographics csvLineToDemographics(Map<String,String> line) {
...
+   * If the values are all zero, an equally weighted uniform distribution
+   * will be set.
+   * @param map The map to check.
+   */
+  private static void nonZeroDefaults(Map<String, Double> map) {
+    // Any null or nan values should be zero
+    for (String key : map.keySet()) {
+      Double value = map.get(key);
+      if (value == null || value.isNaN()) {
+        map.put(key, Double.valueOf(0));
```
hadleynet, line 527: Why not map.put(key, 0.0)?

hadleynet (changes requested):

src/main/java/org/mitre/synthea/world/geography/Demographics.java
jawalonoski: Done.

src/main/java/org/mitre/synthea/world/geography/Demographics.java
jawalonoski: Done.

src/main/java/org/mitre/synthea/world/geography/Demographics.java
jawalonoski: Done.

src/main/java/org/mitre/synthea/world/geography/Demographics.java
jawalonoski: Done.

hadleynet (approved):

PR #1027, opened 2022-03-21
Telemedicine Encounters

eedrummer:
This PR adds a feature to the Encounter state that allows for visits to probabilistically or always happen via telemedicine. Further, it adds a new `virtual` encounter class.

With this PR, Encounter states can have a `telemedicine_possibility` property, which can be set to `never`, `possible` or `always`.

When an ambulatory Encounter has a `telemedicine_possibility` set to `possible`, there is a 0.4% chance prior from January 1, 2016 to April 1 2020 and 15% chance after that where the visit will change to telemedicine. The code for the visit will not change, but the encounter class will be changed from `ambulatory` to `virtual`. For records being exported to FHIR R4, the Location of the Encounter will change to a "patient's home" Location "kind" resource.

Previously created GMF modules were reviewed for telemedicine, with encounters being updated to allow for virtual visits where it makes clinical sense.

Some GMF modules already had telemedicine encounters built in. These were updated to use the virtual encounter class. The Veteran PTSD module was modified to put in a Guard state at 1999 to prevent telemedicine encounters from being performed prior to that. This probably should have been in the module originally. The date of 1999 was selected since that is when the PHQ-9 was created, which is used by the module.

With the changes, running a Synthea simulation for 10K patients results in the following counts of telemedicine encounters over time:

Screen Shot 2022-03-15 at 5 26 20 PM

See synthetichealth/module-builder#310 for the corresponding changes in the Module Builder.


src/main/java/org/mitre/synthea/export/FhirR4PatientHome.java
```diff
@@ -0,0 +1,40 @@
...
+      patientHome.setPhysicalType(new CodeableConcept()
+          .addCoding(new Coding()
+              .setCode(LocationPhysicalType.HO.toCode())
+              .setSystem(LocationPhysicalType.HO.getSystem())
+              .setDisplay(LocationPhysicalType.HO.getDisplay())
+          ));
+      patientHome.setDescription("Patient's Home");
+      // Not really generating a random UUID. Given that this is not tied to a particular provider
+      // or person, this just makes up a person with a hardcoded random seed.
+      patientHome.setId(new Person(1).randUUID().toString());
```
jawalonoski, line 36: Did you do this so that the "Patient Home" location would ALWAYS have the same UUID?

jawalonoski (commented):
So far so good.

dehall:
I happened to be flipping through old notes and was just reminded of this headline from ~5 years ago: [Nearly half of US medical care comes from emergency rooms](https://www.sciencedaily.com/releases/2017/10/171017091849.htm)

Not to hold up this PR for a what-if, but do we feel the approach used here can scale if we want to explore having certain encounters sometimes be `emergency` instead of `ambulatory`? (Or maybe vice versa such in the case of the Appendicitis module.)

eedrummer:
This was marked as draft until the BB2RIF exporter was complete. I think that is now done.

eedrummer:
@dehall I think I understand what you are saying. Instead of having something that would just allow ambulatory encounters be virtual, we could build something that would assign probabilities to a particular encounter class. As an example, we could say that something has a 20% chance of being ambulatory and and 80% chance of being emergency.

This would definitely be a more general solution. The downside is that it wouldn't allow for the general telemedicine probabilities that this approach offers. The probabilities would need to be baked in to each encounter and even then, they wouldn't change based on time in the simulation unless we made things even more complex.

jawalonoski:
I agree that it would be interesting to have a general solution where encounter types can be of any other type, but if we want to go down that road, I think that will be a separate PR. I don't want to delay this for that.

PR #1028, opened 2022-03-24
Metabolic syndrome module refactor

dehall:
This PR refactors and simplifies the metabolic syndrome modules, currently `metabolic_syndrome_disease` and `metabolic_syndrome_care`, by breaking out major pieces into submodules and other top-level modules:

 - `metabolic_syndrome_disease`
   - extracted CKD into a separate module
   - refactored onset - previously diabetes onset would happen with a uniform distribution between 18, now the onset range has been adjust to more accurately reflect onset rates
   - extracted HTN onset into the hypertension module, previously it would only happen at age 18 or at onset of diabetes, now it has a race-and-sex specific rate per year
 - `metabolic_syndrome_care`
   - extracted submodules:
   - medications
   - care for eye conditions
   - care for kidney conditions
   - amputations 
 - `chronic_kidney_disease`
 - `hypertension`
 - `wellness_encounters`
   - obesity check moved here

Note: this was originally done on the `cabg_cardio_gmf_htn_trial` branch, but sending this separately in an attempt to clean up and harmonize our branches.


dehall:
Marked this as draft since I forgot there are corresponding hypertension module changes that need to be included

src/main/resources/modules/metabolic_syndrome/medications.json
```diff
@@ -0,0 +1,389 @@
...
+    "Prescribe_Metformin": {
+      "type": "MedicationOrder",
+      "codes": [
+        {
+          "system": "RxNorm",
+          "code": "860975",
+          "display": "24 HR Metformin hydrochloride 500 MG Extended Release Oral Tablet"
+        }
+      ],
+      "reason": "Diagnose_Diabetes",
```
jawalonoski, line 52: None of the `"reason": "Diagnose_Diabetes"` will work because that state does not exist in this module. Better to switch to the attribute `diabetes_stage`, I believe.

dehall: Updated all the instances of this, and confirmed the medications now have a reason in the exports

jawalonoski (approved):

PR #1042, opened 2022-04-08
Blood pressure model with medication impacts

dehall:
One more PR bringing features over from `cabg_cardio_gmf_htn_trial`. This PR reworks the Blood Pressure model to account for hypertension medications having impact. Ultimately I think this kind of thing could be reworked to medications having impact on arbitrary vital signs, but this is a first step.

Medications are defined in file `src/main/resources/htn_drugs.csv` with, as of this writing, more columns and codes than are required, so I may pare that down from what we had previously. Most important are the "code" and "impact minimum" and "impact maximum" columns, which are the RxNorm code and the expected range in mmHg of impact on blood pressure the medication is expected to have. As of now all drugs have impact range -11.5 to -16.29.

The basic approach of the model is:
 - determine baseline blood pressure
 - take into account medications
 - take into account lifestyle changes (CarePlans)

This model caches a number of things to ensure consistency, so that an individual's BP doesn't fluctuate wildly:
 - the final value at a given time, so that subsequent calls at the same timestamp don't have to recalculate
 - the individual's baseline blood pressure depending on hypertension status
 - the impact each drug has on blood pressure for that individual

Note one downside of this change is that in order to get realistic results at the aggregate level, we had to remove the "trending" aspect of the BloodPressureValueGenerator. We can try to reintegrate that but it will take more time and analysis.

Also note that this doesn't take into account the "titration" concept that was introduced in the htn_trial branch, as that concept was not fully fleshed out and very hackish. It worked for a research effort but it's not ready for master.

### DIFF
--- a/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
+++ b/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
@@ -1,16 +1,22 @@
 package org.mitre.synthea.modules;
 
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.mitre.synthea.engine.Components.Range;
+import org.mitre.synthea.helpers.SimpleCSV;
 import org.mitre.synthea.helpers.TrendingValueGenerator;
+import org.mitre.synthea.helpers.Utilities;
 import org.mitre.synthea.helpers.ValueGenerator;
 import org.mitre.synthea.world.agents.Person;
 import org.mitre.synthea.world.concepts.BiometricsConfig;
+import org.mitre.synthea.world.concepts.HealthRecord;
 
 /**
- * Generate realistic blood pressure vital signs.
- * Can reproducibly look a few days into the past and future.
- * <p></p>
- * See <a href="https://raywinstead.com/bp/thrice.htm">https://raywinstead.com/bp/thrice.htm</a>
- * for desired result
+ * Generate realistic blood pressure vital signs,
+ * with impacts from select medications.
  */
 public class BloodPressureValueGenerator extends ValueGenerator {
   public enum SysDias {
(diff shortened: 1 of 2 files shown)


jawalonoski:
It looks like the Blood Pressure module doesn't have any unit tests. My guess is that whatever unit test coverage it does get is from the unit tests that generate patients.

It is probably worth adding a few tests to make sure the impacts are being done correctly.

Should we have a unit test that checks to see that the codes in this CSV file exist in the modules? Or the codes in a hypertension module exist in the CSV file?

src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
```diff
@@ -47,156 +87,167 @@ public BloodPressureValueGenerator(Person person, SysDias sysDias) {
...
+            // max value
+            return maxDrop;
+          } else if (time < start) {
+            return 0.0;
+          }
+
+          // dy / dx = -14/1 = -14
+          // y = (dy/dx) * x
+          // x = (time - start) / 1 yr
+
```
jawalonoski, line 225: Is this commented out code necessary?

src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
```diff
@@ -47,156 +87,167 @@ public BloodPressureValueGenerator(Person person, SysDias sysDias) {
...
-      currentTime = currentTime + duration + 1L;
-      generatePeriod -= duration;
-      startValue = endValue;
-    }
+    return impact;
   }
 
-  private double calculateMean(Person person, long time) {
-    // TODO: Take additional factors into consideration: age + gender
+  private double calculateMean(Person person, Double startValue, int days, long time) {
```
jawalonoski, line 128: I'm confused by `calculateMean(...)` ... it does not calculate a mean, it just seems to set a static baseline (if one has not been set already), and return that baseline.

There is no mean or even random noise added.

Also, the arguments `startValue`, `days`, and `time` are not used at all.

src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
```diff
@@ -47,156 +87,167 @@ public BloodPressureValueGenerator(Person person, SysDias sysDias) {
...
-  private TrendingValueGenerator getTrendingValueGenerator(long time, boolean createNewGenerators) {
-    // System.out.println("getTVG @ " + time);
-
-    for (int i = 0; i < RING_ENTRIES; i++) {
-      final TrendingValueGenerator trendingValueGenerator = ringBuffer[i];
-      if (trendingValueGenerator != null && trendingValueGenerator.getBeginTime() <= time
-          && trendingValueGenerator.getEndTime() >= time) {
-        return trendingValueGenerator;
-      }
+  private static double getDrugImpact(Person person, long time, String drug, SysDias sysDias,
```
jawalonoski, line 107: The argument `time` is not used.

src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
```diff
@@ -47,156 +87,167 @@ public BloodPressureValueGenerator(Person person, SysDias sysDias) {
...
         }
-      } else {
-        return person.rand(NORMAL_DIA_BP_RANGE);
       }
     }
+
+    return 0.0;
+  }
+
+  private double getMedicationImpacts(Person person, long time) {
```
jawalonoski, line 236: The argument of `time` will not be used once it is removed from `getDrugImpact` where it is also not used.

src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
dehall: It's not really commented out code, it's the math of how to calculate the impact value for a certain date, assuming it goes from 0 to -14 in 1 year. But yeah it's kind of a mess, I'll clean this up.

jawalonoski (approved):

PR #1047, opened 2022-04-12
Specialities on encounters

eedrummer:
Adds a `specialty` property on the `Encounter` state in GMF. This will require the encounter to happen at a Provider that offers that specialty.

The preferred provider infrastructure has been updated to handle this. People now have a preferred provider by encounter class and specialty.

Checkstyle fixes added in as well. It appears that there are now checkstyle violations on master 😲

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -819,6 +819,7 @@ public abstract class State implements Cloneable, Serializable {
     private List<Code> codes;
     private String reason;
     private String telemedicinePossibility;
+    private String specialty;
 
     @Override
     public Encounter clone() {
@@ -871,8 +872,12 @@ public abstract class State implements Cloneable, Serializable {
         } else {
           type = EncounterType.fromString(encounterClass);
         }
+        String desiredSpeciality = ClinicianSpecialty.GENERAL_PRACTICE;
+        if (this.specialty != null && !this.specialty.isEmpty()) {
+          desiredSpeciality = this.specialty;
+        }
         HealthRecord.Encounter encounter = EncounterModule.createEncounter(person, time, type,
-            ClinicianSpecialty.GENERAL_PRACTICE, null);
+            desiredSpeciality, null);
         entry = encounter;
         if (codes != null) {
           encounter.mergeCodeList(codes);
@@ -963,7 +968,7 @@ public abstract class State implements Cloneable, Serializable {
         Provider medicationProvider = person.getCurrentProvider(module.name);
         if (medicationProvider == null) {
           // no provider associated with encounter or medication order
-          medicationProvider = person.getProvider(EncounterType.WELLNESS, time);
+          medicationProvider = person.getProvider(EncounterType.WELLNESS, null, time);
         }
         int year = Utilities.getYear(time);
         medicationProvider.incrementPrescriptions(year);
@@ -1332,7 +1337,7 @@ public abstract class State implements Cloneable, Serializable {
       Provider medicationProvider = person.getCurrentProvider(module.name);
       if (medicationProvider == null) {
         // no provider associated with encounter or medication order
-        medicationProvider = person.getProvider(EncounterType.WELLNESS, time);
+        medicationProvider = person.getProvider(EncounterType.WELLNESS, null, time);
       }
 
       int year = Utilities.getYear(time);
@@ -1556,7 +1561,7 @@ public abstract class State implements Cloneable, Serializable {
       if (person.getCurrentProvider(module.name) != null) {
         provider = person.getCurrentProvider(module.name);
       } else { // no provider associated with encounter or procedure
-        provider = person.getProvider(EncounterType.WELLNESS, time);
+        provider = person.getProvider(EncounterType.WELLNESS, null, time);
       }
       int year = Utilities.getYear(time);
       provider.incrementProcedures(year);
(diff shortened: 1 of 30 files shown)


src/main/java/org/mitre/synthea/engine/State.java
```diff
@@ -871,8 +872,12 @@ public boolean process(Person person, long time) {
         } else {
           type = EncounterType.fromString(encounterClass);
         }
+        String desiredSpeciality = ClinicianSpecialty.GENERAL_PRACTICE;
+        if (this.specialty != null && !this.specialty.isEmpty()) {
+          desiredSpeciality = this.specialty;
```
jawalonoski, line 877: Do we want to make sure that `this.speciality` exists within the array of `ClinicianSpeciality.getSpecialities()`?

src/main/java/org/mitre/synthea/export/CCDAExporter.java
```diff
@@ -117,6 +119,8 @@ public static String export(Person person, long time) {
       }
     }
 
+    person.attributes.put("wellness_provider", person.preferredProviders.getWellnessProvider());
```
jawalonoski, line 122: "wellness_provider" Magic String, would prefer a static final string from somewhere.

src/main/java/org/mitre/synthea/world/agents/Person.java
```diff
@@ -611,42 +615,24 @@ public void setCurrentEncounter(Module module, Encounter encounter) {
...
    * provider at the specified time as the preferred provider for this encounter type.
    */
-  public Provider getProvider(EncounterType type, long time) {
-    String key = PREFERREDYPROVIDER + type;
-    if (!attributes.containsKey(key)) {
-      setProvider(type, time);
+  public Provider getProvider(EncounterType type, String speciality, long time) {
+    if (!preferredProviders.doesRelationshipExist(type, speciality)) {
+      setProvider(type, speciality, time);
     }
```
jawalonoski, line 626: Thoughts on having both `getProvider(EncounterType, long)` and `getProvider(EncounterType, String, long)`?

Where the first is a one-liner that passes a `null` to the String argument of the other function, rather than have `null` arguments all throughout the code?

Seems weird to me as-is.

src/main/java/org/mitre/synthea/world/concepts/PreferredProviders.java
```diff
@@ -0,0 +1,112 @@
...
+   * @return the new preferred Provider
+   */
+  public Provider startNewRelationship(Person person, HealthRecord.EncounterType type,
+                                       String speciality, long time) {
+    Provider provider = Provider.findService(person, type, speciality, time);
+    if (provider == null && Provider.USE_HOSPITAL_AS_DEFAULT) {
+      // Default to Hospital
+      provider = Provider.findService(person, HealthRecord.EncounterType.INPATIENT, speciality,
+          time);
+    }
```
jawalonoski, line 63: Just to be clear, what is the expected behavior here if `USE_HOSPITAL_AS_DEFAULT` is not in play?

Downstream NullPointerException?

jawalonoski (commented):
Left some comments and questions.

src/main/java/org/mitre/synthea/export/CCDAExporter.java
jawalonoski: Not critical though.

src/main/java/org/mitre/synthea/engine/State.java
eedrummer: Now throws an `IllegalArgumentException` if there is a specialty that it doesn't recognize.

src/main/java/org/mitre/synthea/export/CCDAExporter.java
eedrummer: It is a magic string, however it's not used in any other java code.

In context, a bunch of properties get set in the method and there are a ton of magic strings preceding it. I can extract them all out to constants, but IMHO, it doesn't add much value.

src/main/java/org/mitre/synthea/world/agents/Person.java
eedrummer: My $0.02 is that I like the `null`s. The reality is that if the PR is accepted, Synthea is in a little bit of an odd state. In the past, there really wasn't specialty, so everything sort of being general practice was OK. Now, there are likely places where we don't want the default behavior. Having the `null` is a place holder that a future effort can come along and fill in.

src/main/java/org/mitre/synthea/world/concepts/PreferredProviders.java
eedrummer: Good question. This was a copy and paste from `setProvider` that used to be on `Person`. I have now made it more explicit and it throws an `IllegalStateException`. Otherwise, it would NPE when doing the put in the relationships table.

jawalonoski:
Andy, can you resolve the conflicts? Then I'll merge it in if it passes the unit tests.

eedrummer:
Converting this to draft while we figure out the best way to address the underlying needs of this PR. Namely, how can we limit a provider location in Synthea to offer a subset of services. It may be something like fully enabling some of the clinical specialty infrastructure... or something else.

eedrummer:
There have been changes in Synthea since this PR was first submitted. When this was first created, the provider files had specialities listed. An update to the provider files removed specialties.

The more recent commits attempt to add specialties back in using information from NPPES. It's unclear at the current moment whether this approach will be useable as the test suite is breaking because some patients in MA can't find a cardiologist. This may be due to me picking the wrong specialty for cardiology or it may be that the data isn't good enough for our purposes and a new approach is needed.

eedrummer:
After attempting to do some more work, it's clear that the Healthcare Provider Taxonomy Codes in NPPES aren't going to be useful for clinical specialty information for providers in Synthea. There are a few reasons for this:

1. Most hospitals are listed with just the General Acute Care Hospital Code.
2. Although NPPES has information on provides that work at these hospitals, attempting to aggregate or "roll up" this information is difficult. Inconsistencies in organization names (Massachusetts General Hospital vs. MG) and addresses (55 Fruit St. vs. 55 Fruit Street) make require a more sophisticated approach to aggregation than plain string matching.

If specialty information is to be incorporated into Synthea, it will likely have to come from a different source.

Issue #1046, opened 2022-04-12
java.util.List.size() because zipsForCity


AmmarFaiq:
### What happened?

Guys, I found this during the run of the simulation. Even though it can produce most people, sometimes this error log shows. Any ideas why such errors happened?

Thanks,

### Environment

```markdown
- OS: Windows 10
- Java: JDK 18
```

### Relevant log output

```shell
java.util.concurrent.ThreadPoolExecutor@70a898b0[Shutting down, pool size = 16, active threads = 16, queued tasks = 2883, completed tasks = 7101]

and 

java.lang.NullPointerException: Cannot invoke "java.util.List.size()" because "zipsForCity" is null
        at org.mitre.synthea.world.geography.Location.assignPoint(Location.java:375)
        at org.mitre.synthea.modules.LifecycleModule.birth(LifecycleModule.java:168)
        at org.mitre.synthea.engine.Generator.createPerson(Generator.java:661)
        at org.mitre.synthea.engine.Generator.generatePerson(Generator.java:477)
        at org.mitre.synthea.engine.Generator.lambda$run$3(Generator.java:367)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

jawalonoski:
What state and/or city are you running?

jawalonoski:
This is occurring because there are no zip codes for the area being simulated, as per the https://github.com/synthetichealth/synthea/blob/master/src/main/resources/geography/zipcodes.csv file.

It would help if you could tell us what location you were running, so we could try to track down the missing zip code.

In the meantime, I can add a null check on `zipsForCity`.

jawalonoski:
I'm going to close this issue because we made a patch that should stop the program from crashing, but on the other hand, you didn't tell us what city or state you were running so we could properly debug the issue.... so, reopen this if you see it again.

PR #1048, opened 2022-04-13
Add null check to assignPoint methods.

jawalonoski:
Meant to address #1046 

Not sure about this "fix" though because it may just mask the fact that a place exists without a corresponding entry in the zip code file.

Which shouldn't happen.


eedrummer:
Not sure how this is happening. I did a quick check of the `demographics.csv` and `zipcodes.csv` on the current master branch. I could not find any rows in demographics that don't have one or more rows in zip codes and vice-versa.

PR #1054, opened 2022-04-26
Space Health Module

awatson1978:
Contains a simple module that calculates the probability of a vehicle launch accident.  Trivia fact:  there have been ~600 astronauts, and 30 deaths; resulting in an overall 5% mortality rate for space travelers.


jawalonoski:
I'm going to convert this to a draft PR, because it is still a work in progress.

PR #1055, opened 2022-04-27
RIF Export Validation

hadleynet:
This PR adds unit tests to validate the content of the carrier RIF file and fixes issues highlighted by those tests.


jawalonoski (approved):
For some reason I don't think I saw this when it changed...

```java
          // if (hcpcsCode == null) {
          //   continue; // skip this line item
          // }
```

So, we're exporting claim line items without a HCPCS code... technically allowed... but, should we do that?

PR #1103, opened 2022-06-24
Update US Core v4 export and enforce physiology usage more consistently.

jawalonoski:
This PR updates the FHIR R4 exporter with US Core v4 profile names.

Note that US Core v4 changes are already implemented in `master` about 95% of the way, this PR cleans up some corner cases.

It also fixes a transient exporter test bug associated with the physiology state and exporting FHIR SampledData. These changes enforce physiology to be used for unit tests in order to check this corner case.


config/simulations/ecg.yml
```diff
@@ -37,7 +37,7 @@ charts:
       axisParamX: time
       axisLabelX: time (s)
       axisLabelY: Current (A)
-      lowerBoundY: -0.05
-      upperBoundY: 0.15
+      #lowerBoundY: -0.05
+      #upperBoundY: 0.15
```
jawalonoski, line 41: These are commented out because they are not valid parameters on the class.

src/main/java/org/mitre/synthea/engine/Components.java
```diff
@@ -101,7 +101,7 @@ public static class SampledData implements Serializable {
     public Double lowerLimit; // Lower limit of detection
     public Double upperLimit; // Upper limit of detection
     public List<String> attributes; // Person attributes containing TimeSeriesData objects
-    public transient List<TimeSeriesData> series; // List of actual series data collections
+    public List<TimeSeriesData> series; // List of actual series data collections
```
jawalonoski, line 104: Not sure why this was ever `transient` but it ended up making the export tests fail (because the TestHelper serializes all the patients).

jawalonoski:
This PR is ready for review.

src/main/java/org/mitre/synthea/engine/Components.java
eedrummer: I'm not 100% sure that I fully understand the code, but I think this was marked as `transient` because it is copying information and could be recreated if deserialized. `setSeriesData` looks like it copies information from `Person.attributes` and puts it in `series`, which could be rerun on a rehydrated instance if necessary.

jawalonoski: Yes, you are correct in that it could be recreated if deserialized. But I see two issues with that:

1) the `Person.attributes` will likely have changed from the moment the `SampledData.series` was first initialized... meaning the value of the EKG variable in the year 2010 of the simulation won't be the same as the value of the EKG variable in the year 2015 of the simulation. We reuse patient attributes.

2) In the case of the export test failures, these `SampledData` objects are actually stored in `HealthRecord.Observation.value` so that during deserialization the `setSeriesData` method will not run...  which is why some of the export validations were breaking. In other words, Observations without values fail to validate.

PR #1108, opened 2022-07-13
Fix template to account for properties with values of false

hadleynet:
The original code just looked for the presence of the alcoholic, homeless and smoker properties but ignored their value, this PR checks that the value is both present and true.


jawalonoski (changes requested):
one minor change (that was already there):

```
Patient quite smoking
```
Obviously, that should be `quit`.

jawalonoski (approved):

PR #1118, opened 2022-08-09
Various updates to BFD RIF SNF and HHA exports

hadleynet:
Note that there are three new mapping files (available on the file share) utilized by this PR.

1. Fix a bug where all procedures would happen in one day
2. Reduce probability of various procedures during SNF stay
3. RIF revenue center codes are now mapped from procedure codes
4. Update BFD RIF SNF exporter so that claim lines without HCPCS codes are not skipped (matches real data)
5. Add mappings to SNF PPS and PDPM codes to better match real data
6. Centralize logic that determines which events map to which BFD RIF claim files

### DIFF
--- a/src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
+++ b/src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
@@ -58,6 +58,8 @@ import org.mitre.synthea.export.BB2RIFStructure.OUTPATIENT;
 import org.mitre.synthea.export.BB2RIFStructure.PDE;
 import org.mitre.synthea.export.BB2RIFStructure.SNF;
 import org.mitre.synthea.helpers.Config;
+import org.mitre.synthea.helpers.ConsolidatedServicePeriods;
+import org.mitre.synthea.helpers.ConsolidatedServicePeriods.ConsolidatedServicePeriod;
 import org.mitre.synthea.helpers.RandomCollection;
 import org.mitre.synthea.helpers.RandomNumberGenerator;
 import org.mitre.synthea.helpers.SimpleCSV;
@@ -2578,6 +2580,10 @@ public class BB2RIFExporter {
     HashMap<HHA, String> fieldValues = new HashMap<>();
     long claimCount = 0;
     int homeVisits = 0;
+
+    long maxGapForContinuousHHAService = Utilities.convertTime("days", 2);
+    ConsolidatedServicePeriods servicePeriods = new ConsolidatedServicePeriods(
+            maxGapForContinuousHHAService);
     for (HealthRecord.Encounter encounter : person.record.encounters) {
       if (encounter.stop < startTime || encounter.stop < claimCutoff) {
         continue;
@@ -2585,8 +2591,10 @@ public class BB2RIFExporter {
       if (!getClaimTypes(encounter).contains(ClaimType.HHA)) {
         continue;
       }
+      servicePeriods.addEncounter(encounter);
+    }
 
-      homeVisits += 1;
+    for (ConsolidatedServicePeriod servicePeriod: servicePeriods.getPeriods()) {
       long claimId = BB2RIFExporter.claimId.getAndDecrement();
       long claimGroupId = BB2RIFExporter.claimGroupId.getAndDecrement();
       long fiDocId = BB2RIFExporter.fiDocCntlNum.getAndDecrement();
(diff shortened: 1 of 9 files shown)


src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -2569,46 +2607,88 @@ private long exportHome(Person person, long startTime, long stopTime) throws IOE
...
+                hcpcsCode = hcpcsCodeMapper.map(code.code, person, true);
+                if (hhaRevCntrMapper.canMap(code.code)) {
+                  revCenter = hhaRevCntrMapper.map(code.code, person);
+                }
+                break; // take the first mappable code for each procedure
+              }
+            }
+            if (hcpcsCode == null) {
+              revCenter = HHA_TOTAL_CHARGE_REV_CNTR;
+            }
```
jawalonoski, line 2660: I might be reading this wrong, but it looks like multiple line items per claim that do not have a HCPCS mapping could have the total charge rev center, but don't you only want that once per claim?

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -2617,19 +2697,12 @@ private long exportHome(Person person, long startTime, long stopTime) throws IOE
...
         }
       }
-
       // PTNT_DSCHRG_STUS_CD: 1=home, 2=transfer, 3=SNF, 20=died, 30=still here
-      String dischargeStatus = null;
-      if (encounter.ended) {
-        dischargeStatus = "1"; // TODO 2=transfer if the next encounter is also inpatient
-      } else {
-        dischargeStatus = "30"; // the patient is still here
-      }
```
jawalonoski, line 2627: Looks like you got rid of this? Is that because you never saw it in the data? It should be possible.

jawalonoski (commented):
Still reviewing...

jawalonoski:
I ran a thousand patients...

image

image

jawalonoski:
For SNF...

image


image

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hadleynet: Right, all of the events with a null HCPCS code get rolled up into the 0001 rev center. Note that REV_CNTR_UNIT_CNT indicates how many.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hadleynet: I never saw it in the exports I ran. If we break up service intervals (one long SNF encounter, multiple consecutive HHA encounters) into multiple claims then that code would be used for non-final claims.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -432,6 +447,66 @@ public void export(Person person, long stopTime, int yearsOfHistory) throws IOEx
...
+
+  private boolean isVAorIHS(HealthRecord.Encounter encounter) {
+    boolean isVA = (ProviderType.VETERAN == encounter.provider.type);
+    // IHS facilities have valid 6 digit id, IHS centers don't
+    boolean isIHSCenter = (ProviderType.IHS == encounter.provider.type)
+            && encounter.provider.id.length() != 6;
+    return isVA || isIHSCenter;
+  }
+
+  private Set<ClaimType> getClaimTypes(HealthRecord.Encounter encounter) {
```
jawalonoski, line 469: This is nice.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -432,6 +447,66 @@ public void export(Person person, long stopTime, int yearsOfHistory) throws IOEx
...
+      } else if (isInpatient || isEmergency) {
+        types.add(ClaimType.INPATIENT);
+      } else if (isPrimary || isWellness || isUrgent) {
+        types.add(ClaimType.CARRIER);
+      } else if (isAmbulatory || isOutpatient || isVirtualOutpatient) {
+        types.add(ClaimType.OUTPATIENT);
+      }
+
+      if (types.isEmpty()) {
+        System.out.printf("BFD RIF unhandled encounter type");
```
jawalonoski, line 503:
```suggestion
        System.out.printf("BFD RIF unhandled encounter type (" + encounter.type
            + ") and provider type (" + encounter.provider.type.toString() + ")");
```

Just in case this happens... say if we add new encounter types or provider types...  it would be nice to print out the odd combination.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
```diff
@@ -2569,46 +2607,88 @@ private long exportHome(Person person, long startTime, long stopTime) throws IOE
...
+              consolidatedClaimLines.addClaimLine(hcpcsCode, revCenter, lineItem);
+            }
+          }
+        }
+      }
+
+      // Use the final encounter in the service period to set all of the remaining field values that
+      // are the same for all claim lines
+      HealthRecord.Encounter encounter = servicePeriod.getEncounters().get(
+              servicePeriod.getEncounters().size() - 1);
```
jawalonoski, line 2676: I'm not totally in love with this, because it is making some assumptions that the providers, clinicians, primary diagnosis reason, etc -- are all the same for all of the Encounters in the servicePeriod.

It isn't a terrible assumption, and it should mostly work. We can leave it for now.

jawalonoski (approved):
Approved.

src/main/java/org/mitre/synthea/export/BB2RIFExporter.java
hadleynet: Good point. We could address that in the ConsolidatedServicePeriods class by using provider, clinician etc to stratify service periods. I'll add a TODO.

PR #1123, opened 2022-08-10
Populate your FHIR endpoint with generated Synthea data

lkhudisman:
Populate your FHIR endpoint with generated Synthea data.
URL parameter is your FHIR server endpoint ending with /fhir/
You will run this script after the run_synthea script is finished generating test patients for you.
To execute: python ./populate_data.py

### DIFF
--- a/populate_data.py
+++ b/populate_data.py
@@ -1,0 +1,47 @@
+import os
+import requests
+
+"""
+Populate your FHIR endpoint with generated Synthea data.
+URL parameter is your FHIR server endpoint ending with /fhir/
+You will run this script after the run_sythea script is finished generateing test patients for you.
+To execute: python ./populate_data.py
+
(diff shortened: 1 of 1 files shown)


eedrummer:
@lkhudisman Thank you for your contribution. I think this code falls outside the scope of Synthea for a number of reasons.

First, Synthea does not directly handle external system integration. Please see this section of our wiki: https://github.com/synthetichealth/synthea/wiki/Contributing#external-systems-integrations

Second, the code is written in python. I have nothing against python and actually use it often. However, Synthea is primarily a java based project. To support this, we would have to update our install docs to include python and worry about another dependency.

The code is useful. I suggest that you put it in a GitHub gist. It could then be referred to in the project wiki for those who are interested in uploading to a FHIR server.

PR #1133, opened 2022-08-31
Provider update

jawalonoski:
Update organizational provider data from CMS.

Does not impact IHS or VA facilities, and should not impact Synthea-international repositories (i.e., changes should be backward compatible).

Quality data no longer is included, so the quality-based provider finder is eliminated.


jawalonoski:
image

hadleynet (approved):

PR #1157, opened 2022-09-23
Graphviz updates

eedrummer:
This should wait until #1144 is merged.

Updates the version of the graphviz library used by Synthea and makes code changes to deal with the new API.

This also cleans up some oddness discovered in HIV submodules related to date comparisons.

### DIFF
--- a/src/main/java/Graphviz.java
+++ b/src/main/java/Graphviz.java
@@ -20,11 +20,16 @@ import java.io.IOException;
 import java.io.StringReader;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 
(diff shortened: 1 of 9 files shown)


src/main/java/Graphviz.java
```diff
@@ -272,7 +270,7 @@ private static void generateJsonModuleGraph(JsonObject module, File outputFolder
 
     File outputFile = outputFolder.toPath().resolve(relativePath + ".png").toFile();
     outputFile.mkdirs();
-    guru.nidi.graphviz.engine.Graphviz.fromGraph(g).rasterizer(Rasterizer.BATIK)
```
dehall, line 275: I'm trying to remember why this was necessary in the first place... I think it may have been because the default rasterizer didn't work on CI? (Which was probably still Travis at the time) I do seem to recall that the default was a lot faster on my laptop though so if this still works on CI without it then that's a win. 
(Though I have to ask, is maintaining this graphviz support still necessary given that everybody uses the module builder?)

jawalonoski: > (Though I have to ask, is maintaining this graphviz support still necessary given that everybody uses the module builder?)

That depends on the definition of "necessary" and the definition of "everybody."

I think the static images are worthwhile. We post them in our gallery, which people still look at. The images can also be used for documentation/figures without resorting to manual screenshots.

jawalonoski (approved):

PR #1164, opened 2022-09-29
Payer Tweak

jawalonoski:
- Fix loading of plans when payer does not exist because it is out of state.


jawalonoski:
The test plan for this PR requires generating data in Alberta, Canada using the synthea-international configuration files.

RobertScalfani (approved):
I could also see a different approach where PayerManager:166 could include a check to see if the next plan it's about to load has a payer that exists and, only if it does, will execute the original csvLineToPlan() method. But I think this approach works just as well.

PR #1175, opened 2022-10-19
Update synthea.properties

aanchaltripathy:


jawalonoski:

However, our default output is only HL7 FHIR, because we believe that is the path forward for health interoperability.

We realize many people enjoy our CSV format, but it is not standards-based, so we leave that for each user to individually configure.  In other words, we do not want CSV to be a default.

Issue #1104, opened 2022-06-28
FHIR Resource Type Filter


jawalonoski:
### Requested Feature

Modify the FHIR exporters to only export specific resources, similar to how the CSV exporter can include/exclude specific files:

https://github.com/synthetichealth/synthea/blob/c266ab656670fb7b7506b892245060944e4d9576/src/main/resources/synthea.properties#L36-L43

willdalh:
Any update on this? 

I have found that I can hack a solution for this by commenting out certain parts in `org.mitre.synthea.export.FhirR4.java`. In my case the lines for `Claim` and `ExplanationOfBenefit`.

https://github.com/synthetichealth/synthea/blob/5d90e086985aecf69e8a1140d9793ad8f104fe73/src/main/java/org/mitre/synthea/export/FhirR4.java#L267-L337

Would you recommend any other way? Ideally the resources I want to omit should not be generated at all, but I guess there are some dependencies between these.

jawalonoski:
I don't believe anyone is actively working on this yet. If you are motivated enough to issue a Pull Request, we'd happily consider it.

The hack you are using is fine for a quick local change, but you are correct that there are dependencies... or at least, references between the resources. Ideally each time a reference is created, it would instead be conditionally created based on what resources are being included/exported. This would need some pretty good test cases though to validate the results (broken references would be invalid).

dehall:
Implemented in #1178

PR #1178, opened 2022-10-19
FHIR export resource type filter

dehall:
Addresses #1104

Adds new config settings to support filtering the resource types that are exported in the FHIR exporter, by either an include list or exclude list. (Similar to the existing feature for CSVs)
This approach checks each resource type before instantiating the relevant HAPI resource class.

A few notes on current status:
 - this leaves the Patient and Encounter resources as always included, even if specifically excluded, primarily because of the number of changes it would require.  
 - this doesn't currently attempt to be "friendly" and include related resources, for example a DiagnosticReport will reference Observations, but if you include DiagnosticReport without including Observation, you will get a DiagnosticReport with no outgoing references. 
 - this is only FHIR R4, but I'd like to get any feedback before applying this to the other exporters. (Or maybe we don't even want to spend time on the DSTU2 and STU3 exporters anymore?)
 - I'd also like to explore some changes to work with resource type enums instead of strings

This also includes a tweak to build.gradle to fix some error messages I was seeing in Eclipse. The issue was that JSBCL includes JUnit 4.10 as a transitive dependency, but we use JUnit 4.13 as a direct dependency for tests. Eclipse was getting confused and using JUnit 4.10 at compile time and it didn't recognize some method calls. The tweak removes the transitive dependency on JUnit from JSBCL because I don't think there's any need for the main build to include JUnit.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/FhirR4.java
+++ b/src/main/java/org/mitre/synthea/export/FhirR4.java
@@ -1891,15 +1891,16 @@ public class FhirR4 {
 
     if (!procedure.reasons.isEmpty()) {
       Code reason = procedure.reasons.get(0); // Only one element in list
-      for (BundleEntryComponent entry : bundle.getEntry()) {
-        if (entry.getResource().fhirType().equals("Condition")) {
-          Condition condition = (Condition) entry.getResource();
-          Coding coding = condition.getCode().getCoding().get(0); // Only one element in list
-          if (reason.code.equals(coding.getCode())) {
-            procedureResource.addReasonReference().setReference(entry.getFullUrl())
-                .setDisplay(reason.display);
-          }
-        }
+
+      BundleEntryComponent reasonCondition = findConditionResourceByCode(bundle, reason.code);
+      if (reasonCondition != null) {
+        procedureResource.addReasonReference()
+          .setReference(reasonCondition.getFullUrl())
+          .setDisplay(reason.display);
+      } else {
+        // we didn't find a matching Condition,
+        // fallback to just reason code
+        procedureResource.addReasonCode(mapCodeToCodeableConcept(reason, SNOMED_URI));
       }
     }
 
@@ -2207,16 +2208,16 @@ public class FhirR4 {
     if (!medication.reasons.isEmpty()) {
       // Only one element in list
       Code reason = medication.reasons.get(0);
-      for (BundleEntryComponent entry : bundle.getEntry()) {
-        if (entry.getResource().fhirType().equals("Condition")) {
-          Condition condition = (Condition) entry.getResource();
-          // Only one element in list
-          Coding coding = condition.getCode().getCoding().get(0);
-          if (reason.code.equals(coding.getCode())) {
-            medicationResource.addReasonReference()
-                .setReference(entry.getFullUrl());
-          }
-        }
+
+      BundleEntryComponent reasonCondition = findConditionResourceByCode(bundle, reason.code);
+      if (reasonCondition != null) {
+        medicationResource.addReasonReference()
+          .setReference(reasonCondition.getFullUrl())
+          .setDisplay(reason.display);
+      } else {
+        // we didn't find a matching Condition,
+        // fallback to just reason code
+        medicationResource.addReasonCode(mapCodeToCodeableConcept(reason, SNOMED_URI));
       }
     }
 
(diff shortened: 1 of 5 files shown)


dehall:
Marking this as ready to review per separate conversation -- Patient and Encounter will always be required, and if people like this approach and want it for FHIR DSTU2 or STU3 I'll copy it over, pending any other suggestions.

KevinCranmer:
@dehall this is awesome stuff that I would love to begin using! Would it be possible to include a filter for `Provenance` as well, or is that a required resource similar to Patient and Encounter? I'm decently new to fhir myself.

dehall:
@KevinCranmer thanks and good catch, that was an unintentional omission. I've now added a check for Provenance. I'll remind our team this is ready for review and hopefully we can get it merged soon

jawalonoski (approved):
The only thing I would say is that many reasons are now lost, even if they are known.

For example, assume we are exporting only `MedicationRequest` resources.

If a `MedicationRequest` has a reason, but `Condition` is filtered out, then `MedicationRequest.reasonReference` is not populated (as expected)... however, because the `reasonCode` is known, you should be able to populate the `MedicationRequest.reasonCode`.

Do you want to make a change now or in a separate PR?

src/main/java/org/mitre/synthea/export/FhirR4.java
```diff
@@ -191,6 +194,56 @@ public class FhirR4 {
...
+   *  from synthea.properties into a list of FHIR ResourceTypes.
+   * @param propString String directly from Config, ex "Patient,Condition , Procedure"
+   * @return normalized list of filenames as strings
+   */
+  private static List<String> propStringToList(String propString) {
+    List<String> files = Arrays.asList(propString.split(","));
+    // normalize filenames by trimming
+    files = files.stream().map(f ->  f.trim()).collect(Collectors.toList());
+
+    // TODO: convert these into FHIR resource enums
```
hadleynet, line 242: Instead of using an enum you could use the class. E.g. `Class clazz = Class.forName("org.hl7.fhir.r4.model." + className)`.
This function could return a `List<Class>` instead of `List<String>`
The `Class.forName` method can throw a ClassNotFound exception. You could use that to catch spelling errors in the list of resources to include/exclude.

src/main/java/org/mitre/synthea/export/FhirR4.java
```diff
@@ -267,75 +320,108 @@ public static Bundle convertToFHIR(Person person, long stopTime) {
     for (Encounter encounter : person.record.encounters) {
       BundleEntryComponent encounterEntry = encounter(person, personEntry, bundle, encounter);
 
-      for (HealthRecord.Entry condition : encounter.conditions) {
-        condition(person, personEntry, bundle, encounterEntry, condition);
+      if (shouldExport("Condition")) {
```
hadleynet, line 323: If you take my suggestion above about using classes instead of strings, this would become `shouldExport(Condition.class)` which would make it less prone to typo errors.

hadleynet (changes requested):
Just one suggestion about switching to using Classes instead of Strings for you lists of resources to include or exclude.

dehall:
@hadleynet I like the idea but there are a couple ways the Java code winds up being a little less elegant in practice. 1) To prevent compiler warnings, we can't use just `Class`, we need to parameterize it, which becomes `Class<? extends Resource>`. 2) Because of the overlap between FHIR resource type names and our HealthRecord class names, we can only import one class with the same name at the same time, so some of the `shouldExport(ClassName.class)` s become `shouldExport(org.hl7.fhir.r4.model.ClassName.class)`. 

Neither is the end of the world obviously but if people think this is ugly then I can try something else. I've pushed that change to a separate branch `resource_type_filter_with_classes` . Feel free to tell me I'm overthinking it and I'll move that back to this branch. See diff here: https://github.com/synthetichealth/synthea/compare/resource_type_filter...resource_type_filter_with_classes

hadleynet:
You might be able to get away with `Class<?>` instead of `Class<? extends Resource>` but the latter is more specific/correct.  If you don't like the `shouldExport(org.hl7.fhir.r4.model.ClassName.class)` you could switch which class you import and fully qualify the Synthea class name instead.

PersonalIy, think the type-safety aspect is worth the extra effort.

dehall:
Ok I've gone ahead and added the type-safe changes here. Rather than change more unrelated lines I'll keep the fully qualified FHIR class names where necessary

hadleynet (approved):

PR #1186, opened 2022-10-27
Update CCDA and remove MDHT

eedrummer:
This PR updates several aspects of CCDA export as well as testing.

For the CCDA export:
* Fixes the vital signs section. Previously, all `HealthRecord.Observation`s were dumped into the vital signs section, which was wrong. Now it is only `Observation`s with a category of `vital-sign`
* Fixes the Sepsis module so that problematic observations are no longer vital signs
* Implements the Functional Status Section, particularly the Assessment Scale Observation portion. All `HealthRecord.Observation`s with a category of `survey` will go here.

For CCDA testing:
* MDHT was removed. The library crashes for unexplainable reasons. Documents that fail validation would pass on subsequent runs through the tool.
* Testing now executes a series of XPath expressions to check to make sure that the mandatory sections are present
* `CodeResolveAndExportTest` had its CCDA related testing removed. The test scenario doesn't really make sense. The fix to vital signs broke the test, which was depending on bad behavior. Changing the test would ripple through export testing on the other formats, which isn't worth the effort.


src/main/java/org/mitre/synthea/export/CCDAExporter.java
```diff
@@ -88,15 +94,31 @@ public static String export(Person person, long time) {
...
+
+    person.attributes.put("ehr_vital_signs", vitalSigns);
+
+    List<Observation> surveyResults = superEncounter.observations
+            .stream()
+            .filter(vs -> vs.category != null && vs.category.equals("survey"))
+            .filter(vs -> vs.value != null && vs.value instanceof Double)
+            .collect(Collectors.toList());
+
+    // sadly, the correct plural of status is statuses and not stati
```
jawalonoski, line 118:
```
    // sadly, the correct plural of status is statuses and not stati
```
👍

src/test/java/org/mitre/synthea/export/CCDAExporterTest.java
```diff
@@ -27,19 +33,51 @@
 import org.mitre.synthea.world.agents.PayerManager;
 import org.mitre.synthea.world.agents.Person;
 import org.mitre.synthea.world.geography.Location;
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
 
 /**
- * Uses Model Driven Health Tools (MDHT) to validate exported CCDA R2.1.
- * https://github.com/mdht/mdht-models
+ * This is a rudimentary check of the CCDA Export. It uses XPath expressions to ensure that the
```
jawalonoski, line 40: > It uses XPath expressions

I am so sorry. You have my condolences.

src/main/resources/templates/ccda/ccda.ftl
```diff
@@ -132,11 +132,6 @@
       <#if ehr_careplans?has_content>
         <#include "care_goals.ftl">
       </#if>
-      <#if ehr_imaging_studies?has_content>
-        <#include "diagnostic_imaging_reports.ftl">
-      <#else>
-        <#include "diagnostic_imaging_reports_no_current.ftl" parse=false>
-      </#if>
```
jawalonoski, line 139: So, we're no longer using the `diagnostic_imaging*.ftl` templates at all?

Maybe we should just remove them?

jawalonoski (approved):

src/main/resources/templates/ccda/ccda.ftl
eedrummer: I'll remove those templates.

I removed them from CCDA output because the diagnostic imaging section is not one that is included in the CCD document template that we are using. The document templates list mandatory and optional sections. I wasn't able to easily find an answer as to whether including a section template outside of those listed was valid.

src/test/java/org/mitre/synthea/export/CCDAExporterTest.java
eedrummer: I honestly don't find XPath to be all that bad. The worst part of writing this code was getting the expressions into java strings and ensuring that I didn't exceed the line length checkstyle rule.

jawalonoski (approved):

PR #1188, opened 2022-11-01
RIF Deicide

hadleynet:
This PR breaks up the BB2RIFExporter class into a set of utility classes plus one class per RIF file type, all in a new package.


src/main/java/org/mitre/synthea/export/rif/BB2RIFExporter.java
```diff
@@ -0,0 +1,359 @@
...
+    exportCounts.put(EXPORT_SUMMARY.PDE_CLAIMS,
+            Long.toString(new PDEExporter(startTime, stopTime, this).export(person)));
+    exportCounts.put(EXPORT_SUMMARY.DME_CLAIMS,
+            Long.toString(new DMEExporter(startTime, stopTime, this).export(person)));
+    exportCounts.put(EXPORT_SUMMARY.HHA_CLAIMS,
+            Long.toString(new HHAExporter(startTime, stopTime, this).export(person)));
+    exportCounts.put(EXPORT_SUMMARY.HOSPICE_CLAIMS,
+            Long.toString(new HospiceExporter(startTime, stopTime, this).export(person)));
+    exportCounts.put(EXPORT_SUMMARY.SNF_CLAIMS,
+            Long.toString(new SNFExporter(startTime, stopTime, this).export(person)));
```
jawalonoski, line 336: Classically (potentially still) the most expensive operation in Java was `new` Object creation...

We'd probably get better performance at scale if you create the `*Exporter` classes once, and just reuse them.

src/main/java/org/mitre/synthea/export/rif/RIFExporter.java
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Base class for all of the type-specific exporters.
+ */
+public abstract class RIFExporter {
+
+  protected static final String BB2_PARTD_CONTRACTS = "BB2_PARTD_CONTRACTS";
+  protected static final AtomicLong nextCarrClmCntlNum = new AtomicLong(Config.getAsLong(
+          "exporter.bfd.carr_clm_cntl_num_start", -1));
+  protected static final AtomicLong nextPdeId = new AtomicLong(Config.getAsLong(
+          "exporter.bfd.pde_id_start", -1));
```
jawalonoski, line 28: Seems weird that some of these are in here... might make more sense to put things like `nextCarrClmCntlNum` or `nextPdeId` (possibly others) in the specific exporter classes that inherit from this base class.

jawalonoski (commented):
Not finished reviewing but here is round 1 of feedback:

- questions about class and method visibility
- package organization could be better (too many files?)
- creation of `new` exporters for every patient seems like it would have performance impacts
  - quick test results shows about ~29 seconds over 10K patients, so not much.
- some class attributes probably don't belong in the abstract RIFExporter base class.


Potentially more feedback later.

src/main/java/org/mitre/synthea/export/rif/BeneficiaryExporter.java
```diff
@@ -0,0 +1,471 @@
...
+   * @param person the person to export
+   * @return the beneficiary ID
+   * @throws IOException if something goes wrong
+   */
+  public String export(Person person) throws IOException {
+    String beneIdStr = Long.toString(RIFExporter.nextBeneId.getAndDecrement());
+    person.attributes.put(RIFExporter.BB2_BENE_ID, beneIdStr);
+    String hicId = RIFExporter.nextHicn.getAndUpdate((v) -> v.next()).toString();
+    person.attributes.put(RIFExporter.BB2_HIC_ID, hicId);
+    String mbiStr = RIFExporter.nextMbi.getAndUpdate((v) -> v.next()).toString();
```
jawalonoski, line 49: It is kind of unfortunate that the HICN and MBI need to pass a "lambda" to the `getAndUpdate` method, rather than do something like `RIFExporter.nextBeneId.getAndDecrement()`... but I suppose it can't be avoided unless you want to just wrap that in a method call.

src/main/java/org/mitre/synthea/export/rif/DMEExporter.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+
+      fieldValues.clear();
+      exporter.staticFieldConfig.setValues(fieldValues, BB2RIFStructure.DME.class, person);
+
+      // complex fields that could not easily be set using cms_field_values.tsv
+      fieldValues.put(BB2RIFStructure.DME.CLM_ID, "" + claimId);
+      fieldValues.put(BB2RIFStructure.DME.CLM_GRP_ID, "" + claimGroupId);
+      fieldValues.put(BB2RIFStructure.DME.CARR_CLM_CNTL_NUM, "" + carrClmId);
```
jawalonoski, line 64: Looking at:

```java
    long carrClmId = RIFExporter.nextCarrClmCntlNum.getAndDecrement();
```
and then

```java
     fieldValues.put(BB2RIFStructure.DME.CARR_CLM_CNTL_NUM, "" + carrClmId);
```

I wondered if we're doing this correctly... so I looked up `BB2RIFStructure.DME.CARR_CLM_CNTL_NUM` and couldn't find a definition for it anywhere.

jawalonoski: I'm wondering if this is supposed to reference an actual carrier claim, meaning whether or not it should have actual referential integrity (which this approach will not satisfy).

jawalonoski (commented):
Done reviewing this iteration.

src/main/java/org/mitre/synthea/export/rif/DMEExporter.java
hadleynet: I agree that we should look into this but it seems out of scope for this PR since this functionality has not been changed.

src/main/java/org/mitre/synthea/export/rif/BeneficiaryExporter.java
```diff
@@ -0,0 +1,471 @@
...
+          for (int i: coveredMonths) {
+            fieldValues.put(BB2RIFStructure.beneficiaryPartCContractFields[i - 1],
+                    partCContractIDStr);
+            fieldValues.put(BB2RIFStructure.beneficiaryPartCPBPFields[i - 1],
+                    partCPBPIDStr);
+          }
+        }
+      }
+
+      // TODO: make claim copay match the designated cost sharing code
```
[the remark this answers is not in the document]
hadleynet: I've moved the comment to the PDE exporter.

src/main/java/org/mitre/synthea/export/rif/RIFExporter.java
hadleynet: I've moved the ids that are specific to a particular file to the corresponding exporter. General ids that are used by multiple exporters I've left in the base class.

src/main/java/org/mitre/synthea/export/rif/BB2RIFExporter.java
hadleynet: Fixed.

src/main/java/org/mitre/synthea/export/rif/FixedLengthIdentifier.java
```diff
@@ -0,0 +1,101 @@
+package org.mitre.synthea.export.rif;
```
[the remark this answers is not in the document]
hadleynet: Agree but will leave it in the rif.identifiers package for now.

src/main/java/org/mitre/synthea/export/rif/ConsolidatedClaimLines.java
```diff
@@ -0,0 +1,91 @@
...
+import org.mitre.synthea.world.agents.Clinician;
+import org.mitre.synthea.world.concepts.Claim;
+import org.mitre.synthea.world.concepts.HealthRecord.Encounter;
+
+/**
+ * Utility class for consolidating multiple claim lines of the same type for the same clinician.
+ */
+class ConsolidatedClaimLines extends Claim.ClaimCost {
+
+  static class ConsolidatedClaimLine extends Claim.ClaimCost {
```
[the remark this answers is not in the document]
hadleynet: fixed

jawalonoski (approved):

PR #1206, opened 2022-11-22
RIF Hospice Consolidation

hadleynet:
- Break up RIF `HospiceExporter.export` function to make the flow clearer
- Break up RIF `SNFExporter.export` function to make the flow clearer
- Add claim line consolidation to Hospice export to reduce the number of claim lines per claim
- Fix deprecation warning re `new Double(0.0)`
- Fix gradle deprecation warning re dependency version fixing

Note: requires new hospice_rev_cntr_code_map.json file available from the usual places.


jawalonoski (approved):
Reviewed and tested.

In my test, 10k beneficiaries had 583 hospice claims, with an average line item count of 14.7.

Issue #1171, opened 2022-10-04
BFD: Hospice Length of Stay


jawalonoski:
## Actual Hospice Length of Stay
![image](https://user-images.githubusercontent.com/1054765/193834809-8fbb89a1-a614-4447-802e-14969be83a6e.png)

## Synthea Hospice Length of Stay
![image](https://user-images.githubusercontent.com/1054765/193834871-df5fb4d7-11b6-4d29-9423-a012b139fb0e.png)

jawalonoski:
Latest revisions...
<img width="476" alt="image" src="https://user-images.githubusercontent.com/1054765/223241237-933e26e0-cedd-48f8-9435-e339c70671d9.png">

Issue #1170, opened 2022-10-04
BFD: SNF Length of Stay


jawalonoski:
## Actual SNF Length of Stay
![image](https://user-images.githubusercontent.com/1054765/193834548-ae11aac0-5828-4f1b-bd2c-5aef921e90da.png)

## Synthea SNF Length of Stay
![image](https://user-images.githubusercontent.com/1054765/193834621-9bd8ac3b-c8e6-415a-96df-f75cfaf7017c.png)

jawalonoski:
Latest revisions...
<img width="476" alt="image" src="https://user-images.githubusercontent.com/1054765/223242685-62f6f85d-8211-4119-8a86-58571a81c2cc.png">

Issue #1169, opened 2022-10-04
BFD: Inpatient Length of Stay


jawalonoski:
## Actual Inpatient Length of Stay
![image](https://user-images.githubusercontent.com/1054765/193834242-be387c6d-7647-4782-a4a1-1efda2583081.png)

## Current Synthea Inpatient Length of Stay
![image](https://user-images.githubusercontent.com/1054765/193834334-af9d9278-5749-4e1c-9d0d-c2726c0cd168.png)

jawalonoski:
Current state from #1238 

<img width="572" alt="image" src="https://user-images.githubusercontent.com/1054765/212158779-1f5764ea-58d3-494e-a3fb-2101ab11bc8e.png">

jawalonoski:
Latest revisions... RIF output (line item==1, rev_center ==1)
<img width="476" alt="image" src="https://user-images.githubusercontent.com/1054765/223245034-b111be33-4b3c-49f6-b20c-9a1fe82a106a.png">

jawalonoski:
These improvements have been incorporated into `master` and the data has been generated and deployed.

Issue #1168, opened 2022-10-04
BFD: Distribution


jawalonoski:
### Requested Feature

We need a better distribution in GMF for determining length of stay.

Some type of exponential distribution, possibly Poisson or Geometric...

PR #1218, opened 2022-12-08
Add exponential distribution for Length of Stay

jawalonoski:
This pull request adds the [exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution) [Wikipedia] to the list of available distributions.

This distribution can be used for determining "length of stay", for instance.

Addresses #1168 for the purposes of #1169, #1170, #1171


jawalonoski:
Using the new distributions...

## SNF Length of Stay
image

## Hospice Length of Stay
This currently has a pretty large spike at 1-day that needs to be accounted for.

image

jawalonoski:
The spike in zero/one day hospice encounters in the graph above is due to https://github.com/synthetichealth/synthea/issues/1221

PR #1219, opened 2022-12-12
Flexporter

dehall:
This PR introduces the FHIR Flexible Exporter, aka "Flexporter", seeking public comment. Please let us know if this feature is useful to you, or what we can do to make it more useful.

We get a lot of requests from people who want to use Synthea, but they need some slight changes to the data. For example, they want Appointment resources for every Encounter or another resource type, or they want to filter the data in a certain way, or they want certain fields set on certain resources, or they may want to ensure the output conforms to the IG they are building or working with. The flexporter is meant to allow to allow users to make these kinds of changes to Synthea output without making major changes to the underlying Synthea engine and/or all the modules.

The basis of the Flexporter is a YAML mapping file which contains a series of actions to apply to FHIR bundles as they are exported from Synthea.   Full details on all the different actions supported by the flexporter, and how to use them, are on the wiki: https://github.com/synthetichealth/synthea/wiki/Flexporter

From a technical perspective, the Flexporter relies heavily on FHIRPath, in particular HAPI's implementation of the FHIRPath engine.  A community contribution to HAPI introduced the FHIRPathResourceGenerator, which I've adopted and modified for the needs of the flexporter. Basically this class takes a mapping of `{ FHIRPath: value, ... }` to populate resources. The most interesting features of the flexporter, including creating new resources and setting values on existing resources, depend on this feature.
There is also a JavaScript engine allowing for arbitrary scripting (though external libraries are not supported)


hadleynet (commented):
I built, tested and ran it locally with a mapping file I created by following the wiki instructions.

Code is well structured and there are sufficient tests.

Only a couple of things jumped out at me:
1. Several Javadoc comments with embedded <> that will be treated as HTML, need to wrap these with {@code ...}
2. The data structures used in RandomCodeGenerator seem unnecessarily generic. E.g. codeListCache is a Map<String, List<Object>> but the individual list items are typically just cast to Map<String, String>. Consider making this explicit in the declaration of codeListCache as Map<String, List<Map<String, String>>> to avoid all the casts or introduce a new class that inherits from HashMap or something to make the meaning clearer and enforce map keys. E.g. you could use Map<String, List<CodeMap>> whether CodeMap extends or contains a HashMap<String, String> but offers explicit setters and getter for code, code system etc.

jawalonoski (approved):
Below was the test case I used... fields that are "choice" type can be a gotcha when writing FHIR Path (e.g., `Procedure.performed[x]`)... and all the warning messages on stdout are annoying, but otherwise this is good to go.

```yaml
---
name: Add category to procedures that don't have one

applicability: true

actions:
  - name: Category
    set_values:
      - applicability: Procedure.where(category.empty())
        fields:
          - location: Procedure.category.coding
            value:
              system: http://mitre.org
              code: "XYZZY"
              display: "A flexi category"
  - name: Patient Profile
    profiles:
      - profile: http://fake.org/fhir/Patient/MALE
        applicability: Patient.where(gender="male")
      - profile: http://fake.org/fhir/Patient/FEMALE
        applicability: Patient.where(gender="female")
  - name: testCreateResources_createBasedOn
    create_resource: 
      - resourceType: ServiceRequest
        based_on:
          resource: Procedure
        fields:
          - location: ServiceRequest.intent
            value: plan
          # - location: ServiceRequest.encounter.reference
          #   value: $getField([Procedure.encounter.reference])
          - location: ServiceRequest.authoredOn
            value: $getField([Procedure.performed.start])
          - location: ServiceRequest.subject.reference
            value: $findRef([Patient])
        writeback:
          - location: Procedure.basedOn.reference
            value: $setRef([ServiceRequest])
  - name: testKeepResources
    keep_resources:
      - Patient
      - ServiceRequest
```

PR #1225, opened 2022-12-20
Utility to map from the BB2 RIF format to the CCW RIF format.

hadleynet:


src/main/java/org/mitre/synthea/export/rif/tools/BB2RIF2CCW.java
```diff
@@ -0,0 +1,110 @@
...
+      CsvMapper mapper = new CsvMapper();
+      CsvSchema schema = CsvSchema.emptySchema().withHeader().withColumnSeparator('|');
+      MappingIterator<LinkedHashMap<String, String>> sourceRows = mapper
+              .readerFor(LinkedHashMap.class).with(schema).readValues(file);
+      boolean firstOutputRow = true;
+      SequenceWriter writer = null;
+      File outputFile = outputDir.toPath().resolve(file.getName()).toFile();
+      while (sourceRows.hasNextValue()) {
+        LinkedHashMap<String, String> outputRow = transformRow(sourceRows.next(), nameMap);
+        if (firstOutputRow) {
```
jawalonoski, line 59: Minor point, but this will only write out the headers (and the file) if the RIF file contains at least one record.

src/main/java/org/mitre/synthea/export/rif/tools/BB2RIF2CCW.java
```diff
@@ -0,0 +1,110 @@
...
+    csv.forEach(entry -> {
+      map.put(entry.get("BB2"), entry.get("CCW"));
+    });
+    return map;
+  }
+
+  private static File[] getSourceFiles(String filePrefix, File inputDir) {
+    if (filePrefix.equals("beneficiary")) {
+      return inputDir.listFiles((dir, filename) -> {
+        return filename.startsWith(filePrefix) && !filename.contains("history");
```
jawalonoski, line 102: Since the `export/beneficiary_history_bb2_ccw.csv` doesn't exist, I don't think this `!filename.contains("history")` boolean condition is necessary.

jawalonoski (commented):
I don't feel strongly about either of these two comments and I could be persuaded to merge it as-is.

Issue #1224, opened 2022-12-19
Using modified demographics files with synthea-with-dependencies.jar


m-goggins:
### What happened?

Hello!

I am trying to use the downloaded synthea jar file (v3.1.1) on Windows 11 Pro with Java version 11.0.16.1 and modifying the following demographic files:

demographics.default_file
geography.zipcodes.default_file
geography.timezones.default_file
geography.sdoh.default_file
When I try passing in params like:
java -jar synthea-with-dependencies.jar -p 10 --generate.demographics.default_file="C:\test\modified_demographics.csv"

I get the following errors:

```
ERROR: unable to load demographics
java.lang.IllegalArgumentException: resource C:\test\modified_demographics.csv not found.
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
        at com.google.common.io.Resources.getResource(Resources.java:198)
        at org.mitre.synthea.engine.Generator.init(Generator.java:226)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:208)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:184)
        at App.main(App.java:229)
Caused by: java.lang.IllegalArgumentException: resource C:\test\modified_demographics.csv not found.
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
        at com.google.common.io.Resources.getResource(Resources.java:198)
        at org.mitre.synthea.helpers.Utilities.readResource(Utilities.java:410)
        at org.mitre.synthea.world.geography.Demographics.load(Demographics.java:388)
        at org.mitre.synthea.world.geography.Location.<init>(Location.java:64)
        at org.mitre.synthea.engine.Generator.init(Generator.java:226)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:208)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:184)
        at App.main(App.java:229)
Exception in thread "main" java.lang.ExceptionInInitializerError
        at org.mitre.synthea.world.geography.Location.<init>(Location.java:102)
        at org.mitre.synthea.engine.Generator.init(Generator.java:226)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:208)
        at org.mitre.synthea.engine.Generator.<init>(Generator.java:184)
        at App.main(App.java:229)
Caused by: java.lang.IllegalArgumentException: resource C:\test\modified_demographics.csv not found.
[...]

PR #1226, opened 2022-12-28
Allow config settings that are usually in resources to reference any file location

dehall:
Addresses #1224 

As of today we have a number of config settings which implicitly point to locations under `src/main/resources`, but there's no technical reason they necessarily have to. But as shown in the linked issue, if you try to point to an absolute file path somewhere else, it doesn't work.

This PR refactors the `Utilities.readResource` family of methods into 4:

`Utilities.readResource(String, boolean, boolean)`:
  Method containing all the logic and configurability. Allow for optionally reading from different locations and optionally stripping BOM.

`Utilities.readResource(String)`:
  Default method to match the previous implementation, calls the above with the options set to false.

`Utilities.readResourceOrPath(String)`:
  Method to allow for reading from either src/main/resources, or from a free path on the filesystem. Calls `readResource(filename, false, true)`

`Utilities.readResourceAndStripBOM(String)`:
  Method to allow for reading from only src/main/resources, and strips BOM if present. Calls `readResource(filename, true, false)`

Then I updated references to these methods by the following rules:

1. If the file we read is a CSV, we should strip BOM.
2. If the file we read comes from a config setting, we should allow free paths. (Aka, if the file path is hardcoded, don't bother allowing free paths)
3. If both the above are true, use `Utilities.readResource(String, boolean, boolean)`
4. If neither is true, use `Utilities.readResource(String)` to minimize the number of changes
5. Otherwise use the appropriate named method for clarity

Note: I have not yet tested this with all combinations, such as via gradle vs via uberJar, on Mac vs Linux vs Windows, etc. We should confirm this with as many combinations as we can think of.

### DIFF
--- a/src/main/java/org/mitre/synthea/world/agents/Provider.java
+++ b/src/main/java/org/mitre/synthea/world/agents/Provider.java
@@ -449,7 +449,7 @@ public class Provider implements QuadTreeElement, Serializable {
       return;
     }
 
-    String resource = Utilities.readResourceOrPath(filename);
+    String resource = Utilities.readResource(filename, true, true);
     Iterator<? extends Map<String,String>> csv = SimpleCSV.parseLineByLine(resource);
 
     while (csv.hasNext()) {
--- a/src/main/java/org/mitre/synthea/world/agents/Provider.java
+++ b/src/main/java/org/mitre/synthea/world/agents/Provider.java
@@ -434,7 +434,7 @@ public class Provider implements QuadTreeElement, Serializable {
    * THIS method is for loading providers and generating clinicians with specific specialties
    *
    * @param location the state being loaded
-   * @param filename Location of the file, relative to src/main/resources
+   * @param filename Location of the file
    * @param providerType ProviderType
    * @param servicesProvided Set of services provided by these facilities
    * @param random Source of randomness for provider generation
@@ -449,7 +449,7 @@ public class Provider implements QuadTreeElement, Serializable {
       return;
     }
 
-    String resource = Utilities.readResource(filename);
+    String resource = Utilities.readResourceOrPath(filename);
     Iterator<? extends Map<String,String>> csv = SimpleCSV.parseLineByLine(resource);
 
     while (csv.hasNext()) {
--- a/src/main/java/org/mitre/synthea/world/geography/Location.java
+++ b/src/main/java/org/mitre/synthea/world/geography/Location.java
@@ -104,7 +104,7 @@ public class Location implements Serializable {
     String filename = null;
     try {
       filename = Config.get("generate.geography.zipcodes.default_file");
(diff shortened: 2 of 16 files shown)


src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -449,7 +449,7 @@ public static void loadProviders(Location location, String filename,
       return;
     }
 
-    String resource = Utilities.readResource(filename);
+    String resource = Utilities.readResourceOrPath(filename);
```
jawalonoski, line 452: This is a CSV file, so should it be `readResource(filename, true, true)`?

src/main/java/org/mitre/synthea/world/geography/Location.java
```diff
@@ -405,7 +405,7 @@ private static LinkedHashMap<String, String> loadAbbreviations() {
     String filename = null;
     try {
       filename = Config.get("generate.geography.zipcodes.default_file");
-      String csv = Utilities.readResource(filename);
+      String csv = Utilities.readResourceOrPath(filename);
```
jawalonoski, line 408: This should also be `readResource(filename,  true, true)` just as in Line 107.

src/main/java/org/mitre/synthea/world/geography/Location.java
```diff
@@ -466,7 +466,7 @@ private static Map<String, String> loadTimezones() {
     String filename = null;
     try {
       filename = Config.get("generate.geography.timezones.default_file");
-      String csv = Utilities.readResource(filename);
+      String csv = Utilities.readResourceOrPath(filename);
```
jawalonoski, line 469: Again, this is a CSV file.... so `readResource(filename,  true, true)`?

src/test/java/org/mitre/synthea/world/geography/LocationTest.java
```diff
@@ -102,11 +102,11 @@ public void testTimezone() {
   @Test
   public void testAllDemographicsHaveLocations() throws Exception {
     String demoFileContents =
-        Utilities.readResource(Config.get("generate.demographics.default_file"));
+        Utilities.readResourceOrPath(Config.get("generate.demographics.default_file"));
     List<LinkedHashMap<String, String>> demographics = SimpleCSV.parse(demoFileContents);
 
     String zipFileContents =
-        Utilities.readResource(Config.get("generate.geography.zipcodes.default_file"));
+        Utilities.readResourceOrPath(Config.get("generate.geography.zipcodes.default_file"));
```
jawalonoski, line 109: Technically, these are both CSV files, so should we use `readResource(filename,  true, true)`?

jawalonoski (commented):
Not sure about these, but the usage doesn't look consistent...

jawalonoski (approved):

PR #1233, opened 2023-01-09
Make LocalDate/timestamp conversions more consistent

dehall:
This PR tweaks the logic in conversions between LocalDate and timestamps to make them more consistent. These inconsistencies were actually the root cause of the test failures reported by a number of people due to a FileNotFoundException in the EntityManagerTest.

Short version of the root cause is that the date logic in `Utilities.localDateToTimestamp` and `Entity.seedAt` was not symmetric. 
`localDateToTimestamp` converts a LocalDate (ex, 2023-01-03) into a timestamp (1672782196746), and `seedAt` converts a timestamp into a LocalDate to check which range that time is in. The misalignment here results in seedAt returning null, which then causes a NullPointerException in generating this patient, which then causes the json output folder to not be created so the test throws a FileNotFoundException.

This fix ensures the conversions are always done in the context of UTC, not the system timezone. (Note: I don't actually think it matters which one we use, as long as we are consistent in all conversions. Other conversions between human-readable dates and timestamps use UTC, for example the other methods in the `Utilities` class and the `Date` GMF logic)

That EntityManagerTest test appeared to be flaky because we never saw it ourselves, but depending on time of day and local time zone, it could fail. Since the core development team is in Eastern time and runs Synthea primarily during business hours, we were never running the test at a time that the local date is different from the date in UTC time. I was able to reproduce it right away with a Docker container running the test in an Australian time zone


src/test/java/org/mitre/synthea/identity/EntityTest.java
```diff
@@ -53,7 +53,8 @@ public void seedAtTimestamp() {
         .atStartOfDay().toInstant(ZoneOffset.UTC).toEpochMilli();
     seed = testEntity.seedAt(duringOpenEnd);
     assertEquals("1416", seed.getSeedId());
-    long startOfFirstSeed = -813960000000L;
+    long startOfFirstSeed = -813960000000L; // Fri Mar 17 1944 04:00:00 GMT+0000
+    // Note this timestamp is 1944-03-17 in EDT (GMT-4) but 1944-03-16 in EST (GMT-5)
```
jawalonoski, line 57: https://en.wikipedia.org/wiki/March_1944#March_17,_1944_(Friday)

jawalonoski (approved):
This looks good to me.

@eedrummer do you think this will negatively effect Entity / FixedRecords?

PR #1234, opened 2023-01-09
BFD missing code map report

hadleynet:
Track Synthea codes that can't be mapped to a suitable BFD code and export a CSV report with: map file, code, code description, and number of times the code was requested.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/FhirR4.java
+++ b/src/main/java/org/mitre/synthea/export/FhirR4.java
@@ -1675,18 +1675,20 @@ public class FhirR4 {
             new AllergyIntolerance.AllergyIntoleranceReactionComponent();
         reactionComponent.addManifestation(mapCodeToCodeableConcept(manifestation, SNOMED_URI));
         HealthRecord.ReactionSeverity severity = allergy.reactions.get(manifestation);
-        switch (severity) {
-          case MILD:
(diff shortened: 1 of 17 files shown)


src/main/java/org/mitre/synthea/export/FhirR4.java
```diff
@@ -1675,18 +1675,20 @@ private static BundleEntryComponent allergy(RandomNumberGenerator rand,
...
-            break;
-          case MODERATE:
-            reactionComponent.setSeverity(AllergyIntolerance.AllergyIntoleranceSeverity.MODERATE);
-            break;
-          case SEVERE:
-            reactionComponent.setSeverity(AllergyIntolerance.AllergyIntoleranceSeverity.SEVERE);
-            break;
-          default:
-            // do nothing
+        if (severity != null) {
```
jawalonoski, line 1678: Adding this `null` check makes sense... but do you know which modules were creating allergies without severities? Maybe we should fix those.

jawalonoski (commented):
Still reviewing...

src/main/java/org/mitre/synthea/export/FhirR4.java
hadleynet: I don't *think* there are any, this was flagged by the sonatype-lift bot so I fixed it preemptively.

jawalonoski (approved):

PR #1235, opened 2023-01-10
RIF Bene File Splitter

hadleynet:
Add new gradle task to split original 3 file BFD bene export into one bene file per year.

Usage: `./gradlew rifBeneSplit -Pargs="inputDir"`

Split files will be written to the same directory. The splitter will work either before or after conversion from BFD RIF to CCW RIF.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/rif/tools/BB2RIFBeneSplitter.java
+++ b/src/main/java/org/mitre/synthea/export/rif/tools/BB2RIFBeneSplitter.java
@@ -1,0 +1,89 @@
+package org.mitre.synthea.export.rif.tools;
+
+import com.fasterxml.jackson.databind.MappingIterator;
+import com.fasterxml.jackson.databind.SequenceWriter;
+import com.fasterxml.jackson.dataformat.csv.CsvMapper;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mitre/synthea/export/rif/tools/BB2RIFBeneSplitter.java
```diff
@@ -0,0 +1,89 @@
...
+   * @throws java.io.IOException if something goes wrong
+   */
+  public static void main(String[] args) throws IOException {
+    if (args.length != 1) {
+      System.out.println("Usage ./gradlew rifBeneSplit -Pargs=\"inputDir\"");
+      System.exit(-1);
+    }
+    Map<String, SequenceWriter> writers = new HashMap<>();
+    for (String inputFileName: inputFiles) {
+      File inputFile = Path.of(args[0]).resolve(inputFileName).toFile();
```
jawalonoski, line 46: Not an issue worth stopping this pull request, but this path code was not working for me when the path contained whitespace (possibly an escaping issue on the command line) or when the path contained a tilde (`~`) to the user's home folder.

Issue #1236, opened 2023-01-10
Runs with same seed are not producing identical results


KevinCranmer:
### What happened?

Reading the wiki it looks like "Populations generated with the same seed and the same version of Synthea should be identical". 

But I've noticed that this is not the case. I'm consistently seeing the same 3 types of differences across multiple runs on the same Synthea version.

1. Array ordering: https://www.diffchecker.com/7pX3yd1w/
   - not a big issue for my use case
2. DocumentReference data: https://www.diffchecker.com/i3UikbSa/
   - more of an issue. The key difference, when decoded, is one line saying "Patient currently has Aetna." vs "Patient currently has Humana."
3. Actual IDs changing: https://www.diffchecker.com/OFqRg6FW/
   - big issue for my use case as I can no longer trust IDs are the same across multiple runs.

Command I'm using:
`./run_synthea -s 1668660039331 -cs 1668660039331 -r 20230101 -p 83 Florida`

### Environment

```markdown
- OS: macOS 12.6
- Java: openjdk version 16.0.2
```

jawalonoski:
This is a bug that we'll look at as a high priority.

The DocumentReference and IDs will likely be fixable.

The Array ordering is probably not something we can fix -- because the FHIR JSON export is done by an underlying dependency -- but we'll take a look.

dehall:
@KevinCranmer How frequently are you seeing the issue of resource IDs changing between runs? I'm consistently able to reproduce the first two issues (array ordering and insurance plan) and I'm working on fixes, but across a dozen tests I haven't seen any instances of resource IDs being different across runs. (Though I have found a separate issue of ImagingStudy DICOM UIDs occasionally differing)

I'm wondering if I'm missing something or if this is just an exceptionally rare case we need to chase down

KevinCranmer:
@dehall It seems like the IDs changing only happens if I re-clone Synthea and run the same command. I'm cloning the same commit each time, so the Synthea versions would be the same.

But this would mean if two machines were to both clone and run Synthea, they should experience the ID changing difference.

Do you see this ID difference if you re-clone?

dehall:
[...]

PR #1237, opened 2023-01-11
Reproducibility fixes

dehall:
Addresses #1236 and fixes some instances where running the same population multiple times produces different results

Issues identified and fixed so far:

1. The ordering of allergy reactions
2. DocumentReference content and Insurance plan selection

(Essentially both of these are the result of using Sets where order is not guaranteed. The current fix is not ideal, it does some extra sorting at the end, so there may be a more "root" fix further upstream. Note that I was not able to reproduce this second issue when running with only a single thread, which also suggests to me further analysis is required)

3. Resource UUIDs are occasionally different
This is because the UUIDs are generated using the same random number generator as used during generation. When the simulation is re-run with a different end date, the generation process may tick up an extra time stamp and roll a few extra random numbers. The fix here is to essentially remove all randomness from the FhirR4 exporter, done in two ways:
A) Select UUIDs during the generation process, by assigning them to every `HealthRecord.Entry` at creation time
B) For FHIR resources that don't align 1:1 with a synthea model class, use a new method `ExportHelper.buildUUID` to consistently create a UUID based on the Person, time, and a "key" for uniqueness. A few examples are the Provenance added to the end of each Bundle, DiagnosticRequests and DocumentReferences created for clinical notes, and ExplanationOfBenefits.

4. DICOM UIDs are occasionally different
This is because the Series object and the Instances it contains are being unintentionally leaked between patients. (Also doesn't occur when running with only a single thread.) I've put in a fix to ensure objects are always cloned, but I suspect there might be a more minimal fix possible. The "cleanest" would be an extra bit in `ImagingStudy.clone()` to match, for example, `ObservationGroup.clone()` where the observations in a list are individually cloned, but that's tougher in this case because we want certain instances to have a different/random number of objects in the list.

5. Differences across OSes
Observed on MacOS vs a Linux Docker image, likely to occur across other combinations as well. Patients with the same config were producing wildly different results. This was because in a couple places we were selecting a random index based on a HashMap, where sort order is not guaranteed. Seems like order is consistent for a given OS but still we shouldn't even bank on that. I replaced instances of HashMap with TreeMap for a minimal change that should ensure a consistent order.

Not currently implemented as part of this PR, I suggest we implement these separately:
 - FHIR 2 & 3 updates to match R4
 - CSV exporter to use the existing UUIDs
 - Update any other exporter that generates UUIDs
(Do we want these to be part of this PR?)

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Module.java
+++ b/src/main/java/org/mitre/synthea/engine/Module.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.TreeMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -66,7 +67,7 @@ public class Module implements Cloneable, Serializable {
   private static final Map<String, ModuleSupplier> modules = loadModules();
 
   private static Map<String, ModuleSupplier> loadModules() {
-    Map<String, ModuleSupplier> retVal = new ConcurrentHashMap<>();
+    Map<String, ModuleSupplier> retVal = new TreeMap<>();
     int submoduleCount = 0;
 
     retVal.put("Lifecycle", new ModuleSupplier(new LifecycleModule()));
--- a/src/main/java/org/mitre/synthea/engine/State.java
+++ b/src/main/java/org/mitre/synthea/engine/State.java
@@ -1938,13 +1938,16 @@ public abstract class State implements Cloneable, Serializable {
     @Override
     public boolean process(Person person, long time) {
       // Randomly pick number of series and instances if bounds were provided
-      duplicateSeries(person, time);
-      duplicateInstances(person, time);
+      // Don't modify the instance series as it gets reused - create a local copy
+      List<HealthRecord.ImagingStudy.Series> mySeries = duplicateSeries(person, time);
+      duplicateInstances(person, time, mySeries);
 
       // The modality code of the first series is a good approximation
       // of the type of ImagingStudy this is
-      String primaryModality = series.get(0).modality.code;
-      entry = person.record.imagingStudy(time, primaryModality, series);
+      String primaryModality = mySeries.get(0).modality.code;
+      entry = person.record.imagingStudy(time, primaryModality, mySeries);
+      // note the imagingStudy call will choose dicomUIDs,
+      // so they don't need to be selected here
 
       // Add the procedure code to the ImagingStudy
       String primaryProcedureCode = procedureCode.code;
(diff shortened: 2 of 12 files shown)


dehall:
For posterity, I used this little script to assist in testing
```sh
#!/bin/bash

rm -rf output/fhir output/fhir_* log_*.txt

for s in 1 2 3
do
  echo running $s

  ./run_synthea -s 1668660039331 -cs 1668660039331 -r 20230101 -p 83 Florida > log_${s}.txt

  mv output/fhir output/fhir_${s}

done

cd output

for file in `ls fhir_1`; do echo $file; diff fhir_1/$file fhir_2/$file ; done
```

src/main/java/org/mitre/synthea/engine/Module.java
```diff
@@ -66,7 +67,7 @@ public class Module implements Cloneable, Serializable {
   private static final Map<String, ModuleSupplier> modules = loadModules();
 
   private static Map<String, ModuleSupplier> loadModules() {
-    Map<String, ModuleSupplier> retVal = new ConcurrentHashMap<>();
+    Map<String, ModuleSupplier> retVal = new TreeMap<>();
```
dehall, line 70: Note to reviewers: this change looks scary because it's no longer "concurrent", but we were never actually writing anything to the map outside of this function in regular usage.  I wanted to replace the `return retVal;` at the end with `return Collections.unmodifiableMap(retVal);` to make that obvious, but it did break unit tests that modify the map for a specific test reason.

dehall:
Again for posterity, some more tidbits I used to help debug the native vs Docker differences:

Dockerfile
```dockerfile
FROM gradle:jdk19

ENV PLATFORM="docker"
ENV TZ="America/New_York"
COPY . /home/gradle/source
COPY ./cacerts /opt/java/openjdk/lib/security/cacerts
WORKDIR /home/gradle/source
CMD ./reproducibility_test.sh
```

reproducibility_test.sh
```sh
#!/bin/bash

if [ "$PLATFORM" = "docker" ]; then
    TARGET=persist
    echo running in docker
else
    TARGET=output
    echo running natively
fi


rm -rf log_*.txt ${TARGET}/*

for s in 1 # 2 # 3 4
do
  echo running $s

  ./run_synthea -s 1668660039331 -cs 1668660039331 -r 20230101 -e 2023011${s} -p 83 Florida > log_${s}.txt

  # ./run_synthea -s 1 -cs 1 -r 20230101 -e 2023011${s} -p 1 Florida > log_${s}.txt

  mv output/fhir ${TARGET}/fhir_${s}
  mv log_${s}.txt ${TARGET}/
done

cd ${TARGET}

for file in `ls fhir_1`; do echo $file; diff fhir_1/$file fhir_2/$file ; done 
```

Then run both with
```sh
docker build -t syntheadockertest . && docker run --memory=8g --mount type=bind,source="$(pwd)"/persist,target=/home/gradle/source/persist syntheadockertest  ; ./reproducibility_test.sh
```

FHIR generated by the native run will be in ./output, FHIR generated by the docker run will be in ./persist

jawalonoski (approved):
Code reviewed. Looks fine. I'm going to run some tests before merging it.

Issue #1221, opened 2022-12-14
Encounter Leak


jawalonoski:
### What happened?

There are situations in which Encounters do not currently have the correct duration or "length of stay."

When this happens, it is because the Encounter is not properly ended. It appears the encounters are not closed due to an "Encounter Leak" where the "current encounter" is replaced while another encounter is active.

How or why this happens could be the result of several issues. Further diagnosis is required. However at least one cause is that the patient is dying in the middle of the encounter -- for example, in hospice.

jawalonoski:
Some investigation into this issue...

![EncounterDiagram-Before](https://user-images.githubusercontent.com/1054765/213469500-e2a8f966-26b2-49f5-a884-bcd48609576e.png)

![Spaghetti_and_meatballs_on_plate](https://user-images.githubusercontent.com/1054765/213470228-0e7882fd-46c9-4b60-a5d1-8dae82550b6a.jpg)

jawalonoski:
This should be fixed now with PR 1245.

That PR now requires that modules are very particular about ending encounters appropriately with `EncounterEnd` states.

This is because now there is the equivalent of a global lock on encounters -- only one module may open/close an Encounter at a time. The sole exception being "Wellness Encounters" -- where any number of encounters can "share" the current wellness encounter. However, even the wellness encounters need to be properly ended with an `EncounterEnd` state within each module.

Failure to end an encounter will mean that modules can get stuck waiting for the opportunity to proceed into an encounter and potentially doing anything at all as the patient exists in limbo.

PR #1238, opened 2023-01-12
Inpatient Length of Stay

jawalonoski:
Adjusts modules to tune inpatient length of stay. See #1169

Some `outpatient` encounters were incorrectly classified as `inpatient`.

Many inpatient encounters were modified so that length of stay follows an exponential distribution.

### DIFF
--- a/src/main/resources/modules/acute_myeloid_leukemia.json
+++ b/src/main/resources/modules/acute_myeloid_leukemia.json
@@ -3,94 +3,38 @@
   "states": {
     "Initial": {
       "type": "Initial",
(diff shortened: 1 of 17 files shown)


jawalonoski:
I'm going to mark this PR done and ready-for-review.

There is more work that can be done on inpatient LOS, but the "encounter leak" issue (#1221) makes further improvement challenging -- so I have switched over to that and may come back to this issue later with a follow-up PR.

jawalonoski:
Current status

hadleynet (approved):

PR #1259, opened 2023-03-02
US Core 5, plus code fixes

dehall:
Adds support for US Core 5: http://hl7.org/fhir/us/core/STU5/index.html

Adds a new config setting: `exporter.fhir.us_core_version` , when set to something starting with "4" it will export US Core 4-compliant FHIR R4, when set to something starting with "5" it will export US Core 5 compliant FHIR R4.

For the most part this is just a few new profiles on Observation plus a change in the Condition profile. (US Core Condition has been split into two profiles, but we had already been tagging all conditions as encounter-diagnosis so I continued that here).
In terms of the Observation profiles, they are primarily driven by the observation.category, plus the code is used to select the SDOH-assessment profile. The codes used for this are the codes from the PRAPARE survey (`encounter/sdoh_hrsn.json`) plus I manually added a couple extras that we use in the homelessness, functional status assessments, and HARK screening modules. Overall the required changes to the FHIR exporter are pretty minimal.

Additionally, while I was running the validator to make sure things looked good I saw hundreds of warnings about invalid code displays, so I took this opportunity to fix a lot of them. (There are certainly more that I missed). The changes are mostly straightforward however I saw a few instances where the suggested text didn't align with the current text, so I would look a little deeper and see we were probably using the wrong code rather than the wrong display.


dehall:
For posterity, I used this little script to assist with validating the new US Core 5 stuff and that I didn't break anything w/r/t US Core 4

```sh
#!/bin/bash

SEED=12345

rm -rf output/fhir*

./run_synthea -p 25 -s $SEED -cs $SEED -r 20230217 --exporter.fhir.us_core_version=4

mv output/fhir output/fhir_uscore4

echo Running validator on 4

java -jar ~/Downloads/validator_cli.jar  -version 4.0.1 \
  -ig hl7.fhir.us.core#4.0.0 output/fhir_uscore4/ \
  -extension any -sct us -no-extensible-binding-warnings -level warnings > validator_output_4.txt



./run_synthea -p 25 -s $SEED -cs $SEED -r 20230217 --exporter.fhir.us_core_version=5

mv output/fhir output/fhir_uscore5

echo Running validator on 5

java -jar ~/Downloads/validator_cli.jar  -version 4.0.1 \
  -ig hl7.fhir.us.core#5.0.1 output/fhir_uscore5/*.json \
  -extension any -sct us -no-extensible-binding-warnings -level warnings > validator_output_5.txt


echo Done.
```

src/main/java/org/mitre/synthea/export/FhirR4.java
```diff
@@ -185,21 +186,58 @@ public class FhirR4 {
...
       Config.getAsBoolean("exporter.fhir.transaction_bundle");
+
+  /* For the most part the US Core changes from v4 -> v5 are backwards compatible,
+   * so for those cases just check USE_US_CORE_IG.
+   * If there are version differences then use USE_US_CORE_(#) to separate them out.
+   */
   protected static boolean USE_US_CORE_IG =
       Config.getAsBoolean("exporter.fhir.use_us_core_ig");
+  protected static String US_CORE_VERSION =
+      Config.get("exporter.fhir.us_core_version", "");
```
jawalonoski, line 197: Seems like you'd want to have a default value, otherwise you won't be using any version of US Core, right?

src/main/java/org/mitre/synthea/export/FhirR4.java
```diff
@@ -303,6 +341,43 @@ private static Table<String, String, String> loadMapping(String filename) {
...
+      String version = line.get("VERSION");
+
+      if (StringUtils.isBlank(version)) {
+        // blank means applies to ALL versions
+        versions.values().forEach(table -> table.put(system, code, url));
+      } else {
+        Table<String, String, String> mappingTable = versions.get(version);
+        if (mappingTable == null) {
+          throw new IllegalArgumentException("Unknown version " + version);
+        }
```
jawalonoski, line 373: Perhaps a more informative exception message would be helpful.

Maybe, "Unknown FHIR US Core version number: " or something like that.

src/main/java/org/mitre/synthea/export/FhirR4.java
```diff
@@ -2531,7 +2660,7 @@ private static void clinicalNote(Person person,
       documentReference.addContent()
           .setAttachment(reportResource.getPresentedFormFirstRep())
           .setFormat(
-            new Coding("http://ihe.net/fhir/ValueSet/IHE.FormatCode.codesystem",
+            new Coding("http://ihe.net/fhir/ihe.formatcode.fhir/CodeSystem/formatcode",
```
jawalonoski, line 2663: This doesn't look correct to me. See http://hl7.org/fhir/R4/valueset-formatcodes.html

Despite the fact that URI contains `ValueSet` it is listed as the code system.

src/main/resources/immunization_schedule.json
```diff
@@ -304,7 +304,7 @@
     "code": {
       "system": "http://hl7.org/fhir/sid/cvx",
       "code": "113",
-      "display": "Td (adult) preservative free"
+      "display": "Td (adult), 5 Lf tetanus toxoid, preservative free, adsorbed"
```
jawalonoski, line 307: TIL `adsorb` is a real word.

src/main/resources/modules/metabolic_syndrome/kidney_conditions.json
```diff
@@ -609,8 +609,8 @@
       "codes": [
         {
           "system": "LOINC",
-          "code": "24321-2",
-          "display": "Basic Metabolic 2000 Panel"
+          "code": "51990-0",
+          "display": "Basic metabolic panel - Blood"
```
jawalonoski, line 613: I think this is supposed to be `24321-2`, because it allows for the optional inclusion of `Glomerular filtration rate`.

The `51990-0` panel does not.

jawalonoski (commented):
Thank you for updating all of these display values, but it is really a hopeless and never-ending battle... as each version of the terminology updates the descriptions change. Even some of the descriptions between SNOMED US and SNOMED International are different.

src/main/resources/modules/metabolic_syndrome/kidney_conditions.json
dehall: Ahh I didn't notice that difference. The issue here is if you look at the codes for the individual observations they are all "X in Blood" vs "X in Serum/Plasma". For example right below this here is `"2339-0" Glucose [Mass/volume] in Blood` but then in the gallstones module you'll see `"2345-7" Glucose [Mass/volume] in Serum or Plasma`. So I can change this back to the 24321-2 panel but then I think the individual component codes all have to be changed to the "serum" versions as well.

src/main/java/org/mitre/synthea/export/FhirR4.java
dehall: This is one of the changes I made just because of the validator. With the original I was getting this warning:
`Unable to check whether code is in value set http://hl7.org/fhir/ValueSet/formatcodes: [Code System URI 'http://ihe.net/fhir/ValueSet/IHE.FormatCode.codesystem' is unknown so the code cannot be validated]`

Then if you go to http://ihe.net/fhir/ValueSet/IHE.FormatCode.codesystem in a browser it redirects you to https://profiles.ihe.net/fhir/ihe.formatcode.fhir/1.1.0/CodeSystem-formatcode.html where it says `Official URL: http://ihe.net/fhir/ihe.formatcode.fhir/CodeSystem/formatcode` and so using that gets rid of the validator warnings

jawalonoski: OK, I see.

src/main/resources/modules/metabolic_syndrome/kidney_conditions.json
jawalonoski: As long as the UOM are correct, then we should probably do that.

dehall: Ok - updated all 5 panels in this file to be `24321-2` and changed the codes to be the corresponding "in serum" observations. All the units matched though the one for that Glomerular filtration rate did appear to have the wrong unit. We use that code in a bunch of other modules and they all had unit = `mL/min/{1.73_m2}` so I switched it to that

src/main/java/org/mitre/synthea/export/FhirR4.java
dehall: Updated to include more detail including the filename and what versions are expected

src/main/java/org/mitre/synthea/export/FhirR4.java
dehall: Added 5.0.1 as the default value, so it's the same as the default value in the properties file

jawalonoski:
Was about to generate and examine some data locally when I encountered this when running the unit tests:

```
org.mitre.synthea.export.FHIRR4ExporterTest > testFHIRR4Export FAILED
    java.lang.AssertionError: Validation of exported FHIR bundle failed: Unknown Code System 'http://hl7.org/fhir/us/core/CodeSystem/us-core-observation-category'|Unknown Code System 'http://hl7.org/fhir/us/core/CodeSystem/us-core-observation-category'|Unknown Code System 'http://hl7.org/fhir/us/core/CodeSystem/us-core-observation-category'|Unknown Code System 'http://hl7.org/fhir/us/core/CodeSystem/us-core-observation-category'|Unknown Code System 'http://hl7.org/fhir/us/core/CodeSystem/us-core-observation-category'
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.mitre.synthea.export.FHIRR4ExporterTest.testFHIRR4Export(FHIRR4ExporterTest.java:191)
```

dehall:
Hmm ok I'm not getting that error locally or on CI but I'll keep poking

dehall:
So unfortunately the "right" way to address that error isn't going to work. Loading all of US Core into the validator is going to be a pain, and there's a bug that seems like we can't add just the one CodeSystem. I've added a workaround in that test for this specific CodeSystem, and included as a comment some notes on why and what we should eventually do.

PR #1261, opened 2023-03-06
UTI Module Update

eedrummer:
This PR provides an update to the urinary tract infections module based on the module guide.
It also contains some small fixes for TypeOfCare transitions that were uncovered by this module being the first to use them.


src/main/resources/modules/uti/abx_tx.json
```diff
@@ -0,0 +1,1080 @@
...
+          "value_set": ""
+        }
+      ],
+      "reason": "uti",
+      "name": "Sulfa 10 day",
+      "prescription": {
+        "dosage": {
+          "amount": 1,
+          "frequency": 1,
+          "period": 2,
```
dehall, line 410: I think the `frequency` and `period` are flipped for the sulfa drugs in this module. Looks like everything is "twice daily" and all of the other states show frequency:2 period:1, but the 3 sulfa drugs show frequency:1 period:2
```suggestion
          "frequency": 2,
          "period": 1,
```

src/main/resources/modules/uti/abx_tx.json
```diff
@@ -0,0 +1,1080 @@
...
+          "value_set": ""
+        }
+      ],
+      "reason": "uti",
+      "name": "Sulfa 3 day",
+      "prescription": {
+        "dosage": {
+          "amount": 1,
+          "frequency": 1,
+          "period": 2,
```
dehall, line 356:
```suggestion
          "frequency": 2,
          "period": 1,
```

src/main/resources/modules/uti/abx_tx.json
```diff
@@ -0,0 +1,1080 @@
...
+          "value_set": ""
+        }
+      ],
+      "reason": "uti",
+      "name": "Sulfa 7 day",
+      "prescription": {
+        "dosage": {
+          "amount": 1,
+          "frequency": 1,
+          "period": 2,
```
dehall, line 383:
```suggestion
          "frequency": 2,
          "period": 1,
```

src/main/resources/modules/uti/ed_path.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "UTI_ED",
```
dehall, line 2: Super minor but I was confused for a second when I came back to this module since the name in the UI didn't match the filename and I wasn't sure which "UTI ED" module I was looking at. (I do think prefixing the submodule with the directory name is nice for clarity too even though technically it also doesn't match the filename)
```suggestion
  "name": "UTI_ED_Path",
```

src/main/resources/modules/uti/abx_tx.json
```diff
@@ -0,0 +1,1080 @@
+{
+  "name": "Abx_Tx",
+  "remarks": [
+    "A blank module"
+  ],
```
dehall, line 5: Should probably just delete these rather than leave the default "A blank module"

src/main/resources/modules/uti/telemed_path.json
```diff
@@ -0,0 +1,535 @@
+{
+  "name": "UTI_Telemed",
```
dehall, line 2: Similar note as above
```suggestion
  "name": "UTI_Telemed_Path",
```

src/main/resources/modules/uti/telemed_path.json
```diff
@@ -0,0 +1,535 @@
...
+        },
+        {
+          "distributions": [
+            {
+              "transition": "Emergency Follow Up",
+              "distribution": 0.1
+            },
+            {
+              "transition": "Ambulatory Encounter",
+              "distribution": 1
```
dehall, line 447:
```suggestion
              "distribution": 0.9
```

Technically this doesn't matter since the engine logic will correct this internally, but it's nice to have the right number show

dehall (approved):
Looks good although I have a few very minor suggestions. Feel free to merge immediately or make those few changes

jawalonoski:
Reviewing and running some tests...

PR #1271, opened 2023-03-17
Fixes for employment condition and stress level

eedrummer:
Previously, the SDoH HSRN submodule would change employment condition every wellness visit, even if the actual condition did not change.

This change keeps the same condition across wellness visits of the status does not change.

There was a similar problem with stress level. That is also fixed by this change.


eedrummer:
All of this code, plus more, is in #1277. This will likely be closed in favor of that PR.

eedrummer:
Closing. This was included in #1277

PR #1286, opened 2023-04-27
Fix CI test failures due to US Core 5 artifacts

dehall:
This PR fixes the failing CI tests, which I believe are because certain US Core 5 ValueSets don't contain relevant codes directly, they reference other valuesets. (Why the tests used to work and suddenly stopped one day a few weeks ago... I don't have an answer for that)

This does a few things:
 - Brings in the referenced value sets for "ONC Administrative Sex", "Race", "Ethnicity", and "Other or unknown or refused to answer". The HAPI validator is smart enough to be able to find these when trying to validate whether codes belong to, for example, the US Core birthsex valueset which references the ONC one.
 - Replaces the official US Core DocumentReference Type compose (which points to LOINC with type DOC) with the two specific LOINC codes we actually use. This was apparently what we did for R4 already: see https://github.com/synthetichealth/synthea/commit/78765353f732030532940aba8cbbe8289511f005#diff-0bdfbfbdef306b8401497948f30c06a573b80ab7ec944e0d6633dea92e9ec91c
 - Adds methods to ValidationResources so that we can either get a STU3 or R4 version. Since the objects here aren't static, there's no reason to load the STU3 artifacts for the R4 test, or vice versa.
 - Adds parameters to the ValidationSupportR4 to select whether US Core 4 or 5 (or both or neither) should be loaded. I can change this into an enum if people would prefer
 - Prevents toggling static fields back & forth between values in the FHIRR4ExporterTest as that seemed to be creating weird combinations in testing. Now there are 3 sub-tests with specific combinations of parameters.
 - Adds support to the ParallelTestingService to request fewer than 10 people. (so that the above test remains consistent on time, instead of running 3x as long. I think this is one of the longest tests already)

FYI, there is another unrelated uscore 5 bug (1283) which I will address in this PR if I can fix it before this is merged.

### DIFF
--- a/src/test/java/org/mitre/synthea/ParallelTestingService.java
+++ b/src/test/java/org/mitre/synthea/ParallelTestingService.java
@@ -12,20 +12,20 @@ import java.util.concurrent.TimeUnit;
 import org.mitre.synthea.world.agents.Person;
 
 public class ParallelTestingService {
- /**
-  * Runs the provided PersonTester in parallel. The PersonTester will be supplied with 10 people.
-  * Tests are run in a fixed thread pool with some of the tests having to wait until the first few
-  * tests complete. Tests can return a List of validation errors (in String form).
-  * @param pt An implementation of PersonTester
-  * @return A list of errors
-  * @throws Exception when bad things happen during the test
-  */
+  /**
+   * Runs the provided PersonTester in parallel. The PersonTester will be supplied with 10 people.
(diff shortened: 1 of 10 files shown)


jawalonoski (approved):
These changes look good. Any other pending changes?

dehall:
Nothing else pending. I think this should cover all the test failures we've seen lately

PR #1291, opened 2023-05-10
Claims CSV contains invalid uuids

SimonUzL:
Hi,
claims, exported as CSV, contains both, according to the wiki, a Primary Patient Insurance ID and a Secondary Patient Insurance ID of type UUID. However, the code produces 0, if no Payer exists. 0 is not a valid UUID and hence we are unable to import the exported CSV files into a PostgreSQL database.
In case of no Payer exists, I suggest to generate nothing (NULL in SQL speak) instead of 0.


jawalonoski (approved):

jawalonoski:

Issue #1297, opened 2023-05-23
Make pretty-printing of exported data configurable


dehall:
### Requested Feature

Synthea should have a configuration option to disable pretty-printing - this could reduce file size and improve performance

PR #1307, opened 2023-06-05
Configuration option to disable pretty printing for XML and JSON

jawalonoski:
The following exporters will now use the `exporter.pretty_print` configuration setting:

- FHIR (all versions)
  - Pretty printing settings are ignored when using `exporter.fhir.bulk_data` 
- CCDA
- JSON

The savings in C-CDA file size is low, only around 2% smaller.
The savings in FHIR file size can be quite large, approximately 33% smaller in most cases.


jawalonoski:
Addresses #1297

hadleynet (approved):

PR #1312, opened 2023-06-09
Symptoms

jawalonoski:
- Add symptom unit tests
- Possible fix for intermittent C-CDA export issue with "super Encounter" and coverage at time X.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/CCDAExporter.java
+++ b/src/main/java/org/mitre/synthea/export/CCDAExporter.java
@@ -75,6 +75,7 @@ public class CCDAExporter {
       // causes an exception, then we fake an insurance
       // plan for the purposes of creating the super encounter.
       person.coverage.setPlanToNoInsurance(time);
+      person.coverage.setPlanToNoInsurance(Long.MAX_VALUE);
     }
     // create a super encounter... this makes it easier to access
     // all the Allergies (for example) in the export templates,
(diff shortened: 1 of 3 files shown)


src/main/java/org/mitre/synthea/export/CCDAExporter.java
```diff
@@ -75,6 +75,7 @@ public static String export(Person person, long time) {
       // causes an exception, then we fake an insurance
       // plan for the purposes of creating the super encounter.
       person.coverage.setPlanToNoInsurance(time);
+      person.coverage.setPlanToNoInsurance(Long.MAX_VALUE);
```
jawalonoski, line 78: Unrelated bug fix to intermittent C-CDA unit test failure.

src/test/java/org/mitre/synthea/engine/ExpressedSymptomTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Test
+  public void testSymptomInfo() {
+    String name = "pain";
+    ExpressedSymptom symptom = new ExpressedSymptom(name);
+    String cause = "test";
+    Integer value = 100;
+    Long time = 0L;
+
+    SymptomInfo info = symptom.new SymptomInfo(cause, value, time);
```
eedrummer, line 22: I don't ever remember coming across this syntax before. I had to look it up. I guess it is because we always use `static` inner classes elsewhere in the project.

PR #1346, opened 2023-08-03
Code comments to resolve Javadoc errors

awatson1978:
This PR is related to discussion item https://github.com/synthetichealth/synthea/discussions/1345.  Need to evaluate if this level of code commenting is appropriate, or whether we need to relax linting rules.


dehall:
My vote is that comments like these where it's just a rehash of the code (or in this case, the code itself verbatim) are not useful, and in fact have negative value because now you have neither a meaningful comment nor a warning that says "you need to have comments".
Ideally we'd have real comments describing the what and/or why of everything but I realize that's a nontrivial effort. Maybe there's a flag we can use to disable that linter on a given field such as these BB enums? Loosening the restrictions on comments globally sounds like admitting defeat

awatson1978:
Replaced by https://github.com/synthetichealth/synthea/pull/1582

PR #1351, opened 2023-08-11
WIP: Add support for external X12 exporter

hadleynet:


dehall:
I *just* deleted the branch because I thought we were never going to use it, but would this be an opportunity to use a more generic approach as seen in https://github.com/synthetichealth/synthea/pull/1350 ? We can restore that branch with one click if interested

dehall:
Closing this PR, using the approach from #1350 instead

Issue #1372, opened 2023-10-14
Vital signs must have finite values - NaN is invalid


arvindshmicrosoft:
### What happened?

I got this exception `java.lang.IllegalArgumentException: Vital signs must have finite values - NaN is invalid` multiple times with a build from commit 6192ed703d8517e7bfab801d9fbee622ec4e8161 (latest as of 2023-10-07). Symptomatically it's identical to #1038. 

A sample command line was:

`java -Xmx24G -jar ./build/libs/synthea-with-dependencies.jar --exporter.fhir.export false --exporter.csv.export true --exporter.baseDirectory '/synthea_data/Arizona Phoenix 706' --generate.log_patients.detail none --exporter.clinical_note.export false -p 781512 Arizona Phoenix`

Relevant output from the process:
```
Running with options:
Population: 781512
Seed: 1696672364458
Provider Seed:1696672364458
Reference Time: 1696672364458
Location: Phoenix, Arizona
Min Age: 0
Max Age: 140
```

As I mentioned, the exception is observed in many runs, for different cities / areas. For example, I ran all the cities in California, the setVitalSign / NaN exception is seen in the following cities. How impactful is this exception? Is the entire generated dataset suspect, or will the impact be localized only to specific individual?

```
California Anaheim (2 occurrences)
California Arden-Arcade
California Bakersfield
California Blythe
California Concord
California Davis
California El Monte
California Fresno
California Inglewood
California Irvine
California Menlo Park
California Moreno Valley
California Orinda
California Oxnard
California San Diego (2 occurrences)
California San Jose
California Santa Ana
California Santa Monica
California Soledad
California Stanton
California Vallejo
California Westmont
California Windsor
```

### Environment

OS:
```
Linux version 4.18.0-425.10.1.el8_7.x86_64 (mockbuild@dal1-prod-builder001.bld.equ.rockylinux.org) (gcc version 8.5.0 20210514 (Red Hat 8.5.0-15) (GCC)) #1 SMP Thu Jan 12 16:32:13 UTC 2023
```

Java:
```
openjdk 11.0.20 2023-07-18 LTS
OpenJDK Runtime Environment (Red_Hat-11.0.20.0.8-1) (build 11.0.20+8-LTS)
OpenJDK 64-Bit Server VM (Red_Hat-11.0.20.0.8-1) (build 11.0.20+8-LTS, mixed mode, sharing)
```

### Relevant log output

```shell
[...]

PR #1381, opened 2023-10-23
Fix issue where vital signs were sometimes being set to `NaN`

eedrummer:
Synthea's simulation of height and weight under the age of 20 depends on growth charts published by CDC. The growth charts are [provided with a method](https://www.cdc.gov/growthcharts/percentile_data_files.htm) for transforming values into percentiles and vice-versa. It appears that for higher percentiles, this method can break down. 

The case that was used for testing this issue is transforming the ~99.74th percentile BMI for females at 240 months into an actual BMI value. 

This code change addresses the issue by converting any request above the 99.5th percentile to just the 99.5th percentile.

This addresses #1372.

### DIFF
--- a/src/main/java/org/mitre/synthea/world/concepts/GrowthChartEntry.java
+++ b/src/main/java/org/mitre/synthea/world/concepts/GrowthChartEntry.java
@@ -6,6 +6,16 @@ import java.io.Serializable;
  * Represents the LMS parameters on the CDC growth chart for a particular sex and age in months.
  */
 public class GrowthChartEntry implements Serializable {
+
+  /**
+   * The highest percentile to use when attempting to calculate a value. Tests with the BMI growth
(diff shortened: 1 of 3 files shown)


dehall (commented):
Code change looks good once the function actually uses the new constant. 
Do you have a sense of whether this really only affects the ~1/200 records that are > 99.5 percentile? Or in practice is it less, or more? I'm having a little trouble following the BMI <-> percentile logic as the flow seems to convert one to the other often, meanwhile the height and weight are both also changing.

dehall (approved):
I ran 200 patients against master and this branch and saw only 31 differences in the observations.csv (out of a total of 137,000 lines). Most were actually "Head Occipital-frontal circumference" but maybe that's just how the dice rolled for that population. Overall looks good

Issue #1380, opened 2023-10-23
NullPointerException seen in various executions


arvindshmicrosoft:
### What happened?
I'm seeing various instances of `NullPointerException` in executions for different cities (1152 out of the 43944 cities in demographics.csv), each with a stack like one of the below examples. The code base I've built the JAR from corresponds to commit https://github.com/synthetichealth/synthea/commit/6192ed703d8517e7bfab801d9fbee622ec4e8161 (latest as of 2023-10-07).

## Pattern 1
Here is one call stack pattern:
```
java.lang.NullPointerException: Cannot invoke "org.mitre.synthea.engine.State.clone()" because the return value of "java.util.Map.get(Object)" is null
	at org.mitre.synthea.engine.Module.process(Module.java:415)
	at org.mitre.synthea.engine.Module.process(Module.java:368)
	at org.mitre.synthea.engine.State$CallSubmodule.process(State.java:270)
	at org.mitre.synthea.engine.State.run(State.java:193)
	at org.mitre.synthea.engine.Module.process(Module.java:411)
	at org.mitre.synthea.engine.Module.process(Module.java:368)
	at org.mitre.synthea.engine.State$CallSubmodule.process(State.java:270)
	at org.mitre.synthea.engine.State.run(State.java:193)
	at org.mitre.synthea.engine.Module.process(Module.java:411)
	at org.mitre.synthea.engine.Module.process(Module.java:368)
	at org.mitre.synthea.engine.State$CallSubmodule.process(State.java:270)
	at org.mitre.synthea.engine.State.run(State.java:193)
	at org.mitre.synthea.engine.Module.process(Module.java:411)
	at org.mitre.synthea.engine.Module.process(Module.java:368)
	at org.mitre.synthea.engine.Generator.updatePerson(Generator.java:712)
	at org.mitre.synthea.engine.Generator.createPerson(Generator.java:668)
	at org.mitre.synthea.engine.Generator.generatePerson(Generator.java:478)
	at org.mitre.synthea.engine.Generator.lambda$run$3(Generator.java:383)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```
[...]

PR #1384, opened 2023-10-26
Proposal for fixing issue with NullPointerException in LifecycleModule.

eeevgeniev:
Proposal for fixing bug NullPointerException in LifecycleModule. 

The bug is produced with this parameters: -p 10 Alaska

The bug is described in this issue: https://github.com/synthetichealth/synthea/issues/1380, pattern 2.

### DIFF
--- a/src/main/java/org/mitre/synthea/modules/LifecycleModule.java
+++ b/src/main/java/org/mitre/synthea/modules/LifecycleModule.java
@@ -123,7 +123,10 @@ public final class LifecycleModule extends Module {
     calculateFallRisk(person, time);
     person.attributes.put(Person.DISABLED, isDisabled(person, time));
     if (person.ageInYears(time) >= 18) {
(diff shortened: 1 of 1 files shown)


eedrummer:
@eeevgeniev I appreciate the proposed fix, but this doesn't address the root cause of the issue. 

This PR will ensure that the NPE isn't thrown, but the key issue is that the employment model isn't there. Synthea assumes that every person should have an employment model. This should be set in [Location.setSocialDeterminants()](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/world/geography/Location.java#L390). If there is an issue figuring out what the employment model should be, a different exception should be thrown.

Something is going on either in setting up the employment model or it is being removed from the persons attributes. If that part isn't fixed, the simulation of employment and job loss will be silently stopped (or never properly started).

eedrummer:
See #1386 for a different proposed fix.

PR #1385, opened 2023-10-26
Fix export filter keeping old entries when there is an active entry of the same type

### LINKED ISSUE #1383: Export filter keeps all instances of a condition type when an active instance exists
### What happened?

The export filter is incorrectly keeping historical instances of Condition when there is an active instance of the same type.

Relevant code: 
https://github.com/synthetichealth/synthea/blob/6192ed703d8517e7bfab801d9fbee622ec4e8161/src/main/java/org/mitre/synthea/export/Exporter.java#L605

Consider for example a patient that has multiple instances of Viral Sinusitis, or in the example I found, "Medication Review Due". If they currently have an active instance of the condition, this predicate will return `true` for **every** instance of that condition across their record, which isn't what we want. I believe we only need to look at start & stop date but this requires more review.

Sample patient attached: [Ciera654_Boyle917_45aa9ffe-bb1e-c459-9f42-1944832cb8d1.json](https://github.com/synthetichealth/synthea/files/13119216/Ciera654_Boyle917_45aa9ffe-bb1e-c459-9f42-1944832cb8d1.json)

### Environment

```markdown
- OS: macOS 13.6
- Java: openjdk version "17.0.8" 2023-07-18
- Run command: `./run_synthea -s 1 -cs 1 -r 20231024 -ps -4964420948893066024`
```

### Relevant log output

_No response_

dehall:
Fixes #1383 

This PR addresses an issue where the export filter isn't filtering out old entries when there is a separate active entry of the same "type", ie, code. This is especially likely to happen with "Medication Review Due" from the Med Rec module, but it could also happen for any potentially-recurring entry such as Viral Sinusitis or chronic medications.

The issue is that the logic is using `record.conditionActive` which only checks whether the given code is in the "active" set, it doesn't look at the specific entry at all. The change here is to only look at dates, and the date logic can be pretty simple: "if the entry is open (stop = 0) or the stop is after the cutoff, keep the entry"

### DIFF
--- a/src/main/java/org/mitre/synthea/export/Exporter.java
+++ b/src/main/java/org/mitre/synthea/export/Exporter.java
@@ -598,25 +598,15 @@ public abstract class Exporter {
 
     long cutoffDate = endTime - Utilities.convertTime("years", yearsToKeep);
     Predicate<HealthRecord.Entry> notFutureDated = e -> e.start <= endTime;
+    Predicate<HealthRecord.Entry> entryIsActive = e -> e.stop == 0L || e.stop > cutoffDate;
 
     for (Encounter encounter : record.encounters) {
       List<Claim.ClaimEntry> claimItems = encounter.claim.items;
-      // keep conditions if still active, regardless of start date
-      Predicate<HealthRecord.Entry> conditionActive = c -> record.conditionActive(c.type);
-      // or if the condition was active at any point since the cutoff date
-      Predicate<HealthRecord.Entry> activeWithinCutoff = c -> c.stop != 0L && c.stop > cutoffDate;
-      Predicate<HealthRecord.Entry> keepCondition = conditionActive.or(activeWithinCutoff);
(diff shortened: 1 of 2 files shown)


jawalonoski (commented):
I'm not sure whether or not to approve this PR. On the one hand, it _massively_ reduces the amount of bloat in the record due to repetition.... mostly related to entries like `Medication review due (situation)` and employment status...

But it also removes more valuable entries like `Normal Pregnancy` -- which is double-edged sword. We lose the ability to track how inaccurate our pregnancy model is (the reproduction rate is too high), but instead of fixing it in the model it just "looks better" because they get filtered in the output.

dehall:
Hmm that's an interesting point. The thing is though we're already filtering those entries out today if there's not an active instance of the same type. This PR is just making the logic more consistent with how we intended it. 
For the pregnancy stats, I would say we shouldn't be looking at filtered records anyway when we're reviewing data.

jawalonoski (approved):

PR #1399, opened 2023-11-28
Module cleanup

dehall:
This PR cleans up a few items that pop up pretty often in generated records and look odd:

1. There are several ConditionOnsets that do not have a corresponding ConditionEnd and are left open forever, even though they are acute conditions which are addressed in the module:
   - Appendicitis & rupture of appendix
   - Opioid addiction: Drug overdose (note there are a few other conditions in this module which don't get ended, like "Impacted molars", "Chronic pain", and migraine, which are not really treated so I'm ok leaving them open)
   - Miscarriage (added a "history of miscarriage" condition)
   - Covid19 outcomes: Hypoxemia (seems like this was just missed in the long list of conditions, I don't see anything in the old notes for why this one specifically should stay open)
2. Prediabetes can be diagnosed after diabetes, and based on medication impacts it happens often, so that's no longer possible
3. A couple modules include "patient transfer" as a condition where it really doesn't fit. There's already a Procedure for the transfer so I just removed the Condition.

### DIFF
--- a/src/main/resources/modules/appendicitis.json
+++ b/src/main/resources/modules/appendicitis.json
@@ -309,8 +309,7 @@
         "Avg operative time is ~55 minutes",
         "https://www.ncbi.nlm.nih.gov/pubmed/17658102"
       ],
-      "direct_transition": "Appendicitis_Symptom1_Ends",
-      "name": "Appendectomy"
+      "direct_transition": "End_Appendicitis"
     },
     "Appendicitis_Symptom1_Ends": {
       "type": "Symptom",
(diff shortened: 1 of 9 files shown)


jawalonoski:
> I don't really want to get rid of the reasons on anything, but "Transition from acute care to home-health care (finding)" and "Patient transfer to skilled nursing facility (procedure)" really don't feel like they should be Conditions, and we don't have anything else that can be set as a reason. Ideally whichever module put the patient into SNF/home health would also set a specific attribute they set which these could use as a reason, but I don't think that's possible in the module builder today. If there's no better option I can revert those two modules

`SetAttribute` can be set as a `Code`, so we could try that. It is hardwired in the module builder to be LOINC, but if you manually edit the JSON to be SNOMED the engine will respect that. Maybe that will suffice, especially if we only set the attribute if it is nil.

Then later, we could modify the modules put the patient into SNF/home care to set that attribute.... which would be a better long term solution, and setting the attribute only-if-nil would be a backup.

dehall:
Just pushed some updates to bring back the `reason` in the SNF and home health modules. The ConditionOnset that was originally there has been replaced by a SetAttribute, called only if the attribute isn't already nil, and the attribute is cleared out at the end of the module.

Home health was only invoked by 2 other modules, Total Joint Replacement and Home Hospice. TJR was easy to add the appropriate codes to for the "reason", but in the Home Hospice module, there are ~40 different possible conditions the patient might have so I skipped that. For SNF, there are a bunch of instances in the Injuries module and that same group of 40 conditions from the Home Hospice module, so I didn't add any more specific reasons. If you think it's worthwhile I can add the codes for the respective Injuries module branches that call the SNF submodule.

PR #1403, opened 2023-12-12
Templates and Keep modules

dehall:
This PR supports two "new" categories of modules:

## Templates
Templates are snippets of modules intended to help authors create a new module quickly. A corresponding change to the module builder (https://github.com/synthetichealth/module-builder/pull/329) adds the ability to load these templates. After submitting this PR I will put together some documentation on how to use these to create a new module.

## Keep Modules
Keep modules have been supported for a while; this PR adds a set of built-in keep modules and updates the `-k` flag to support either arbitrary paths or these built-in ones. We had two already `must_have_cabg.json` and `must_have_cardiac_surgery.json` which I moved from the repo root to this new directory so any existing scripts that reference them can still work without changes. This should work both running via gradle/run_synthea or as a JAR file.

Tech Notes:
 - I migrated `Module.fixPathFromJar` into the Utilities class since I was using it in App as well now
 - Changed `Generator.keepPatientsModulePath` from File to Path since you can't reference files in a zip/jar as a File


src/main/java/org/mitre/synthea/helpers/Utilities.java
```diff
@@ -638,4 +642,29 @@ public static Map cleanMap(Map input) {
...
+
+  /**
+   * Enable reading the given URI from within a JAR file.
+   * For example, for command-line args which may refer to internal or external paths.
+   * Note that it's not always possible to know when a user-provided path
+   * is within a JAR file, so this method should be called if it is possible the
+   * path refers to an internal location.
+   * @param uri URI to be accessed
+   */
+  public static void fixPathFromJar(URI uri) throws IOException {
```
eedrummer, line 654: I know that you're just moving this method, but I don't think the name of it is particularly clear.

It's not doing anything with the Path. It's just letting the JVM know about another place it can look for files. Maybe something like `updateReadableLocations`?

dehall: Updated to `enableReadingURIFromJar` and rebased to fix the PulseOx test failure

PR #1438, opened 2024-03-29
Module Metrics

jawalonoski:
This branch revamps the module, state, and transition metrics.

Previously, use of the `generate.track_detailed_transition_metrics` property outputted metrics to system out. Using this feature roughly quadrupled the runtime.

This change outputs the metrics as JSON (`./output/metrics/**`), one file per module. Using this feature will roughly double the runtime.

Sample output looks something like this for each state, where the name of the state (e.g., `Chance_of_Stroke` matches a state name in the corresponding GMF module):

```json
  "Chance_of_Stroke": {
    "entered": 58987,
    "duration": 152635104000000,
    "population": 101,
    "current": 100,
    "destinations": {
      "Stroke": 1,
      "Chance_of_Stroke": 58886
    }
  }
```

Field | Description
-----|-----
`entered` | Count, how many times this state was entered.
`duration` | Sum of time (milliseconds) spent in this state.
`population` | Count, how many unique individuals entered this state (this count is not reliable for submodules).
`current` | Count, how many individuals are currently in this state at simulation end.
`destinations` | Map of states and counts that processing transitioned to.

### DIFF
--- a/src/main/java/org/mitre/synthea/helpers/TransitionMetrics.java
+++ b/src/main/java/org/mitre/synthea/helpers/TransitionMetrics.java
@@ -3,20 +3,21 @@ package org.mitre.synthea.helpers;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Table;
 import com.google.common.collect.Tables;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
-import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.mitre.synthea.engine.Module;
-import org.mitre.synthea.engine.State;
-import org.mitre.synthea.world.agents.Person;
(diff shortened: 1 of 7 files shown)


src/main/java/org/mitre/synthea/helpers/TransitionMetrics.java
```diff
@@ -105,138 +87,56 @@ public Metric getMetric(String moduleName, String stateName) {
...
 
-        if (!stats.destinations.isEmpty()) {
-          System.out.println(" Transitioned to:");
-          long total = stats.destinations.values().stream().mapToLong(ai -> ai.longValue()).sum();
-          stats.destinations.forEach((toState, count) ->
-                System.out.println(" --> " + toState + " : " + count + " = "
-                                    + decimal(count.get(), total) + "%"));
-        }
-        System.out.println();
+    List<ModuleSupplier> suppliers = Module.getModuleSuppliers();
```
hadleynet, line 114: Just a thought but if you modified `List<ModuleSupplier> getModuleSuppliers(Predicate<String> predicate)` in Module.java to take a Predicate<ModuleSupplier> instead then you could get rid of the loop that follows while also making that method more flexible.

hadleynet (commented):
Tested and verified locally, just one suggestion that you can ignore if you'd prefer to keep things as is.

src/main/java/org/mitre/synthea/helpers/TransitionMetrics.java
jawalonoski: That sounds like a good idea. I'll take a look at it.

jawalonoski: OK, I made a change. Everything should be ready to go.

PR #1450, opened 2024-04-25
Add support for US Core 3.1.1 and 6.1.0

dehall:
Adds support for US Core 3.1.1 and 6.1.0 and makes 6.1.0 the new default. (My understanding is that since 5 is no longer relevant for certification in the US, nobody is likely to use it much but it's still there anyway) The code is starting to be kind of a mess so I started thinking about breaking all the US Core changes into a separate file, but to keep the diff here minimal I ultimately kept things all in the same file. Happy to refactor & break things out if people want that.

This chart was helpful in describing when things were added/removed from US Core: https://hl7.org/fhir/us/core/STU6.1/changes-between-versions.html
Note that just like for 4 & 5, the included documentreference-type VS for 3 and 6 was pre-loaded with the codes we actually use. (The official version uses a compose which apparently our internal ValidationSupport can't handle)

The one module change is in response to a validator error that required that "magic" code to be present.

In addition to the unit tests I used this little script to call the CLI validator to test out the different versions, see output files validate_results_uscore_${v}.txt for results:

```sh
for v in 3.1.1 4.0.0 5.0.1 6.1.0; do
  rm -rf output/
  echo Running $v
  ./run_synthea -p 10 --exporter.fhir.us_core_version=$v
  java -jar /Users/dehall/Downloads/validator_cli_6_3_4.jar -ig "hl7.fhir.us.core#${v}" -display-issues-are-warnings -disable-default-resource-fetcher -best-practice ignore -level errors -extension any output/fhir/*.json | tee validate_results_uscore_${v}.txt
done
```

There will still be some errors using the CLI validator like "Unknown code 'X9999-0'" (for our custom codes) and "Error from http://tx.fhir.org/r4: Unknown filter "tty = SCDG"" (Grahame says this is a VSAC problem) but there shouldn't be anything that suggests a profile error.


jawalonoski:
Cascading error with 4.0.0:

`Error @ Bundle.entry[51].resource/*MedicationRequest/afd04554-8569-7b5f-29d9-17cb446a3b9a*/.medication.ofType(Reference) (line 4227, col8): Unable to find a match for profile urn:uuid:609cc372-6d0b-35f8-de3c-30171ef10a44 among choices: http://hl7.org/fhir/us/core/StructureDefinition/us-core-medication
`

Looks like a false positive, since the reference is indeed a `us-core-medication`, but which failed with the `Unknown filter "tty = SCDG"` message.

jawalonoski (approved):

PR #1457, opened 2024-05-10
Module syntax

jawalonoski:
This pull request targets module syntax cleanliness.

Some of these changes are not functional, for example, removing unused extra fields from the JSON.

Some of them are errors that were luckily avoided during runtime, due to code preferring one parameter to another, or relying on defaults or not.

Many of these syntax issues are artifacts of the module builder. For example, copying & pasting modules rather than "downloading" them results in extra `name` fields. Inserting a state into a model can shift a transition, leaving a state with two transitions (but only rendering the last one), and so forth.

### DIFF
--- a/src/main/resources/modules/allergies/food_allergy_incidence.json
+++ b/src/main/resources/modules/allergies/food_allergy_incidence.json
@@ -524,7 +524,7 @@
           "display": "Wheat (substance)"
         }
       ],
-      "reaction": [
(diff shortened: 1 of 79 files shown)


hadleynet (approved):
Ran tests, tested rendering via graphvis, manually reviewed diffs and loaded modules into module builder for additional review where I didn't immediately understand the change..

PR #1520, opened 2024-10-11
Updated snomed codes, based on the exchange discusssion with Dylan

florim14:
#### Overview

This pull request aims to update the SNOMED codes across various Synthea modules, ensuring they are current and active. The updates were made using a script specifically developed for this purpose, and from an exchange discussion with Dylan.

#### Changes Made

1. **Script Development:** A custom script was created to review and update the SNOMED codes within the Synthea modules. The script performed the following tasks:
    - Identified inactive or outdated SNOMED codes.
    - Check for active versions of these codes. If the code is still active but doesn't match with the current display, we update it (we also have a list of codes, which for now we ignore, and with be replaced in the future). On the other hand, if no code is found, we search based on the display and the semantic tag (if it has one), and check if it matches with at least 70% of similarity.
    - Updated the modules with the active SNOMED codes.

2. **Code Updates:**
    - The script successfully updated all SNOMED codes across the modules, except for 12 codes for which no active matches were found, and the ones we choose to ignore for now.
    - Out of these 12 unmatched codes, 11 are located in the "TNM_Diagnosis" module file.

#### What are the benefits:

- Ensures all SNOMED codes used in the Synthea modules are active, improving the integrity and reliability of the synthetic data generated.
- The developed script (once it is published) can be reused for future updates, simplifying the maintenance process for SNOMED codes in Synthea modules.

#### Remaining Issues

- **Unmatched Codes:**
    - 12 SNOMED codes could not be matched with active versions, and the ones we decided to ignore.
    - 11 of these unmatched codes are located in the "TNM_Diagnosis" module.

#### Future Work

- **Script Contribution:**  While this pull request focuses on the updated SNOMED codes, the script used for these updates will be shared in a future contribution. This script will assist other users in maintaining up-to-date SNOMED codes within their Synthea modules.

Thank you for considering this pull request.

### DIFF
--- a/src/main/resources/modules/allergies.json
+++ b/src/main/resources/modules/allergies.json
@@ -101,7 +101,7 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": 185347001,
+          "code": "185347001",
           "display": "Encounter for problem (procedure)"
         }
       ],
@@ -130,7 +130,7 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": 395142003,
+          "code": "395142003",
           "display": "Allergy screening test (procedure)"
         }
       ],
@@ -158,14 +158,14 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": 384758001,
+          "code": "384758001",
           "display": "Self-care interventions (procedure)"
         }
       ],
       "activities": [
         {
           "system": "SNOMED-CT",
-          "code": 58332002,
+          "code": "58332002",
           "display": "Allergy education (procedure)"
         }
       ],
--- a/src/main/resources/modules/allergies.json
+++ b/src/main/resources/modules/allergies.json
@@ -101,8 +101,8 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": "185347001",
-          "display": "Encounter for problem"
+          "code": 185347001,
+          "display": "Encounter for problem (procedure)"
         }
       ],
       "conditional_transition": [
@@ -130,8 +130,8 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": "395142003",
-          "display": "Allergy screening test"
+          "code": 395142003,
+          "display": "Allergy screening test (procedure)"
         }
       ],
       "direct_transition": "Allergy_Panel"
@@ -158,15 +158,15 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": "384758001",
+          "code": 384758001,
           "display": "Self-care interventions (procedure)"
         }
       ],
       "activities": [
         {
           "system": "SNOMED-CT",
-          "code": "58332002",
-          "display": "Allergy education"
+          "code": 58332002,
+          "display": "Allergy education (procedure)"
         }
       ],
       "conditional_transition": [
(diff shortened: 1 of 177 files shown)


src/main/resources/modules/allergies.json
```diff
@@ -158,15 +158,15 @@
       "codes": [
         {
           "system": "SNOMED-CT",
-          "code": "384758001",
+          "code": 384758001,
```
dehall, line 161: I didn't notice this before but there are some instances in the diff like this one, where neither the code nor display value changed, but the code was changed from a string to a number. Any idea what happened here? It doesn't seem like all codes were changed. This isn't a huge deal but if it's possible to not change the type if there wasn't a value change, or just make them all strings instead of numbers, that would be preferable to making them numbers

florim14:
@dehall  - I have noticed that as well when I was writing the script, some were numbers, and some were strings, so I thought it would be better to be consistent and when I was doing the check, changed them to numbers. I can make the adjustment and make them strings, not a problem at all. Should I do it?

dehall:
Yes, if that's a simple change please do that. Thanks!

For more detail: All codes are really strings, in the case of SNOMED it's usually safe to represent them as numbers but not always, so better to keep them as strings. (And LOINC for example includes a dash so the codes are necessarily strings.) The reason many of them are stored as numbers now is our module builder UI only has one text entry component and it tries to infer the type from the content, so a string that looks numeric turns into a number. This has been a problem before: https://github.com/synthetichealth/module-builder/issues/280

florim14:
@dehall - perfect, I see. I did the changes and converted them into strings, and committed the changes. If there is any other change needed, please let me know.

florim14:
@dehall  - thank you as well for the discussions and the help to make this PR great. I will reach to you later when you are more free and we can discuss on the way to update the rest of the codes. Thank you one more time, and wish you a great day

dehall:
Hi @florim14 , thank you! I hope you've had a good holiday season as well! 

I'm fairly busy through this month but would be happy to meet sometime in February if that works for you. I'm on US Eastern time and Mondays and Fridays are probably best for me to avoid conflicts.

One other thing that might be helpful in sorting these last few codes out would be if you could include the name of the state the code appears in. This would help to clarify what the intent of the code was, because I know there are some instances where the code we are currently using doesn't match what we intended.

For instance from one of the files you sent previously:
```
  {
    "filePath": "modules\\allergies.json",
    "code": "419199007",
    "display": "Allergy to substance (finding)"
  },
```

This comes from a state named `Allergy_Unspecified`, so when we choose a new code for this we'll want to take that idea of "unspecified" into account. Adding a state name field to the report like the following would be great to have all the info in one spot, and hopefully not a big effort.

```
  {
    "filePath": "modules\\allergies.json",
    "stateName": "Allergy_Unspecified",
    "code": "419199007",
    "display": "Allergy to substance (finding)"
  },
```

florim14:
Hi @dehall,

thanks a lot for you time. February sounds great, and Monday or Fridays, both should be fine for me. What time does work for you better?

Thank you for the clarification, yes definitely, that is helpful. I will add the stateName as well, and we can go through them.

dehall:
If you're in Europe then probably the start of my workday is around the end of yours -- I'm generally available starting at 8 AM ET (2 PM CET). Feel free to email me dehall@mitre.org to get something on the calendar

PR #1521, opened 2024-10-15
ICD-10-CM Code Mapping

hadleynet:
1. Add support for mapping condition codes to ICD-10-CM in the FHIR R4 exporter.
2. Fix some checkstyle warnings.


jawalonoski:
General comment on the FHIR exporter... it would make sense to repeat the pattern for `Encounter.reason` to the other resources that also have a `reason` field... for example `Procedure` and `MedicationRequest`. Some of that code looks like it could get stuffed into a method as well.

jawalonoski (approved):

PR #1527, opened 2024-11-08
Flexporter - Update randomCode fn to take output type as an optional second param

dehall:
Updates to address flexporter issues identified by @elsaperelli :

1. I realized the `randomCode` function was improperly named since technically it produces a Coding not a code. Rather than rename it and break backwards compatibility, this gives it an optional second parameter for "return type". This can be "code" to return a string, "Coding" to return a Coding (ie `{system, code, display}`), or "CodeableConcept" to return a CodeableConcept. Default if unspecified is Coding, for backwards compatibility.

2. Adds support for defining custom ValueSets in a flexporter mapping file, so that you can select a randomCode from your handpicked set of codes rather than having to use a terminology server and pre-existing VS. The syntax is a YAML version of the ValueSet resource, to minimize inventing new things. There are some quirks to this though so I'm open to suggestions here:
    - Our YAML parser doesn't work with the HAPI FHIR model classes, so this uses a roundabout way of loading that parses the YAML to `Map<String,Object>`, serializes that to JSON, then parses the JSON as FHIR. The alternative here would be to define a custom new model to represent a VS, or to re-implement the ValueSet class.
    - The ValueSets need to be loaded once per mapping file, so I added a new function to be called in the main flexporter entrypoints (App and RunFlexporter). Trying to load them multiple times wouldn't break anything but I don't want to do that (YAML) -> (JSON) -> (FHIR classes) sequence multiple times unnecessarily.

3. Added some extra error handling for the value set $expand operation, to assist in debugging when it returns an OperationOutcome

### DIFF
--- a/src/main/java/App.java
+++ b/src/main/java/App.java
@@ -220,6 +220,8 @@ public class App {
             if (flexporterMappingFile.exists()) {
               Mapping mapping = Mapping.parseMapping(flexporterMappingFile);
               exportOptions.addFlexporterMapping(mapping);
+              mapping.loadValueSets();
+
               // disable the graalVM warning when FlexporterJavascriptContext is instantiated
               System.getProperties().setProperty("polyglot.engine.WarnInterpreterOnly", "false");
             } else {
--- a/src/main/java/RunFlexporter.java
+++ b/src/main/java/RunFlexporter.java
@@ -1,9 +1,6 @@
 import ca.uhn.fhir.parser.DataFormatException;
 import ca.uhn.fhir.parser.IParser;
 
-import com.fasterxml.jackson.core.type.TypeReference;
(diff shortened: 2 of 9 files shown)


elsaperelli (approved):
Lgtm! Great new feature

dehall:
Update: I'm converting this to draft and will add the ability to define custom value sets within a flexporter mapping. It's a little roundabout but it will allow for choosing from just a small set of codes rather than forcing the user to use a terminology service and real VS

elsaperelli (approved):
This custom value set functionality is awesome! Tested out with our qpp qicore mapping file to split the MedicationDispense status ValueSet and it worked very easily!

jawalonoski (approved):

PR #1554, opened 2025-01-22
Update synthea.properties

VeluuPL:
exporter.csv.export = true


jawalonoski:
Thanks for the pull request, but our default output is HL7 FHIR only, and CSV is an option.

You are welcome to have CSV output on by default in your local instances, but we are not going to change the default for all users.

PR #1567, opened 2025-04-04
feat(provider_finder): Add 'prefer_one' provider finding strategy

andrequina:
Introduces a new provider finding strategy, `ProviderFinderPreferOne`, which allows configuring Synthea to prioritize a specific provider based on their NPI.

Key changes include:
- Added `ProviderFinderPreferOne.java` implementing the new logic.
- Integrated `ProviderFinderPreferOne` into `Generator.java` to:
    - Set the initial simulation state based on the preferred provider's location.
    - Allow overriding patient demographics based on the preferred provider.
- Added the `prefer_one` option to the provider finding mechanism in `Provider.java`.
- Refactored provider file path handling in `Provider.java` for clarity and to support locating the preferred provider.

### DIFF
--- a/src/main/java/org/mitre/synthea/engine/Generator.java
+++ b/src/main/java/org/mitre/synthea/engine/Generator.java
@@ -43,6 +43,7 @@ import org.mitre.synthea.modules.LifecycleModule;
 import org.mitre.synthea.world.agents.PayerManager;
 import org.mitre.synthea.world.agents.Person;
 import org.mitre.synthea.world.agents.Provider;
+import org.mitre.synthea.world.agents.behaviors.providerfinder.ProviderFinderPreferOne;
 import org.mitre.synthea.world.concepts.Costs;
 import org.mitre.synthea.world.concepts.HealthRecord.Encounter;
 import org.mitre.synthea.world.concepts.HealthRecord.EncounterType;
@@ -211,7 +212,11 @@ public class Generator {
 
   private void init() {
     if (options.state == null) {
-      options.state = DEFAULT_STATE;
+      if (ProviderFinderPreferOne.isUsingPreferredProvider()) {
+        options.state = Provider.findProviderStateByNPI(ProviderFinderPreferOne.getPreferredNPI());
+      } else {
+        options.state = DEFAULT_STATE;
+      }
     }
     int stateIndex = Location.getIndex(options.state);
     if (Config.getAsBoolean("exporter.cdw.export")) {
@@ -646,6 +651,10 @@ public class Generator {
     // Initialize person.
     Person person = new Person(personSeed);
     person.populationSeed = this.options.seed;
+
+    // Check if we need to override demographics based on preferred provider setting
+    ProviderFinderPreferOne.overrideDemographicsIfPreferredProvider(demoAttributes);
+
     person.attributes.putAll(demoAttributes);
     person.attributes.put(Person.LOCATION, this.location);
     person.lastUpdated = (long) demoAttributes.get(Person.BIRTHDATE);
--- a/src/main/java/org/mitre/synthea/world/agents/Provider.java
+++ b/src/main/java/org/mitre/synthea/world/agents/Provider.java
@@ -28,6 +28,7 @@ import org.mitre.synthea.helpers.Utilities;
 import org.mitre.synthea.world.agents.behaviors.providerfinder.IProviderFinder;
 import org.mitre.synthea.world.agents.behaviors.providerfinder.ProviderFinderNearest;
 import org.mitre.synthea.world.agents.behaviors.providerfinder.ProviderFinderNearestMedicare;
+import org.mitre.synthea.world.agents.behaviors.providerfinder.ProviderFinderPreferOne;
 import org.mitre.synthea.world.agents.behaviors.providerfinder.ProviderFinderRandom;
 import org.mitre.synthea.world.concepts.ClinicianSpecialty;
 import org.mitre.synthea.world.concepts.HealthRecord.EncounterType;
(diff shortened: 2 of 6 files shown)


src/main/java/org/mitre/synthea/world/geography/Location.java
```diff
@@ -437,6 +438,18 @@ public static String getAbbreviation(String state) {
...
+   * @param abbreviation State abbreviation. e.g. "MA"
+   * @return state name. e.g. "Massachusetts"
+   */
+  public static String getStateNameFromAbbreviation(String abbreviation) {
+    for(Entry<String, String> entry : stateAbbreviations.entrySet()) {
+      if (entry.getValue().equals(abbreviation)) return entry.getKey();
+    }
+    return null;
+  }
+
```
jawalonoski, line 452: This method duplicates:
https://github.com/synthetichealth/synthea/blob/2cc0a62b07e86bcd7019f177731201d772df233d/src/main/java/org/mitre/synthea/world/geography/Location.java#L458-L470

src/main/java/org/mitre/synthea/engine/Generator.java
```diff
@@ -646,6 +651,10 @@ public Person createPerson(long personSeed, Map<String, Object> demoAttributes)
     // Initialize person.
     Person person = new Person(personSeed);
     person.populationSeed = this.options.seed;
+
+    // Check if we need to override demographics based on preferred provider setting
+    ProviderFinderPreferOne.overrideDemographicsIfPreferredProvider(demoAttributes);
+
```
jawalonoski, line 657: Yeah, I don't actually like how this works, unless there is some reasoning that I'm not understanding.

If you really want to do this, then you should override the `options.city` above in the same way you override the `options.state`.

src/main/java/org/mitre/synthea/world/agents/Provider.java
```diff
@@ -743,6 +754,30 @@ public static List<Provider> getProviderList() {
...
+  
+        while (csv.hasNext()) {
+          Map<String,String> row = csv.next();
+          String currNpi = row.get("npi");
+
+          if (npi.equals(currNpi) && row.get("state") != null) return Location.getStateNameFromAbbreviation(row.get("state"));
+        }
+      } catch (IOException e) {
+        System.err.println("ERROR: unable to find state for provider by NPI: " + e.getMessage());
+      }
```
jawalonoski, line 774: I don't think this is something you want to catch. You probably should error out of the program if the NPI input is invalid.

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+
+  @Override
+  public Provider find(List<Provider> providers, Person person, EncounterType service, long time) {
+
+    String preferredNpi = Config.get(PREFER_ONE_NPI, null);
+
+    System.out.println("PreferredNpi: " + preferredNpi);
```
jawalonoski, line 113: This is going to print a lot...

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
```diff
@@ -0,0 +1,147 @@
...
+  public static boolean isUsingPreferredProvider() {
+    return Config.get("generate.providers.selection_behavior", "nearest").equals(Provider.PREFER_ONE);
+  }
+
+  public static boolean isIgnoringSuitable() {
+    return Boolean.valueOf(Config.get(PREFER_ONE_IGNORE_SUITABLE, "false"));
+  }
+
+  public static String getPreferredNPI() {
+    return Config.get(PREFER_ONE_NPI, null);
```
jawalonoski, line 32: You may want to check this against `isUsingPreferredProvider` because the `Config.get` could possibly return whitespace instead of `null`... although I'm not sure about this.

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
```diff
@@ -0,0 +1,147 @@
...
+
+    Provider preferredProvider = null;
+    // Find the preferred provider by NPI from the loaded list
+    // Note: This assumes the provider is within the initially loaded set.
+    for (Provider p : Provider.getProviderList()) {
+      if (preferredNpi.equals(p.npi)) {
+        preferredProvider = p;
+        break;
+      }
+    }
```
jawalonoski, line 54: You probably only want to perform this search once, ever, per run... rather than every time `find` is called... which can be often.

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
```diff
@@ -0,0 +1,147 @@
...
+    if (preferredNpi != null && !preferredNpi.isEmpty()) {
+      // first check the list passed in (if the states line up with the detault state, e.g., MA, then it may be found in the list)
+      // if it's not found in the list then look across all providers. The state will be updated on the patient and generator.
+      Provider provider = findPreferredProvider(providers, preferredNpi, person, service, time);
+      if (provider == null) provider = findPreferredProvider(Provider.getProviderList(), preferredNpi, person, service, time);
+      if (provider != null) return provider;
+    }
+
+    // If preferred NPI not set, not found in the list, or not suitable, use the fallback finder.
+    return fallbackFinder.find(providers, person, service, time);
```
jawalonoski, line 124: I don't know about this behavior.... we should discuss.

The patient's state (and potentially city) should already be set and overridden at this point, matching the provider information.... so doing a major reset of the provider list (using `Provider.getProviderList()`) raises my eyebrows a tiny bit.

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private Provider findPreferredProvider(List<Provider> providers, String preferredNpi, Person person, EncounterType service, long time) {
```
jawalonoski, line 127: Maybe I'm missing something, but why do you have `getPreferredProvider` and `findPreferredProvider`?

The first just gives you the hard-coded provider... the second does the same thing, but only if it also provides the service?

You already have the preferred provider in hand... let's discuss.

jawalonoski (changes requested):
Let's talk about the expected behavior of this finder.

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
andrequina: removed getPreferredProvider

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
andrequina: no longer resets the patient location

src/main/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOne.java
andrequina: removed

src/main/java/org/mitre/synthea/engine/Generator.java
andrequina: uses options.city and options.state in Generator.java in the init() function as an alternative to setting the default.

src/test/java/org/mitre/synthea/world/agents/behaviors/providerfinder/ProviderFinderPreferOneTest.java
```diff
@@ -0,0 +1,199 @@
...
+    preferredProvider.servicesProvided.add(EncounterType.AMBULATORY); // Offers the service
+    // Assume accepts patient by default
+
+    // Nearest Provider Setup (for fallback)
+    nearestProvider = new Provider();
+    nearestProvider.npi = NEAREST_NPI;
+    nearestProvider.city = NEAREST_CITY;
+    nearestProvider.state = NEAREST_STATE;
+    nearestProvider.getLonLat().setLocation(NEAREST_PROVIDER_COORDS.getX(), NEAREST_PROVIDER_COORDS.getY()); // Use static variable
+    nearestProvider.servicesProvided.add(EncounterType.AMBULATORY); // Offers the service
```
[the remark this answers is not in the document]
andrequina: The reason to find nearest is that it the preferred provider may not provide the service being requested or may not accept the provider. In that case we'd want the patient to go somewhere else.  

There is a setting to ignore suitable which forces the use of the preferred provider even if they don't accept the patient or have the service which defaults to true.  So you would only fall back to nearest if this was set to false.

generate.providers.prefer_one.ignore_suitable

`          
            // Check if the preferred provider offers the service and accepts the patient
            if (provider.hasService(service) && provider.accepts(person, time)) {
`

Issue #1560, opened 2025-02-12
Unexpected results from CSV Exporter included/excluded files settings


dehall:
### What happened?

The CSV Exporter has 2 settings to configure which files are exported, with the following defaults:
```properties
exporter.csv.included_files =
exporter.csv.excluded_files = patient_expenses.csv
```

If both settings are populated, an error is shown:
```
CSV exporter: Included and Excluded file settings are both set -- ignoring both
```

This means that running synthea with an include list as a command-line setting will produce unexpected results:

```
./run_synthea --exporter.csv.included_files "patients.csv,procedures.csv"
```
--> all files are output. This is unexpected, that error is in the output but isn't necessarily obvious, and the workaround is not obvious either: Either update the properties file or add an additional command line arg `--exporter.csv.excluded_files " "`. It's worse for JAR users because they don't have access to the properties file

### Environment

```markdown
- OS:
- Java:
```

### Relevant log output

```shell

```

facuerrobidart:
Hey, I want to tackle this issue. Can I get assigned to it?

dehall:
> Hey, I want to help tackling this issue. Can I get assigned to it?

Hi @facuerrobidart , yes please feel free to take this on and send a PR! We'd greatly appreciate it.

Off the top of my head though I'm not sure what the best fix for this would be, so it may be worth talking about that before implementing anything.

facuerrobidart:
Hi @dehall, just got the chance to take a look at that implementation. I can imagine some scenarios:

1. Only ignore both options if there is an overlap.
2. Ignore both if one of the two is not the default.
3. Check for overlap between options, and only ignore the file that appears on both.

I'd prefer number 3 of all three, but I'm open to iterating over those.

dehall:
Ok yeah I agree 3 is the best, but instead of just silently ignoring the overlap I'd suggest at least a big error message, possibly an exception

facuerrobidart:
Cool, I like the idea. I'll implement this change

PR #1573, opened 2025-05-05
Allow included and excluded files in CSV exporter if there is no overlap

facuerrobidart:
The `CSVExporter` has been updated to support the simultaneous setting of `included_files` and `excluded_files`. The implementation now includes a check for any overlap between these two lists, raising an error if and only if common elements are found.

Fixes:  https://github.com/synthetichealth/synthea/issues/1560


dehall:

PR #1591, opened 2025-07-07
Fix #570: Add ExplanationOfBenefit to CCDA export

### LINKED ISSUE #570: Add "resourceType": "ExplanationOfBenefit" to CCDA export
The FHIR export contains explanation of benefits, while the CCDA export does not.
A useful feature would be to include the same Explanation of Benefit and other claims data in the CCDA export

ShaileshKumar005:
🔧 Summary
This PR resolves [Issue #570](https://github.com/synthetichealth/synthea/issues/570) by adding support for the ExplanationOfBenefit section in CCDA exports. This brings CCDA output more in line with the existing FHIR output, ensuring consistency across data formats.

✅ What Was Done
Step 1: Enabled the CCDA export feature in the project configuration so that CCDA files are generated during patient simulation.

Step 2: Appended a placeholder section for ExplanationOfBenefit in the ccda.ftl template, under the structured body. This allows the CCDA output to reflect that the resource is supported, even if data isn't populated yet.

Step 3: Ran the generator and confirmed that:

CCDA XML files were successfully created in the output folder.

The Explanation Of Benefit section appears at the end of the XML file.

There were no regressions or disruptions to existing functionality.

💡 Why This Matters
Ensures consistent support for the ExplanationOfBenefit resource across both FHIR and CCDA outputs.

Helps downstream teams relying on CCDA files for interoperability, data pipelines, or clinical system integrations.

Prepares the groundwork for future expansion where real ExplanationOfBenefit data can be dynamically included.

📌 Before the Fix:
CCDA export was not generated because the configuration exporter.ccda.export was set to false by default.
As a result, no .xml files were present in the output/ccda/ directory.

### 📸 After (✅ Issue #570 Fixed – ExplanationOfBenefit added)

image
> The `Explanation Of Benefit` section now appears in the exported CCDA file under `/output/ccda/...xml`.  
> This confirms the successful addition of the missing section, aligning CCDA exports with FHIR output structure.

### DIFF
--- a/src/main/resources/templates/ccda/ccda.ftl
+++ b/src/main/resources/templates/ccda/ccda.ftl
@@ -118,7 +118,7 @@
       <#else>
         <#include "procedures_no_current.ftl" parse=false>
       </#if>
-	    <#if ehr_encounters?has_content>
+      <#if ehr_encounters?has_content>
         <#include "encounters.ftl">
       </#if>
       <#if ehr_vital_signs?has_content>
@@ -143,6 +143,17 @@
       <#if ehr_functional_statuses?has_content>
         <#include "functional_status.ftl">
       </#if>
+
+      <!-- 🔧 Fix for Issue #570 -->
+      <component>
+        <section>
+          <title>Explanation Of Benefit</title>
+          <text>
+            <resourceType>${ehr_explanationofbenefit.resourceType}</resourceType>
+          </text>
+        </section>
+      </component>
+
     </structuredBody>
   </component>
 </ClinicalDocument>


ShaileshKumar005:
Hi maintainers 👋

I’ve created this PR to address Issue #570 by adding a placeholder for the ExplanationOfBenefit section in the CCDA export. This aligns CCDA output with existing FHIR output formats.

Would appreciate it if someone could review this when you get the chance. Thanks!

jawalonoski:
Thank you for the pull request... we'll take a look at this when we have the opportunity!

src/main/resources/templates/ccda/ccda.ftl
```diff
@@ -143,6 +143,17 @@
       <#if ehr_functional_statuses?has_content>
         <#include "functional_status.ftl">
       </#if>
+
+      <!-- 🔧 Fix for Issue #570 -->
+      <component>
```
KeeyanGhoreshi, line 148: This component should probably be contained in its own ftl file and be included like the others, checking to make sure the attribute key has content before inclusion.

KeeyanGhoreshi (commented):
The changes here cause an error on my machine, shown below:

`Caused by: freemarker.core.InvalidReferenceException: The following has evaluated to null or missing:
==> ehr_explanationofbenefit  [in template "ccda.ftl" at line 152, column 29]`

The key is not defined in the export file as an attribute on the Person resource so it is failing to find it.

Issue #1545, opened 2024-12-27
Duplicate IDs in Several Tables


katy-sadowski:
### What happened?

I generated a 1 million patient Synthea dataset following Basic Setup instructions: https://github.com/synthetichealth/synthea/wiki/Basic-Setup-and-Running.  I output to csv files.  Upon loading the csvs into duckdb and running some tests I discovered duplicated primary keys in the following tables: imaging_studies, claims_transactions, claims, encounters.

Is this a known issue?

### Environment

```markdown
- OS: macOS 14.6.1
- Java: java 15.0.1
```

dehall:
No I don't think we've seen this recently. Are you able to pull just the rows with duplicated IDs and attach those here?

katy-sadowski:
Sure!  Files are attached.  For the claims tables and encounters, you'll see the dupe IDs are associated with completely different records.  For imaging, the records appear to be full dupes.  Also there are hundreds of copies of some of the imaging dupes, whereas for claims and encounters each duplicated ID is associated with only 2 rows.

[claims_transactions.csv](https://github.com/user-attachments/files/18263805/claims_transactions.csv)
[claims.csv](https://github.com/user-attachments/files/18263807/claims.csv)
[encounter.csv](https://github.com/user-attachments/files/18263808/encounter.csv)
[imaging.csv](https://github.com/user-attachments/files/18263810/imaging.csv)

dehall:
Thanks! There are two separate issues here. The imaging dupes issue should be a trivial fix, that's a nested loop and we're using the wrong index in one spot. The duplicate IDs across different records is trickier. I'm worried our ID generation approach just isn't random enough for a large dataset like this, so we'll have to think about how we can improve that

katy-sadowski:
Sounds good!  Thanks so much for your quick reply!  My usecase for the million patient dataset was testing performance on an ETL, so not a huge rush on my end for that second improvement.  Though it might be a good idea to assess above which threshold the unique ID generation starts to break down and recommend users not to go above that number.

PR #1601, opened 2025-07-31
Reduce chances of duplicate uuids in large populations

dehall:
Addresses the duplicate UUID issue from #1545 

Our UUID generation uses the seeded random number generator to ensure reproduceability, and picks 2 randLong()s to pass into the UUID constructor. For small populations this works ok but in larger populations we wind up with UUID collisions within the same resource type.

I believe what's happening when we get duplicate UUIDs is we're seeing two separate patients (two separate RNGs with separate seeds) where their RNGs have lined up in terms of state and where they are in that sequence, and so they produce the same 2 long values in a row, resulting in the same UUID. In small populations the range of numbers is so large that collisions are unlikely, but when we get up to say 1,000,000 patients, we start to see a small handful of collisions. By replacing the first randLong with the seed, we eliminate the possibility of duplicate UUIDs happening between different patients. It does mean there is maybe a higher possibility of duplicate UUIDs for the same patient seed, but that should be a lot more rare. I.e., the number of random calls for a single patient tends to be many orders of magnitude lower than the number of possible random values from a Random.

To test this I ran 2 populations of 1,000,000 records, 1 before and 1 after, and checked the IDs in the encounters file with this command to show duplicates:
`cut -d',' -f1 output/csv/encounters.csv | sort | uniq -d`
There should be some IDs printed in the before, but none in the after. (My test instance had 6 duplicates.) Note this doesn't even consider dups across different resource types.
I'm still trying to think of a way to formally prove this is better with a test case, so I'm open to suggestions on that front.

Note the change to the random class produces a different sequence of random values so one hardcoded value in a test case had to change as well.

### DIFF
--- a/src/main/java/org/mitre/synthea/helpers/DefaultRandomNumberGenerator.java
+++ b/src/main/java/org/mitre/synthea/helpers/DefaultRandomNumberGenerator.java
@@ -68,7 +68,7 @@ public class DefaultRandomNumberGenerator implements RandomNumberGenerator, Seri
 
   @Override
   public UUID randUUID() {
-    return new UUID(randLong(), randLong());
+    return new UUID(seed, randLong());
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


jawalonoski (approved):
I've been unable to replicate the error (even after generating 2 million patients). Nonetheless, it seems to have been recreated by others. I've tested the fix on this branch with another million patients. I can confirm the error does not manifest in the fix either. I suppose I'll merge this since others have had success with it, and it seems to do no harm.

Issue #1342, opened 2023-07-26
Make `-r` argument more useful.


eedrummer:
### Requested Feature

Synthea currently allows users to specify a reference time, using the `-r` argument. This argument currently takes a date in YYYYMMDD format. If this argument is being used to reproduce a previous run, it isn't particularly helpful unless that run also used a reference time.

If a user created a "plain" Synthea without specifying a reference date, there is no way to replicate it since Synthea's internal representation of the date is a full time stamp, and not just a date.

It would be nice to change this argument or add another argument that would allow users to pass in a `long` in the same way it is used in Synthea.

jawalonoski:
Alternatively, we could make the default reference date not a full time stamp, and round to the day.

PR #1611, opened 2025-09-05
Make default reference time precise to the day

Jammjammjamm:
This PR address #1342. It makes the default reference time precise to the day, so that a run with a default reference time can be reproduced using `-r`.

On master:
```
> ./run_synthea -p 1
Reference Time: 1757076199758
> ./run_synthea -p 1 -r 20250905
Reference Time: 1757030400000
```

On this branch:
```
> ./run_synthea -p 1
Reference Time: 1757030400000
> ./run_synthea -p 1 -r 20250905
Reference Time: 1757030400000
```


Jammjammjamm:
The current reference times are all set to UTC, so I was copying that. I'm not sure why they're set to UTC, though.

The current reference time is already output during a run. Are you asking me to change the output label from `Reference Time` to `Reference Date`?

Issue #1636, opened 2025-12-03
Synthea fails to run on JDK 25


dehall:
### What happened?

JDK 25 is the latest LTS version of java, released a couple months ago. Unfortunately the Gradle compatibility matrix means that there's no version of Gradle that works with both JDK 11 (our currently listed requirement) and JDK 25. https://docs.gradle.org/current/userguide/compatibility.html - everything prior to JDK 17 was only supported through Gradle 8.x and and JDK 25 requires Gradle 9.1+
Because JDK 11 is so far out of primary support (2 years, per https://endoflife.date/oracle-jdk ) I think we should consider dropping support for 11 in favor of 25, and make JDK 17 our new minimum requirement

### Environment

```markdown
- OS:
- Java:
```

### Relevant log output

```shell

```

jawalonoski:
I'm okay with phasing out Java 11 and moving to Java 17 as our minimum.

We'll need to make sure we do a version change though since we'll be breaking backwards compatibility.

@hadleynet Do you concur?

hadleynet:
Agreed. Major or minor version increment? Seems like a major version to me if a JDK 11 user will have to update their environment.

jawalonoski:
Yeah, probably a major version change just to be on the safe side.

dehall:
Addressed by #1637

PR #1637, opened 2025-12-03
Upgrade to Gradle 9 for JDK 25 support, remove JDK 11 support

dehall:
This PR bumps the Gradle version to the latest as of today, 9.2.1, to enable support for JDK 25, the most recent LTS version of Java. Unfortunately this version of Gradle no longer supports JDK 11, so our new minimum will have to be JDK 17.

One group of dependencies needed to be updated to work with JDK 25 (GraalVM, which powers the JavaScript engine that runs Graphviz and one of the Flexporter actions) but other than that everything should work as is. Many thanks to those who upgraded the build.gradle file already to get it compatible with Gradle 9 for making this easy.

Addresses #1636 .


jawalonoski (approved):
Checked out and tested. Works as advertised.

I think we should wait to merge for the Synthea major version change as discussed elsewhere.

I noticed there are also gradle deprecation warnings in `./build/reports/problems/problem-report.html`... It wasn't clear how much effort it would take to address these warnings.

dehall:
One of the 4 new deprecation warnings is trivial to fix so I pushed that up.
The rest are on the 3 lines in this section, apparently "archives" is deprecated:
```
artifacts {
    archives jar
    archives sourceJar
    archives javadocJar
}
```

It seems like there should be a straightforward replacement for this ( https://docs.gradle.org/9.2.1/userguide/upgrading_version_9.html#archives-configuration ) but I'm not immediately sure of all of the implications, especially for things like maven publishing. I'll keep looking as time permits

dehall:
With that last commit, the remaining deprecation warnings are addressed. Since the migration docs (linked in previous comment) were not super clear, ChatGPT (sorry) pointed to the best answer: there are built-in functions that do the same thing we were already doing a different way. One minor difference is we had a task `sourceJar` but the new equivalent "built-in" one is `sourcesJar`. I don't think anything referenced that directly nor would any user ever invoke it so I don't think this matters.

My main concern about this part of the change was in publishing to Maven Central, which I'm not sure I have a way to officially test without doing a real publish, but comparing the results of "publish --dry-run" and "publishToMavenLocal" before & after, I think it should be good. @hadleynet does this look reasonable to you? Unfortunately I don't see a way to exclude the shadowJar/uberJar when doing a publish, we may need to look into an alternate shadowJar plugin if that's something that becomes a big issue.

`./gradlew publish --dry-run` -->
Before:
```
:versionTxt SKIPPED
:compileJava SKIPPED
:processResources SKIPPED
:classes SKIPPED
:jar SKIPPED
:shadowJar SKIPPED
:generateMetadataFileForSyntheaPublication SKIPPED
:generatePomFileForSyntheaPublication SKIPPED
:javadoc SKIPPED
:javadocJar SKIPPED
:sourceJar SKIPPED
:signSyntheaPublication SKIPPED
:publishSyntheaPublicationToMavenRepository SKIPPED
:publish SKIPPED
```

After:

```
:versionTxt SKIPPED
:compileJava SKIPPED
:processResources SKIPPED
:classes SKIPPED
:jar SKIPPED
:javadoc SKIPPED
:javadocJar SKIPPED
:shadowJar SKIPPED
:sourcesJar SKIPPED
:generateMetadataFileForSyntheaPublication SKIPPED
:generatePomFileForSyntheaPublication SKIPPED
:signSyntheaPublication SKIPPED
:publishSyntheaPublicationToMavenRepository SKIPPED
:publish SKIPPED
```
(slightly different order but it's the same 14 tasks. note the metadata file is generated after the sources & javadoc jars now)

---

`rm -rf ~/.m2/repository/org/mitre/synthea &&
 ./gradlew publishToMavenLocal --offline &&
 tree ~/.m2/repository/org/mitre/synthea` -->
Before:
```
/Users/dehall/.m2/repository/org/mitre/synthea
└── synthea
    ├── 3.4.1-SNAPSHOT
    │   ├── maven-metadata-local.xml
    │   ├── synthea-3.4.1-SNAPSHOT-javadoc.jar
    │   ├── synthea-3.4.1-SNAPSHOT-sources.jar
    │   ├── synthea-3.4.1-SNAPSHOT-with-dependencies.jar
    │   ├── synthea-3.4.1-SNAPSHOT.jar
    │   ├── synthea-3.4.1-SNAPSHOT.module
    │   └── synthea-3.4.1-SNAPSHOT.pom
    └── maven-metadata-local.xml

3 directories, 8 files
```

After:
```
/Users/dehall/.m2/repository/org/mitre/synthea
└── synthea
    ├── 3.4.1-SNAPSHOT
    │   ├── maven-metadata-local.xml
    │   ├── synthea-3.4.1-SNAPSHOT-javadoc.jar
    │   ├── synthea-3.4.1-SNAPSHOT-sources.jar
    │   ├── synthea-3.4.1-SNAPSHOT-with-dependencies.jar
    │   ├── synthea-3.4.1-SNAPSHOT.jar
    │   ├── synthea-3.4.1-SNAPSHOT.module
    │   └── synthea-3.4.1-SNAPSHOT.pom
    └── maven-metadata-local.xml

3 directories, 8 files
```

(same files; the .module file is mostly the same but contains entries for the sources & javadoc jars where it previously didn't, I assume that's the result of the change in order noted above. this seems like an improvement to me)

hadleynet:
This looks good to me. If we are going to merge this as part of a 4.0 release then I could test the publish steps as part of that process before merging.

PR #1656, opened 2026-05-11
Reorder condition onset and diagnosing encounter in asthma module

### LINKED ISSUE #1655: "Childhood Asthma"-Patients aren't switching to "Asthma"
### What happened?

I generated a population of 10000 with exporter.years_of_history = 0.
I recorded 14 patients with "Asthma (disorder)" and 580 patients with "Childhood asthma (disorder)".

While the "Childhood asthma" diagnosis consistently ends between 19 and 20 years of age, the **transition to "Asthma (disorder)" does not occur**. They also **remain on medication despite having no active diagnosis**. 

My guess is that the transition logic in the asthma.json module is broken. The system forces an administrative STOP to the pediatric code when reaching adulthood, but does not initiate the adult code. Maybe there is something wrong with the need for another "Wellness Encounter" to really get the "new" diagnosis. But I am not quite sure.

Greetings and thanks for your work!

### Environment

```markdown
- OS: macOS 26.2
- Java: JDK 26
```

### Relevant log output

```shell

```

hadleynet:
Reorder two states in the branch where childhood asthma becomes lifelong asthma. Previously the diagnosing encounter preceded the condition onset, this PR switches the order so the onset is before the diagnosing encounter.

Fixes: #1655

### DIFF
--- a/src/main/resources/modules/asthma.json
+++ b/src/main/resources/modules/asthma.json
@@ -396,12 +396,12 @@
       "type": "SetAttribute",
       "attribute": "asthma_type",
       "value": "lifelong",
-      "direct_transition": "Next_Adult_Welness_Encounter"
+      "direct_transition": "Lifelong_Asthma_Begins"
     },
     "Next_Adult_Welness_Encounter": {
       "type": "Encounter",
       "wellness": true,
-      "direct_transition": "Lifelong_Asthma_Begins",
+      "direct_transition": "End_Wellness_1",
       "reason": "asthma_condition"
     },
     "Lifelong_Asthma_Begins": {
@@ -415,7 +415,7 @@
           "display": "Asthma (disorder)"
         }
       ],
-      "direct_transition": "End_Wellness_1"
+      "direct_transition": "Next_Adult_Welness_Encounter"
     },
     "Potential_Asthma_Attack": {
       "type": "Simple",


lucabusch:
It worked for me. I now have a lot more Patients with "Asthma (disorder)". With n = 10000 I previously had around 15 asthma patients, I now have 389. 
It also fits the medication way better now. The only ones with asthma medication beyond 18 years without condition "Asthma (disorder)" are in the age group 18-19 (fits follow-up latency).
Thanks a lot for the fix.

jawalonoski (approved):

jawalonoski:
Looks good to me, merge it if you're ready @hadleynet

dehall:
Don't hold up merging for this but I have a few thoughts: Seems to me like "a Condition being diagnosed within the current encounter" should work. There might be other places we do that and it's also silently not working. 
Is it possible it used to work and changes to the engine broke it? Maybe the fact that the target_encounter is the same as the current encounter confuses things? If, instead of reordering the states, you just removed the condition.target_encounter, would that fix it as well? (I'd test these out myself but am without a computer for the next week)

hadleynet:
@dehall I think you are right that the unchanged module *should* work. In the `ConditionOnset.process` (inherited from `OnsetState`) there's this:

```
if (targetEncounter == null || targetEncounter.trim().length() == 0
    || (encounter != null && targetEncounter.equals(encounter.name))) {
  diagnose(person, time);
```

I'll dig in further when I get a chance.

PR #1666, opened 2026-05-27
Fix: Filter organizations and providers to only include exported patients

### LINKED ISSUE #1378: Using a Keep Module results in extra data in hospital/practitionerInfo files
(originally identified in https://github.com/synthetichealth/synthea/discussions/1377 )

When using a Keep Module that filters a lot of patients out, the hospitalInfo and practitionerInfo files still contain info that was associated only to patients that were generated but not exported.  (This may also be the case when using `generate.only_dead_patients=true` or `generate.only_dead_patients=true` ) If the Keep Module is looking for something especially rare the files can get very large. 

There should be a way to filter the hospital and practitioner Info files to only include the necessary records for the patients that were actually exported

denissaurio:
Fixes #1378

When using a Keep Module, the hospitalInfo and practitionerInfo files 
included data from patients that were generated but not exported.

This fix tracks exported provider and clinician IDs in a Set within CSVExporter, and uses that to filter exportOrganizationsAndProviders to only include records linked to exported patients.

A regression test was added to verify the fix.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,10 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Ignored default folder with query files
+/queries/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CompilerConfiguration">
+    <bytecodeTargetLevel target="17" />
+  </component>
+</project>
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="GradleMigrationSettings" migrationVersion="1" />
+  <component name="GradleSettings">
+    <option name="linkedExternalProjectsSettings">
+      <GradleProjectSettings>
+        <option name="externalProjectPath" value="$PROJECT_DIR$" />
+        <option name="gradleJvm" value="#JAVA_HOME" />
+        <option name="modules">
+          <set>
+            <option value="$PROJECT_DIR$" />
+          </set>
+        </option>
+      </GradleProjectSettings>
+    </option>
+  </component>
+</project>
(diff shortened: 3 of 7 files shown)


hadleynet (changes requested):
The functionality looks good but the PR results in some checkstyle violations. Could you run the gradle checkstyleMain and checkstyleTest jobs and fix the warnings.

denissaurio:
Yes, still in progress! Just pushed a fix for the CI failure (test state cleanup after `filterOrganizationsAndProviders`). The workflow is pending approval to run — could you approve it when you get a chance? Thanks!

dehall:
Hi, so this proposed change is ok but unfortunately it doesn't really solve the full issue as noted in #1378. Two things:
1) If I'm reading this correctly, this means that lines will only be added to `providers.csv` and `organizations.csv` if the corresponding provider/organization was already referenced in some patient's Encounter. So this by itself sounds good, there's really no reason to include those providers that are not referenced by anything. But the original issue was really talking about the utilization of those providers. If we're looking at CSVs then the relevant columns are `providers.csv`: ENCOUNTERS,PROCEDURES and `organizations.csv`: UTILIZATION. Eg:

```
Id,ORGANIZATION,NAME,GENDER,SPECIALITY,ADDRESS,CITY,STATE,ZIP,LAT,LON,ENCOUNTERS,PROCEDURES
c0b3f274-3c4c-3d0a-bb8a-3c5394ac2176,74ab949d-17ac-3309-83a0-13b4405c66aa,Ted955 Reilly981,M,GENERAL PRACTICE,881 Main Street,Fitchburg,MA,01420,42.586487,-71.80521,113,0
```

To step through and illustrate the issue, imagine we have just this one provider, and we're trying to generate just one patient using a Keep Module:
- Patient attempt 1 has 3 encounters with the provider, but doesn't meet the Keep criteria so the patient is not exported
- Patient attempt 2 has 1 encounter with the provider, but doesn't meet the Keep criteria so the patient is not exported
- Patient attempt 3 has 6 encounters with the provider, and does meet the criteria so this one is exported, and the simulation is done.

So what we want to see in the provider output is a utilization of 6 encounters, representing the total from patients that were actually exported. But as of today, what we'll see is 10. I don't think it's an easy fix but I haven't looked into it in a long time.

2) This only affects the CSV Exporter, but the original issue occurs in every exporter. (It's really an issue with the data before it even gets to the exporter) Ideally we'd have a more global fix and not have to fix each exporter separately.


So I do think this is a good change and I don't object to merging, but I don't want to close 1378 just yet.

PR #1671, opened 2026-06-10
fix: clamp blood pressure to physiological minimum to prevent negative values

### LINKED ISSUE #1413: RuntimeException: unexpected value -1 within Framingham.stroke10Year (VitalSign.SYSTOLIC_BLOOD_PRESSURE is -1)
### What happened?

When running with the following parameters, I'm getting a RuntimeException as shown in the log output section. 

--exporter.fhir.export false --exporter.csv.export true --exporter.clinical_note.export false -p 5618 Kentucky Campbellsville -s 1705316793119 -cs 1705316793119 

Note: in order to reproduce the issue locally, you also need to change the code [here](https://github.com/synthetichealth/synthea/blob/16ee9bcb6d983c35fc4c9399fd5c464a2760029a/src/main/java/App.java#L95) to `options.referenceTime = 1705316793119L;` to ensure that the reference time is also set to exactly the same as in the original run. I found that merely passing in yyyyMMdd for `-rf` is not sufficient to reproduce this issue locally as the scenario seems to be very sensitive to the exact value of `referenceTime`.

From what I could find debugging locally, it looks like in [this code](https://github.com/synthetichealth/synthea/blob/16ee9bcb6d983c35fc4c9399fd5c464a2760029a/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java#L97-L99) for BloodPressureValueGenerator, the following values are seen:
- `baseline` is ~ 148
- `getMedicationImpacts(person)` returns -110
- `getLifestyleImpacts(person, baseline, time)` returns -15
- `getVariation(person, time)` returns -34 or so

Therefore the final value computed for the BP ends up being negative. Eventually that translates to a -1 for `bp` in the context of `Framingham.stroke10Year`, which in turn finally causes the `RuntimeException: unexpected value -1` exception.

### Environment

```markdown
- OS: Rocky Linux 8.8
Linux version 4.18.0-477.27.1.el8_8.x86_64 (mockbuild@iad1-prod-build001.bld.equ.rockylinux.org) (gcc version 8.5.0 20210514 (Red Hat 8.5.0-18) (GCC)) #1 SMP Wed Sep 20 15:55:39 UTC 2023

- Java: openjdk 11.0.20 2023-07-18 LTS
OpenJDK Runtime Environment (Red_Hat-11.0.20.0.8-1) (build 11.0.20+8-LTS)
OpenJDK 64-Bit Server VM (Red_Hat-11.0.20.0.8-1) (build 11.0.20+8-LTS, mixed mode, sharing)
```

### Relevant log output

```shell
Population: 5618
Seed: 1705316793119
Provider Seed:1705316793119
Reference Time: 1705316793119
Location: Campbellsville, Kentucky
Min Age: 0
Max Age: 140
java.lang.RuntimeException: unexpected value -1 for data [0, 106, 113, 118, 124, 130, 136, 143, 151, 162, 177]
  at org.mitre.synthea.modules.calculators.Framingham.getIndexForValueInRangelist(Framingham.java:468)
  at org.mitre.synthea.modules.calculators.Framingham.stroke10Year(Framingham.java:515)
  at org.mitre.synthea.modules.CardiovascularDiseaseModule.calculateStrokeRisk(CardiovascularDiseaseModule.java:113)
  at org.mitre.synthea.modules.CardiovascularDiseaseModule.process(CardiovascularDiseaseModule.java:40)
  at org.mitre.synthea.engine.Generator.updatePerson(Generator.java:712)
  at org.mitre.synthea.engine.Generator.createPerson(Generator.java:668)
  at org.mitre.synthea.engine.Generator.generatePerson(Generator.java:478)
  at org.mitre.synthea.engine.Generator.lambda$run$3(Generator.java:383)
  at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
  at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
  at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
  at java.base/java.lang.Thread.run(Thread.java:829)
```

MadhukarJukanti:
## Summary
  - When a person has many active hypertension medications, their combined
    systolic impacts (each up to -16 mmHg) plus lifestyle reductions and
    intra-day variation can push the calculated BP below zero
  - The Framingham stroke/CVD lookup tables have no entry for negative values
    and throw `RuntimeException: unexpected value -1 within Framingham.stroke10Year`
  - Added `MIN_SYSTOLIC_BP = 70.0` and `MIN_DIASTOLIC_BP = 40.0` floor
    constants and clamp the computed value with `Math.max` before caching

  ## Root Cause
  In `BloodPressureValueGenerator.getValue()`, the sum of baseline + medication
  impacts + lifestyle impacts + variation had no lower bound. With all 21 HTN
  drugs active (seeds 1705316793119 and 1705568678422 from the issue), the
  combined reduction exceeds the baseline, producing a negative value that
  crashes `getIndexForValueInRangelist` in Framingham.java.

  ## Changes
  - `src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java`:
    clamp computed BP to physiological minimum before caching
  - `src/test/java/org/mitre/synthea/modules/VitalsValueGeneratorTest.java`:
    strengthen existing no-op print loop into a real assertion + add regression
    test activating all 21 HTN drugs and asserting both systolic ≥ 70 and
    diastolic ≥ 40 over a full simulated year

  ## Test plan
  - [ ] `testSystolicTrendGenerator` — systolic always ≥ 70 over 100 days
  - [ ] `testSystolicBloodPressureNeverNegativeWithManyMedications` — all 21 drugs
    active, systolic ≥ 70 and diastolic ≥ 40 over 365 days
  - [ ] Existing test suite passes

  Fixes #1413

### DIFF
--- a/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
+++ b/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
@@ -1,4 +1,4 @@
-package org.mitre.synthea.modules;
+﻿package org.mitre.synthea.modules;
 
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Table;
@@ -89,6 +89,11 @@ public class BloodPressureValueGenerator extends ValueGenerator {
   /** Cycle time for blood pressure variation in milliseconds. */
   private static final long CYCLE_TIME = Utilities.convertTime("hours", 12);
 
+  /** Minimum physiologically plausible systolic blood pressure in mmHg. */
+  private static final double MIN_SYSTOLIC_BP = 70.0;
+  /** Minimum physiologically plausible diastolic blood pressure in mmHg. */
+  private static final double MIN_DIASTOLIC_BP = 40.0;
+
   // simple 1-value cache, so that we get consistent results when called with the same timestamp
   // note that consistency is not guaranteed if we go time A -> time B -> time A across modules.
   // in that case we'd need a bigger cache
@@ -125,10 +130,13 @@ public class BloodPressureValueGenerator extends ValueGenerator {
 
     double baseline = calculateBaseline(person);
 
-    cachedValue = baseline
-      + getMedicationImpacts(person)
-      + getLifestyleImpacts(person, baseline, time)
-      + getVariation(person, time);
+    double computed = baseline
+        + getMedicationImpacts(person)
+        + getLifestyleImpacts(person, baseline, time)
+        + getVariation(person, time);
+
+    double minValue = sysDias == SysDias.SYSTOLIC ? MIN_SYSTOLIC_BP : MIN_DIASTOLIC_BP;
+    cachedValue = Math.max(minValue, computed);
 
     cacheTime = time;
     return cachedValue;
--- a/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
+++ b/src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
@@ -89,11 +89,6 @@ public class BloodPressureValueGenerator extends ValueGenerator {
   /** Cycle time for blood pressure variation in milliseconds. */
   private static final long CYCLE_TIME = Utilities.convertTime("hours", 12);
 
-  /** Minimum physiologically plausible systolic blood pressure in mmHg. */
-  private static final double MIN_SYSTOLIC_BP = 70.0;
-  /** Minimum physiologically plausible diastolic blood pressure in mmHg. */
-  private static final double MIN_DIASTOLIC_BP = 40.0;
-
   // simple 1-value cache, so that we get consistent results when called with the same timestamp
   // note that consistency is not guaranteed if we go time A -> time B -> time A across modules.
   // in that case we'd need a bigger cache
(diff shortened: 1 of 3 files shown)


src/main/java/org/mitre/synthea/modules/BloodPressureValueGenerator.java
```diff
@@ -89,6 +89,11 @@ public enum SysDias {
   /** Cycle time for blood pressure variation in milliseconds. */
   private static final long CYCLE_TIME = Utilities.convertTime("hours", 12);
 
+  /** Minimum physiologically plausible systolic blood pressure in mmHg. */
+  private static final double MIN_SYSTOLIC_BP = 70.0;
+  /** Minimum physiologically plausible diastolic blood pressure in mmHg. */
+  private static final double MIN_DIASTOLIC_BP = 40.0;
```
jawalonoski: Do you have citations for these values? We want to ensure any hard-coded limits could include people with conditions such as chronic hypotension, etc. Even possibly those in some type of shock or end-stage care.

I'd also want to check to see how these limits would interact with editing the config files...

https://github.com/synthetichealth/synthea/blob/d02dcd65b0c8cd96d91ae7660f3570daf0fb41a1/src/main/resources/biometrics.yml#L14-L20

MadhukarJukanti: Thanks for the feedback! I've addressed both points in the latest commit:
 
 1. Config interaction: Removed the hardcoded MIN_SYSTOLIC_BP = 70.0 and MIN_DIASTOLIC_BP = 40.0 constants. The floor is now derived from the existing NORMAL_SYS_BP_RANGE[0] and NORMAL_DIA_BP_RANGE[0]  fields, which are already loaded from biometrics.yml. This means users who configure lower normal BP ranges (e.g. for chronic hypotension or end-stage care populations) will automatically get an
  appropriate floor — no hardcoded override.

 2. Defensive guard in Framingham: Added an early-return in getIndexForValueInRangelist that returns index 0 (lowest-risk bucket) for any value below the table minimum, rather than throwing. This acts as a
second line of defence for any future edge case that reaches the calculator with an out-of-range value.

MadhukarJukanti: @jawalonoski Just following up on this - happy to make any additional changes if needed. Let me know if there's anything blocking the review

PR #1686, opened 2026-07-12
fix Patient identifiers (SSN, DRIVERS, PASSPORT) are not unique — collisions at population scale

### LINKED ISSUE #1685: Patient identifiers (SSN, DRIVERS, PASSPORT) are not unique — collisions at population scale
### What happened?

The synthetic patient identifiers `SSN`, `DRIVERS`, and `PASSPORT` in `patients.csv` are **not unique** — they collide frequently at population scale because each is drawn from a small keyspace with **no uniqueness/retry check**.

Generating a ~100k-living-patient population (115,654 patient rows in `patients.csv`) produced:

| Column | Format | Keyspace | Non-blank values | Distinct | Duplicate rows |
|---|---|---:|---:|---:|---:|
| `SSN` | `999-[10-99]-[1000-9999]` | 810,000 | 115,654 | 107,800 | **7,854** |
| `DRIVERS` | `S999[10000-99999]` | 90,000 | 96,225 | 59,051 | **37,174** |
| `PASSPORT` | `X[0-89999999]X` | 90,000,000 | 90,600 | 90,544 | **56** |

These are birthday-paradox collisions and appear reliably in any large population.

The most serious case is **`DRIVERS`: the keyspace (90,000) is *smaller* than the number of license-holders (96,225)**, so uniqueness is impossible in the current format for any population with more than 90k licensed drivers — no amount of retrying can fix it without widening the format.

Note: `SSN`/`DRIVERS`/`PASSPORT` appear only in `patients.csv` (every other table keys patients by UUID `Id`), so this is a data-quality issue for anyone using these fields as natural keys / for record-linkage or deduplication testing, rather than a referential-integrity break.

### Root cause

Each identifier is generated with a single random draw and stored directly, with no check against previously issued values:

- SSN — [`LifecycleModule.java#L191`](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/modules/LifecycleModule.java#L191)
  ```java
  String ssn = "999-" + ((person.randInt(99 - 10 + 1) + 10)) + "-"
      + ((person.randInt(9999 - 1000 + 1) + 1000));
  ```
- DRIVERS — [`LifecycleModule.java#L307`](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/modules/LifecycleModule.java#L307)
  ```java
  String identifierDrivers = "S999" + ((person.randInt(99999 - 10000 + 1) + 10000));
  ```
- PASSPORT — [`LifecycleModule.java#L328`](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/modules/LifecycleModule.java#L328)
  ```java
  String identifierPassport = "X" + (person.randInt(99999999 - 10000000 + 1) + "X");
  ```

Because each `Person` has its own seeded RNG, there is no global registry of issued identifiers, so collisions across people are never detected. (This is the same class of problem addressed for output UUIDs in #1545 / #1601, but for the demographic identifiers, which are drawn from far smaller spaces and so collide much more often.)

### Bonus: PASSPORT low-bound offset appears to be dropped

Line 328 reads `... + (person.randInt(99999999 - 10000000 + 1) + "X")`. Unlike the SSN and DRIVERS lines, the intended low-bound offset (`+ 10000000`) is not applied — instead the int is string-concatenated with `"X"`. The effective range becomes `0 … 89,999,999`, so passports are variable width (observed `X288X` through `X89999997X`, i.e. 3–8 digits) rather than a fixed 8-digit number. Presumably the intent was:
```java
String identifierPassport = "X" + (person.randInt(99999999 - 10000000 + 1) + 10000000) + "X";
```

### Suggested fix

1. Maintain a population-level set of issued values per identifier type and re-draw on collision (mirroring the approach in #1601 for UUIDs).
2. Widen the `DRIVERS` keyspace so uniqueness is achievable for large populations (e.g. allow the full 8 digits after `S`, which keeps the existing 9-character width, or add more digits) — the current 90,000-value space cannot hold a realistic licensed-driver count.
3. Restore the intended low-bound offset on the `PASSPORT` line so numbers are consistently 8 digits.

Happy to open a PR if the maintainers agree on the preferred approach (global issued-id registry vs. widened formats).

### Environment

- Synthea: `master` @ `2b0a55bab0ab9ae22204320c80f5880ceb8925aa`
- Population: 100,000 living patients, CSV exporter
- OS: Linux

reachpratyuk-dotcom:
Fixes #1685 PASSPORT identifier format losing low-bound offset

SSN/DRIVERS/PASSPORT are not required to be unique per #1685
discussion, so no changes there — this only fixes the format bug.

### DIFF
--- a/src/main/java/org/mitre/synthea/export/rif/identifiers/Passport.java
+++ b/src/main/java/org/mitre/synthea/export/rif/identifiers/Passport.java
@@ -8,27 +8,28 @@ package org.mitre.synthea.export.rif.identifiers;
  */
 public class Passport extends FixedLengthIdentifier {
 
-    private static final char[] FIXED_X = {'X'};
-    private static final char[][] PASSPORT_FORMAT = {FIXED_X, NUMERIC, NUMERIC, NUMERIC, NUMERIC,
-            NUMERIC, NUMERIC, NUMERIC, NUMERIC, FIXED_X};
-    public static final long MIN_PASSPORT = 0;
-    public static final long MAX_PASSPORT = maxValue(PASSPORT_FORMAT);
+  private static final char[] FIXED_X = {'X'};
+  private static final char[][] PASSPORT_FORMAT = {FIXED_X, NUMERIC, NUMERIC, NUMERIC, NUMERIC,
+      NUMERIC, NUMERIC, NUMERIC, NUMERIC, FIXED_X};
+  public static final long MIN_PASSPORT = 0;
+  public static final long MAX_PASSPORT = maxValue(PASSPORT_FORMAT);
 
-    public Passport(long value) {
-        super(value, PASSPORT_FORMAT);
-    }
+  public Passport(long value) {
+    super(value, PASSPORT_FORMAT);
+  }
 
-    /**
-     * Parse a Passport from a String.
-     * @param str the string
-     * @return the Passport
-     */
-    public static Passport parse(String str) {
-        return new Passport(parse(str, PASSPORT_FORMAT));
-    }
+  /**
+   * Parse a Passport from a String.
+   *
+   * @param str the string
+   * @return the Passport
+   */
+  public static Passport parse(String str) {
+    return new Passport(parse(str, PASSPORT_FORMAT));
+  }
 
-    @Override
-    public Passport next() {
-        return new Passport(value + 1);
-    }
+  @Override
+  public Passport next() {
+    return new Passport(value + 1);
+  }
 }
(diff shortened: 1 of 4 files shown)


hadleynet:
Could you fix the checkstyle warnings in the new test files you added. Also, looks like there is a merge conflict.

reachpratyuk-dotcom:
Fixed the checkstyle warnings and conflicts.

hadleynet:
I'm still seeing a merge conflict, could you rebase your PR on the latest master branch - you might need to pull changes into your fork.

reachpratyuk-dotcom:
Hope all the merge conflicts are resolved now. Please let me know

reachpratyuk-dotcom:
Hi @hadleynet,
Just a quick note — I recently sent you a brief email regarding a verification of my contribution in this PR. Whenever you have a moment to take a look, I would greatly appreciate it.
Thank you again for your time and for reviewing the contribution!

PR #1688, opened 2026-07-18
Synthea br integration

richardpalmas:


jawalonoski:
Thank you for the well intentioned contribution, but I don't think we're going to take agent skills into the repo.

In fact, we probably should add `.agent/`, `.agents/`, `.claude/`, `.cursor/` and similar ilk to our `.gitignore` file.

Issue #1696, opened 2026-09-07
CSV export drops the panel observation itself; only its components are written


fmatten:
### What happened?

An observation that carries `component[]` instead of a top-level value is not written to
`observations.csv` at all. Its components are written as independent rows. The information
*that these values belong to one measurement* is present in the FHIR output and absent from
the CSV output of the same run.

### Where it happens

`src/main/java/org/mitre/synthea/export/CSVExporter.java`, `exportObservation`, lines 587-598
on `master` (`d9d07a6eef91ee5144293b42ab64224d84d124f8`):

```java
if (observation.value == null) {
  if (observation.observations != null && !observation.observations.isEmpty()) {
    // just loop through the child observations

    for (Observation subObs : observation.observations) {
      exportObservation(personID, encounterID, subObs);
    }
  }

  // no value so nothing more to report here
  return;
}
```

The panel itself has no top-level value, so it returns before writing its own row.

### Environment

```markdown
- OS: Debian GNU/Linux 13
- Java: openjdk version "21.0.12.1" 2026-08-18
- Synthea: release v4.0.0 asset `synthea-with-dependencies.jar`,
  sha256 `ed43c20ad40ba5c3bc724503a5af032715fe3c491620b766148e7c2361e6ecc1`
- Source lines quoted above are from `master` at commit
  `d9d07a6eef91ee5144293b42ab64224d84d124f8` (2026-08-18), i.e. the behaviour is current,
  not only present in the release jar.
```

### Relevant log output

```shell

```

PR #1710, opened 2026-09-18
Preserve coded panel observations in CSV export

Frankie-Xu:
Addresses #1696.

CSV export currently drops a coded panel observation when its value is null and only its component observations carry values. The component rows are written, but the panel's date, encounter, category, code, and description disappear.

This patch keeps the existing component recursion and writes one parent row for a null-valued observation that has a code and at least one component. The parent row uses the existing observations.csv columns: VALUE, UNITS, and TYPE are empty because the panel has no scalar value; DATE, PATIENT, ENCOUNTER, CATEGORY, CODE, and DESCRIPTION are retained. Uncoded null containers remain non-exported while their children continue to be traversed. No CSV columns or relationship schema are added, and numeric precision remains separate from #1709.

Tests:
- `CSVExporterPanelObservationTest` uses the real CSVExporter and CSVFileManager to cover panel plus two components, scalar observations, and null/no-code containers.
- Existing `CSVExporterTest` (11 tests) passes.
- `checkstyleMain` and `checkstyleTest` pass for the changed files. Checkstyle retains one pre-existing warning in unchanged `src/main/java/App.java:47`.

The focused test command reports 3 new tests plus 11 existing CSV exporter tests, all passing with zero skips. The full Synthea suite was not run.


src/main/java/org/mitre/synthea/export/CSVExporter.java
hadleynet: I think this change could be refactored to be clearer and avoid repeating the loop to write out child observations. Essentially:

```
if observation is valued or coded
    write line for observation
end
if observation is not valued but is coded
    write line for each child observation
end
```

As an added bonus the parent observation would always be written first.

Right now it looks like we skip writing out child observations if the parent has a value - wondering if we should always write out the child observations? If yes, the above could be simplified as:

```
if observation is valued or coded
    write line for observation
end
write line for each child observation
```

